PR #100, opened 2026-03-02
Add mypy to development dependencies

nitulkumardey-star:
Added mypy as a development dependency for testing.


ruvnet:
Closing — `mypy` is already a project dev dependency, listed in `requirements-dev.txt` (not `requirements.txt`). The dev-vs-runtime split is intentional: runtime users shouldn't have to install type-checkers. If you'd like to confirm the dev install works, `pip install -r requirements-dev.txt` should give you mypy and the other dev tools.

PR #139, opened 2026-03-05
 Compose file update by adding restart function for more robust

Harshit10j2004:
Added the restart method in the compose file , for removing the manual work of starting the compose restart after crashes or reboots.


ruvnet:
Rebase needed — current main has had multiple docker-compose changes (per-service env additions, ws ports). Once rebased, this 2-line restart-policy add should land cleanly.

PR #150, opened 2026-03-06
fix: make v1 Python API runnable on WSL2 with real WiFi sensing

SabaPivot:
## Summary

The v1 Python API had several issues that prevented it from starting and serving real-time data. This PR fixes all of them and adds WSL2 WiFi sensing support so the Live Demo and Sensing tabs work end-to-end with real WiFi signals.

### Bug Fixes

- **datetime serialization** — 4 places in `pose_service.py` put raw `datetime` objects into dicts broadcast via WebSocket, causing `TypeError: Object of type datetime is not JSON serializable`. Fixed with `.isoformat()`.
- **Zone confidence averaging** — formula `(0 + confidence) / 2` always halved the first person's confidence, causing values to fall below the stream filter threshold. Replaced with proper incremental mean.
- **CORS/allowed_hosts parsing** — `Settings` declared `cors_origins` and `allowed_hosts` as `List[str]`, but `.env` provides plain strings. Added `_list` properties that accept both JSON arrays and comma-separated strings.
- **Health endpoint 503** — `request.app.state.hardware_service` was never set because the orchestrator stored services only on `self`. Added service bridge in lifespan.
- **Module import paths** — 5 sensing files used `from v1.src.sensing.*` which fails when running from inside `v1/`. Fixed to `from src.sensing.*`.
- **PyTorch hard dependency** — `pose_service.py` imported torch at module level, crashing the API if PyTorch wasn't installed. Made it lazy-import on demand.

### Features

- **WSL2 WiFi support** — `WindowsWifiCollector` auto-detects WSL2 and uses `netsh.exe`; parses Korean-locale output fields (`신호`, `수신 속도`, `전송 속도`); captures throughput rates as additional metrics.
- **RSSI dithering** — `netsh` reports integer dBm, producing variance=0 and collapsing all spectral features. Added Gaussian dithering (σ=0.3 dBm) to overcome quantization, a standard signal processing technique for quantized ADCs.
- **Signal-derived pose stream** — `pose_stream.py` connects to the sensing WebSocket (`ws://localhost:8765`) and converts presence/motion classification into COCO 17-keypoint pose data, so the Live Demo shows real WiFi-based detection.
- **WebSocket proxy** — `/ws/sensing` endpoint in `app.py` bridges the UI to the sensing server, eliminating cross-origin WebSocket issues.
- **Static UI serving** — UI served at `/app/` so everything runs from a single `uvicorn` process.
- **Live source recognition** — UI sensing service now recognizes `windows_wifi`, `linux_wifi`, and `macos_wifi` as live data sources.

## Test Plan

- [x] All imports resolve correctly (`from src.sensing.*`)
- [x] Settings parsing works with both JSON arrays and comma-separated strings
- [x] `_sensing_to_pose()` returns valid 17-keypoint data when present, `None` when absent
- [x] Confidence incremental mean produces correct averages
- [x] WindowsWifiCollector initializes with dithering on WSL2
- [x] Sensing tab shows live RSSI with non-zero features
- [x] Live Demo receives pose frames and renders on canvas
- [x] Health endpoint returns all services healthy

### DIFF
--- a/example.env
+++ b/example.env
@@ -31,10 +31,10 @@ JWT_ALGORITHM=HS256
 JWT_EXPIRE_HOURS=24
 
 # Allowed hosts (restrict in production)
-ALLOWED_HOSTS=*  # Use specific domains in production: example.com,api.example.com
+ALLOWED_HOSTS=["*"]  # Use specific domains in production: ["example.com","api.example.com"]
 
 # CORS settings (restrict in production)
-CORS_ORIGINS=*  # Use specific origins in production: https://example.com,https://app.example.com
+CORS_ORIGINS=["*"]  # Use specific origins in production: ["https://example.com","https://app.example.com"]
 
 # =============================================================================
 # DATABASE SETTINGS
--- a/ui/components/SensingTab.js
+++ b/ui/components/SensingTab.js
@@ -216,7 +216,7 @@ export class SensingTab {
       // Map the service's dataSource to banner text and CSS modifier class.
       const dataSource = sensingService.dataSource;
       const bannerConfig = {
-        'live':              { text: 'LIVE \u2014 ESP32 HARDWARE',           cls: 'sensing-source-live' },
+        'live':              { text: 'LIVE \u2014 REAL WIFI DATA',            cls: 'sensing-source-live' },
         'server-simulated':  { text: 'SIMULATED \u2014 NO HARDWARE',        cls: 'sensing-source-server-sim' },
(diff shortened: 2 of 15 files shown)


ruvnet:
Closing — this PR touches the pre-rename `rust-port/wifi-densepose-rs/` or pre-archive `v1/` paths. After the v2 + archive/v1 reorganizations, the diff can't apply against current main. The underlying feature/fix may still be valuable as a clean PR against `v2/` or `archive/v1/` — please rebase if you want to bring it forward.

PR #151, opened 2026-03-06
fix: fall back when Linux WiFi probing fails

Gujiassh:
## Summary
- probe the Linux WiFi collector with `collect_once()` before selecting it so unreadable `/proc/net/wireless` setups fall back cleanly
- add a regression test covering the simulated fallback path on Linux collector probe failures

## Testing
- `PYTHONPATH=. python3 -m pytest -q -o addopts='' v1/tests/unit/test_ws_server.py`
- `PYTHONPATH=. python3 -m py_compile v1/src/sensing/ws_server.py v1/tests/unit/test_ws_server.py`

### DIFF
--- a/v1/src/sensing/ws_server.py
+++ b/v1/src/sensing/ws_server.py
@@ -65,6 +65,7 @@ ESP32_UDP_PORT = 5005
 # ESP32 UDP Collector — reads ADR-018 binary frames
 # ---------------------------------------------------------------------------
 
+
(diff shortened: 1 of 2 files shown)


Gujiassh:
Closing this stale PR because it has been open since 2026-03-06 with no maintainer response. I am keeping newer follow-up lines active instead of leaving this thread idle.

Issue #56, opened 2026-03-01
mac-os support stopped


bencium:
The wifi-densepose sensing server's --source wifi path only works on Windows (uses
 netsh commands). On macOS, --source auto silently falls back to simulation. The user
 wants real WiFi RSSI sensing on their M2 Pro MacBook — no simulated data.

 macOS 26.3 has removed the legacy airport CLI. CoreWLAN (via Swift) is the working
 API and provides real RSSI, channel, SSID, and noise data for all visible networks.
 BSSIDs are redacted by macOS privacy policy, but SSID+channel uniquely identifies
 each AP for the pipeline.

 Goal: Add a macOS CoreWLAN adapter so --source wifi works on macOS with real RSSI
 data flowing through the 8-stage sensing pipeline (motion detection, breathing
 extraction, presence).

 Approach

 1. Create Swift helper binary: macos-wifi-scan

 A small Swift CLI that calls CoreWLAN and outputs JSON to stdout. The Rust adapter
 calls this via Command::new() — same pattern as netsh on Windows.

 File: rust-port/wifi-densepose-rs/tools/macos-wifi-scan/main.swift

 // Calls CWWiFiClient.shared().interface().scanForNetworks()
 // Outputs JSON array: [{ssid, rssi, noise, channel, band, phy_mode}, ...]
 // Also includes connected network info

 Build with: swiftc -framework CoreWLAN -framework Foundation -o macos-wifi-scan
 main.swift

 2. Create Rust macOS adapter: MacosCoreWlanScanner

 File: rust-port/wifi-densepose-rs/crates/wifi-densepose-wifiscan/src/adapter/macos_sc
 anner.rs

 Implements the WlanScanPort trait:
 - scan() — calls the Swift helper, parses JSON output, returns Vec<BssidObservation>
 - connected() — returns the connected network's observation
 - Uses SSID:channel as a synthetic BSSID (since macOS redacts real BSSIDs)
 - Converts RSSI dBm to amplitude via existing rssi_to_amplitude()
 - Maps channel to BandType via existing BandType::from_channel()
 - Maps PHY mode string to RadioType

 3. Register adapter in module

 File: rust-port/wifi-densepose-rs/crates/wifi-densepose-wifiscan/src/adapter/mod.rs

 Add pub mod macos_scanner; and re-export MacosCoreWlanScanner.

 File: rust-port/wifi-densepose-rs/crates/wifi-densepose-wifiscan/src/lib.rs

[...]

PR #191, opened 2026-03-08
feat: add native macOS CoreWLAN Wi-Fi source

AIflow-Labs:
## Summary
This finishes the native macOS path discussed in #56 and ADR-025.

It adds a real CoreWLAN-backed Wi-Fi source for `sensing-server --source wifi` on macOS instead of falling back silently.

## What changed
- add the canonical Swift helper at `tools/macos-wifi-scan/main.swift`
- add a reproducible helper build script and helper discovery via `RUVIEW_MAC_WIFI_HELPER`
- implement a typed macOS scanner in `wifi-densepose-wifiscan`
- wire macOS probing, auto-detection, and the native Wi-Fi task in `sensing-server`
- document the macOS path honestly as RSSI/scan-based sensing for presence/coarse motion, not CSI parity

## Notes
- no compiled helper binaries are committed
- native macOS observations are labeled `wifi:<ssid>` when available, otherwise `wifi:macos`
- explicit `--source wifi` now returns a clear macOS-specific error when the helper/Wi-Fi path is unavailable instead of silently simulating

## Verification
- `cargo test --workspace`
- `cargo test --workspace --no-default-features`
- `cargo test -p wifi-densepose-wifiscan -p wifi-densepose-sensing-server`
- `./scripts/build-mac-wifi.sh`
- `rust-port/wifi-densepose-rs/target/tools/macos-wifi-scan/macos-wifi-scan --probe`
- `rust-port/wifi-densepose-rs/target/tools/macos-wifi-scan/macos-wifi-scan --scan-once`
- manual macOS run of `cargo run -p wifi-densepose-sensing-server -- --source wifi ...`
- API smoke check returned live `wifi:macos` data with non-zero BSSID observations on macOS

Addresses #56.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `CrossDomainEvaluator` — 6-metric evaluation protocol (MPJPE in-domain/cross-domain/few-shot/cross-hardware, domain gap ratio, adaptation speedup)
 - ADR-027: Cross-Environment Domain Generalization — 10 SOTA citations (PerceptAlign, X-Fi ICLR 2025, AM-FM, DGSense, CVPR 2024)
 - **Cross-platform RSSI adapters** — macOS CoreWLAN (`MacosCoreWlanScanner`) and Linux `iw` (`LinuxIwScanner`) Rust adapters with `#[cfg(target_os)]` gating
-- macOS CoreWLAN Python sensing adapter with Swift helper (`mac_wifi.swift`)
-- macOS synthetic BSSID generation (FNV-1a hash) for Sonoma 14.4+ BSSID redaction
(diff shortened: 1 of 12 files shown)


ruvnet:
Hi @AIflow-Labs — both this PR and #192 (your small follow-up) touch the same set of CHANGELOG/README/ADR-025 files plus the macOS sources. To land cleanly against current main:

1. **Rebase this PR against current main** (the CHANGELOG and the wifi-densepose-wifiscan crate have both moved since this opened).
2. After this lands, #192 should automatically resolve since it's the small follow-up. If you prefer, squash both into one PR — the README + ADR changes are very similar between them and you might end up with the same diff either way.

Architecturally, native CoreWLAN scan + explicit bridge source is exactly the right shape — primary path is the in-process scanner, with the helper-driven bridge as the opt-in fallback. Once rebased I'll do a focused review.

PR #192, opened 2026-03-08
feat: add explicit macOS Wi-Fi bridge source

AIflow-Labs:
## Summary
This is a small follow-up to #191.

It adds an explicit local bridge source for macOS when someone wants a helper-driven connected-AP stream, while keeping the native CoreWLAN scan path as the primary product path.

## What changed
- add `scripts/macos_wifi_bridge.py`
- add `--source macos-bridge` to the sensing server
- accept only loopback NDJSON records with `bridge_kind: connected_rssi`
- reject ESP32 binary payloads and disconnected bridge records
- document the bridge as explicit-only and experimental

## Notes
- `auto` never selects `macos-bridge`
- this does not masquerade as ESP32 or CSI; the API source is labeled `wifi-bridge:macos`
- this PR is stacked on #191 and is easiest to review after that lands

## Verification
- `cargo test --workspace`
- `cargo test --workspace --no-default-features`
- `python3 -m py_compile scripts/macos_wifi_bridge.py`
- `./scripts/build-mac-wifi.sh`
- `rust-port/wifi-densepose-rs/target/tools/macos-wifi-scan/macos-wifi-scan --probe`
- manual bridge smoke test against `--source macos-bridge`
- API smoke check returned `source: "wifi-bridge:macos"` with live ticking bridge data


ruvnet:
Building on #191 — please rebase the pair against current main; I'll review both once #191 is clean. See discussion on #191.

PR #236, opened 2026-03-12
fix(mobile): test the draft settings URL

Gujiassh:
## Summary

- let the mobile API service probe a one-off status URL without mutating the saved base URL
- wire Settings "Test Connection" to the current draft URL so users validate the address they are editing instead of the last saved server
- add one service test and one settings interaction test covering the draft-URL path

## Validation

- `cd ui/mobile && npm install`
- `cd ui/mobile && npm test -- --runInBand src/__tests__/services/api.service.test.ts src/__tests__/screens/SettingsScreen.test.tsx`

## Notes

- `npm run lint` in `ui/mobile` is currently blocked by a pre-existing repo tooling mismatch: the project ships `.eslintrc.js` but the script invokes ESLint 10, which expects flat config and then falls into plugin/version incompatibilities when forced into legacy mode. I did not change repo lint wiring in this PR.

### DIFF
--- a/ui/mobile/src/__tests__/screens/SettingsScreen.test.tsx
+++ b/ui/mobile/src/__tests__/screens/SettingsScreen.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react-native';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react-native';
 import { ThemeProvider } from '@/theme/ThemeContext';
+import { apiService } from '@/services/api.service';
 import { useSettingsStore } from '@/stores/settingsStore';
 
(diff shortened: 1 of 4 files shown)


Gujiassh:
Quick validation from my side on the current branch tip:

- `cd ui/mobile && node ./node_modules/jest/bin/jest.js --runTestsByPath src/__tests__/screens/SettingsScreen.test.tsx` → passed

So the draft-settings URL behavior covered by this PR is locally green on the current head.

Gujiassh:
Closing this stale radar PR to keep the queue clean: it has been open for weeks without maintainer response, and I am not going to keep it in the active follow-up set without renewed upstream interest.

PR #242, opened 2026-03-12
fix(esp32): resolve task watchdog timeout in edge_dsp task

chandusurisetty:
The 0ms delay causing the ESP32-S3 edge_dsp task to trigger the watchdog timer


ruvnet:
Thank you for the contribution. The watchdog timeout was already fixed in commit 024d2583 (v0.5.3). This PR contains primarily whitespace reformatting which would conflict with recent firmware changes (ADR-069 feature vectors, ADR-073 channel hopping). The `pdMS_TO_TICKS(1) → vTaskDelay(1)` change also reduces portability across tick rates. Closing — but thank you for looking into the watchdog issue!

PR #273, opened 2026-03-16
fix(led): Explicitly disable onboard WS2812 LED during CSI collection

Mathew005:
Description
This PR explicitly disables the onboard WS2812 RGB LED (NeoPixel) during CSI collection on ESP32-S3 boards.

The Problem
When compiling the esp32-csi-node firmware using ESP-IDF v5.2, the core Wi-Fi library automatically assumes control of any detected onboard NeoPixel to use as a Wi-Fi status indicator. Because CSI collection generates immense Wi-Fi traffic, this triggers the LED to shine at blinding brightness constantly, drawing unnecessary power and causing a distraction.

The Solution
Added the espressif/led_strip driver (v3.0.0+) to the project components.
Hooked into 
app_main
 to initialize the RMT peripheral on GPIO 38.
Explicitly command the strip to clear (0,0,0) its latched state right before the Wi-Fi station connects.
Testing
Flashed to an ESP32-S3-WROOM-1 N8R8 board.
Successfully verified the RGB LED clears to black during boot and remains off while throwing UDP CSI frames.
Verified LVGL UI elements continue compounding without interruption.

### DIFF
--- a/firmware/esp32-csi-node/main/idf_component.yml
+++ b/firmware/esp32-csi-node/main/idf_component.yml
@@ -8,3 +8,6 @@ dependencies:
 
   ## LCD touch abstraction
   espressif/esp_lcd_touch: "^1.0"
+
+  ## Onboard WS2812 LED Disabling
+  espressif/led_strip: "^3.0.0"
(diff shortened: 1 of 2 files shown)


mknippen:
+1

PR #371, opened 2026-04-08
docs: clarify macOS live sensing setup

Safrolomancer:
## Summary

This PR clarifies the currently working macOS live Wi-Fi sensing flow and improves the UI fallback path for the legacy Python/WebSocket server.

## Changes

- document the tested macOS CoreWLAN live-sensing workflow in the user guide
- add a README note that the main Rust auto-detect path still falls back to simulation on macOS
- teach the UI sensing service to fall back from same-origin `/ws/sensing` to `ws://localhost:8765/ws/sensing` for the legacy macOS/Python path
- ignore named local virtual environments such as `.venv-macos-sensing`

## Validation

- verified live CoreWLAN helper output locally on macOS
- started the legacy `python -m v1.src.sensing.ws_server` path successfully
- confirmed WebSocket frames were received from the `macos_wifi` source
- ran `node --check ui/services/sensing.service.js`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -170,6 +170,7 @@ celerybeat.pid
 # Environments
 .env
 .venv
+.venv-*/
 env/
 venv/
 ENV/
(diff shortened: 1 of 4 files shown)


ruvnet:
Closing — the doc additions point at `python -m v1.src.sensing.ws_server` and `python v1/data/proof/verify.py`, both of which moved to `archive/v1/` (PR #590 + ADR-028 archive). Re-applying this PR cleanly would mean updating those paths and verifying the legacy CoreWLAN/WebSocket entry point still works under `archive/v1/src/sensing/`. The intent (clarify macOS live-sensing setup) is genuinely useful — if you want to re-submit against the archive/v1 paths, happy to review.

PR #372, opened 2026-04-08
fix: fall back to legacy sensing websocket

Safrolomancer:
This keeps the UI working with the legacy Python/macOS sensing server by trying ws://localhost:8765/ws/sensing when same-origin /ws/sensing is unavailable. It also treats macos_wifi as a live source so the UI does not label a real macOS stream as simulated.

### DIFF
--- a/ui/services/sensing.service.js
+++ b/ui/services/sensing.service.js
@@ -9,11 +9,13 @@
  * emit simulated frames so the UI can clearly distinguish live vs. fallback data.
  */
 
-// Derive WebSocket URL from the page origin so it works on any port.
-// The /ws/sensing endpoint is available on the same HTTP port (3000).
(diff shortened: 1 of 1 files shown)


ruvnet:
Closing — superseded by **#572** (`fix(ui): map sensing websocket port for docker`) which landed yesterday. #572 added `buildSensingWsUrl()` with explicit port-mapping (Docker HTTP 3000 → WS 3001, Python 8080 → WS 8765) and the missing-connection guards in `websocket.service.js`. The intent of this PR (fallback when the primary WS path is unavailable) is now covered by that port-mapping logic plus the cleanup-race guards. If there's a specific failure mode #572 doesn't cover, please file a focused issue with the reproduction.

PR #376, opened 2026-04-10
fix(websocket): handle broadcast Lagged error + ping keepalive

schwarztim:
When the broadcast channel's recv() returns Lagged (client fell behind), the server previously treated it as fatal and dropped the connection. The client immediately reconnected, creating a visible 2-4s freeze cycle on the dashboard.

**Fix:**
- Lagged error: skip missed frames with `continue` instead of `break`
- Add 30s ping/pong keepalive to prevent reverse proxy idle timeouts (Caddy, nginx default 60s idle → drops long-lived WS connections)

**Testing:** 11 ESP32-S3 nodes streaming at ~20fps per node, sustained 154 frames over 8s, zero disconnects.

Fixes dashboard freezing in multi-node deployments.


schwarztim:
Closing in favor of #484 — same fix, but rebased onto the post-rename v2/ paths. PR #484 is current and ready for review.

Issue #375, opened 2026-04-09
Firmware sends node_id=1 for all nodes (relates to #232)


shy-workstation:
Despite NVS provisioning each node with a different node_id (1-4), all UDP packets contain node_id=1 at byte offset 4.
   Boot log confirms NVS reads correctly:

  nvs_config: NVS override: node_id=4
  main: ESP32-S3 CSI Node (ADR-018) — Node ID: 4
  csi_collector: CSI collection initialized (node_id=1, channel=11)

  NVS loads node_id=4 but csi_collector still uses 1. Same issue as #232.

  Workaround: The relay script patches byte 4 based on source IP address.

ruvnet:
Fixed in v0.6.1-esp32 (PR #393). Defensive `node_id` capture at init prevents all consumers from reading a clobbered global. Please reflash: https://github.com/ruvnet/RuView/releases/tag/v0.6.1-esp32

Issue #366, opened 2026-04-07
Rust build errors


Leeeeeeedw:
See the below picture （System: Windows11, use openblas installed from vcpkg）

<img width="3020" height="2034" alt="Image" src="https://github.com/user-attachments/assets/25c9dd90-7c15-44b3-9d54-fc6edb9d4ed7" />

Leeeeeeedw:
This errors show after v0.5.5 (I can build successfully based on commit 4e9e92d)

ruvnet:
The screenshot makes it hard to diagnose the specific error. Could you paste the error text directly as a code block?

In the meantime, try building with `--no-default-features`, which skips optional GPU/BLAS dependencies that commonly fail on Windows:

```bash
cargo test --workspace --no-default-features
```

If the errors are OpenBLAS linker errors from vcpkg, this flag should bypass that dependency entirely. The core functionality compiles and tests without it.

---
_Generated by [Claude Code](https://claude.ai/code)_

Leeeeeeedw:
I try to use --no-default-features parameter to build again today (with v0.6.2 source code), it still shows errors, see the below

```
Compiling sysinfo v0.32.1
warning: `wifi-densepose-signal` (lib) generated 5 warnings (run `cargo fix --lib -p wifi-densepose-signal` to apply 1 suggestion)
error[E0432]: unresolved import `ruvector_mincut`
 --> crates\wifi-densepose-sensing-server\src\csi.rs:5:5
  |
5 | use ruvector_mincut::{DynamicMinCut, MinCutBuilder};
  |     ^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `ruvector_mincut`
  |
  = help: if you wanted to use a crate named `ruvector_mincut`, use `cargo add ruvector_mincut` to add it to your `Cargo.toml`

error[E0432]: unresolved import `ruvector_mincut`
  --> crates\wifi-densepose-sensing-server\src\main.rs:27:5
   |
27 | use ruvector_mincut::{DynamicMinCut, MinCutBuilder};
   |     ^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `ruvector_mincut`
   |
   = help: if you wanted to use a crate named `ruvector_mincut`, use `cargo add ruvector_mincut` to add it to your `Cargo.toml`

error[E0425]: cannot find value `CLASSES` in module `adaptive_classifier`
    --> crates\wifi-densepose-sensing-server\src\main.rs:3227:45
[...]

Issue #268, opened 2026-03-15
no nodes found


FLATEARTHER11:
everything i have done through the process but still app doesnt work for me please guide me through a video tutorial to it, i have tried that older version too but local host 3000 website opens but does nothing. app sense but no visualise through esp32 s3. please provide a video link i cant find on youtube

ruvnet:
Thanks for trying RuView! Here's a step-by-step troubleshooting guide since there isn't a video tutorial yet.

## Quick Diagnosis

The "no nodes found" message means the server isn't receiving data from your ESP32-S3. Let's fix that.

### Step 1: Verify the ESP32 is sending data

Open a serial monitor to check the ESP32 output:
```bash
pip install pyserial
python -m serial.tools.miniterm COM_PORT 115200
```
Replace `COM_PORT` with your port (e.g. `COM3` on Windows, `/dev/ttyUSB0` on Linux).

**You should see:**
```
I (xxx) main: Got IP: 192.168.x.x
I (xxx) main: Connected to WiFi
I (xxx) csi_collector: CSI collection initialized
I (xxx) main: CSI streaming active → 192.168.x.x:5005
```

**If you see `Mock CSI mode`** → you're running mock firmware. Download the real firmware from [v0.4.3.1](https://github.com/ruvnet/RuView/releases/tag/v0.4.3.1-esp32) and reflash.

**If you don't see `Got IP`** → WiFi isn't connecting. Re-provision:
```bash
python firmware/esp32-csi-node/provision.py --port COM_PORT \
  --ssid "YourWiFi" --password "YourPassword" \
  --target-ip YOUR_COMPUTER_IP --target-port 5005
```

### Step 2: Find your computer's IP

The `--target-ip` must be your computer's local network IP (not `localhost`):
- **Windows:** Open CMD → `ipconfig` → look for `IPv4 Address` under your WiFi adapter
- **Mac:** System Preferences → Network → WiFi → IP Address
- **Linux:** `ip addr show wlan0`

### Step 3: Verify data is arriving

On your computer, check if UDP packets arrive on port 5005:
```bash
# Linux/Mac
nc -lu 5005 | xxd | head

# Windows (PowerShell)
# Or just check the server logs
```

### Step 4: Start the server correctly

```bash
# Docker (easiest)
docker run -p 3000:3000 -p 5005:5005/udp ruvnet/wifi-densepose:latest

# Then open http://localhost:3000
```

### Common Issues

[...]

PR #377, opened 2026-04-10
docs: troubleshooting guide for ESP32 CSI deployments

schwarztim:
Adds `docs/TROUBLESHOOTING.md` covering 8 common issues encountered during multi-node ESP32-S3 WiFi CSI deployments:

1. **Node not appearing** — limping state after USB flash, watchdog recovery
2. **Person count stuck at 1** — ADR-044 root causes and fixes
3. **Vitals jitter** — last-write-wins from multiple nodes, best-node selection
4. **Signal quality placeholder** — was hardcoded, now uses adaptive p95
5. **Dashboard freezing** — WebSocket disconnect loop (see PR #376)
6. **OTA crash at 59%** — BLE/OTA Core 1 conflict
7. **SSH LAN hang** — Tailscale workaround
8. **USB-C port selection** — left port for UART bridge

Helps users hitting #268, #375, #366.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -1,0 +1,111 @@
+# RuView Troubleshooting Guide
+
+Known issues and fixes from the rebase-to-upstream branch (upstream #301).
+
+---
+
+## 1. Node not appearing in /api/v1/nodes
+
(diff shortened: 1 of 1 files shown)


ruvnet:
Great community docs — merged. Thanks @schwarztim!

schwarztim:
So, about two weeks ago I had 11 ESP32 nodes delivered and I have done a ton of work for RUView. I easily get distracted, but I think it would be good to show you those contributions and discuss what you like and don't like about it. @ruvnet

PR #379, opened 2026-04-10
feat(adr-018): source MAC in CSI frame V2 (magic 0xC5110006)

taylorjdawson:
## Summary

- extend ADR-018 binary frame format with a 6-byte source MAC at offset 20 using V2 magic `0xC5110006` and a 26-byte header
- preserve backward compatibility by teaching the raw CSI JS and Python tooling to accept both V1 and V2 frames through a shared decoder in `scripts/lib/raw-csi.js`
- resolve the protocol collision with ADR-069 feature packets, which keep magic `0xC5110003`
- harden the sensing server parser so malformed packets with zero antennas or zero subcarriers are rejected instead of being treated as empty CSI

## Validation

- `node --check` on the touched JS scripts
- `python3 -m py_compile` on the touched Python scripts
- synthetic V1 and V2 sanity check for `scripts/lib/raw-csi.js`
- `cargo test -p wifi-densepose-hardware esp32_parser --no-default-features`
- `wifi-densepose-sensing-server` still has pre-existing unrelated build failures (`ruvector_mincut` imports, missing `adaptive_classifier::CLASSES`, and borrow checker errors in `main.rs`), so it could not be used as the final backstop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ downloads/
 eggs/
 .eggs/
 lib/
+!scripts/lib/
+!scripts/lib/raw-csi.js
 lib64/
 parts/
 sdist/
--- a/docs/adr/ADR-018-esp32-dev-implementation.md
+++ b/docs/adr/ADR-018-esp32-dev-implementation.md
@@ -35,13 +35,15 @@ This ADR answers *how* to build it — the concrete development sequence, the sp
 
 ### Binary Frame Format (implemented in `esp32_parser.rs`)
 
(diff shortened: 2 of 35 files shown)


taylorjdawson:
Superseding this PR rather than trying to revive the old ADR-018 V2 wire-format change. Upstream has since allocated the relevant packet magic values for the ADR-081 feature-state / rvCSI direction, and host-side ESP32 parsing is moving toward the rvCSI adapter layer. If source MAC still needs to land, it should come back as normalized source metadata in the future rvCSI ESP32 adapter path instead of this conflicting RuView frame variant.

PR #402, opened 2026-04-18
fix: Docker entrypoint arg handling + configurable model directory

voidborne-d:
## Summary

Fixes #384 and #399 by replacing the broken Docker entrypoint pattern and making the model scan directory configurable.

### Issue #384: `--source: not found` when passing flags to `docker run`

The Dockerfile used `ENTRYPOINT ["/bin/sh", "-c"]` with a shell-form CMD. When users passed flags:

```bash
docker run ruvnet/wifi-densepose:latest --source wifi --tick-ms 500
```

Docker replaced CMD entirely, producing `/bin/sh -c "--source wifi --tick-ms 500"` which tries to execute `--source` as a shell command → `--source: not found`.

The `--` separator also fails: `docker run ... -- --source wifi` produces `/bin/sh -c "-- --source wifi"` → `--: not found`.

### Issue #399: `--model` / `--load-rvf` parameters ignored in Docker

The model discovery API (`GET /api/v1/models`, `scan_model_files()`) hardcodes the path `data/models`. When Docker users mount models to `/app/models/`:

```bash
docker run -v /path/to/models:/app/models ruvnet/wifi-densepose:latest \
  /app/sensing-server --model /app/models/wiflow-v1.rvf
```

The `--model` CLI flag loads the specific file correctly, but `GET /api/v1/models` scans `/app/data/models/` and finds nothing → `Discovered 0 model files`.

### 1. `docker/docker-entrypoint.sh` (new)

A proper entrypoint script that handles three usage patterns:

```bash
# Pattern 1: Use env vars (no args)
docker run -e CSI_SOURCE=esp32 ruvnet/wifi-densepose:latest

# Pattern 2: Pass CLI flags directly
docker run ruvnet/wifi-densepose:latest --source esp32 --tick-ms 500

# Pattern 3: Override command entirely
docker run ruvnet/wifi-densepose:latest /bin/sh
```

When the first argument starts with `-` (or no args), the script prepends the server binary with sensible defaults. User-passed flags are appended and override defaults via clap's last-wins behavior.

Also sets `--bind-addr 0.0.0.0` (the binary defaults to `127.0.0.1` which blocks access from outside the container).

### 2. `MODELS_DIR` environment variable

The model scan path is now configurable:

```bash
docker run -v /path/to/models:/app/models -e MODELS_DIR=/app/models \
  ruvnet/wifi-densepose:latest --source esp32
```

[...]

PR #407, opened 2026-04-20
fix/docker: pass args directly to binary instead of shell

### LINKED ISSUE #384: can't start docker container on windows - wifi: 1: --source: not found
when i try to start docker container using command from documentation it just can't start
PS C:\Users\User> docker run --network host ruvnet/wifi-densepose:latest -- --source wifi --tick-ms 500
wifi: 1: --source: not found
PS C:\Users\User> docker run --network host ruvnet/wifi-densepose:latest --source wifi --tick-ms 500
/bin/sh: 0: Illegal option --
When i start container without arguments it can't detect any hardware

fuleinist:
## Problem

Docker run arguments (e.g. `--source wifi`) don't reach the `sensing-server` binary on Windows (and sometimes Linux). Users see:

- **PowerShell**: `wifi: 1: --source: not found` — PowerShell interprets `--` as a parameter prefix
- **Linux sh**: `/bin/sh: 0: Illegal option --` — Docker dispatches `--args` to `/bin/sh -c` which chokes on `--`

## Root Cause

The `Dockerfile.rust` uses shell-form ENTRYPOINT:

```dockerfile
ENTRYPOINT ["/bin/sh", "-c"]
CMD ["/app/sensing-server --source ${CSI_SOURCE} --tick-ms 100 ..."]
```

With `/bin/sh -c`, the entire CMD becomes a shell command string. Any arguments after the container image name are treated as `sh` options or extra positional args to the shell — they never reach `sensing-server`.

## Fix

Switch to exec-form ENTRYPOINT so arguments pass directly to the binary:

```dockerfile
ENTRYPOINT ["/app/sensing-server"]
CMD []
```

The server already reads `CSI_SOURCE` from the environment, so `-e CSI_SOURCE=simulated` continues to work for env-based configuration. Users who prefer CLI args can now write:

```bash
docker run ruvnet/wifi-densepose:latest --source simulated
```

Also update `docker-compose.yml` command to exec-form JSON array.

## Testing

```bash
# Before fix (fails on Windows):
#   docker run ruvnet/wifi-densepose:latest --source wifi
#   wifi: 1: --source: not found

# After fix (works on all platforms):
docker run ruvnet/wifi-densepose:latest --source simulated
# → Server starts in simulated mode

docker run -e CSI_SOURCE=simulated ruvnet/wifi-densepose:latest
# → Still works (env var takes precedence in auto mode)

docker compose run sensing-server --source simulated
# → compose also passes args correctly
```

Fixes #384

### DIFF
--- a/docker/Dockerfile.rust
+++ b/docker/Dockerfile.rust
@@ -50,7 +50,12 @@ ENV RUST_LOG=info
 # Override at runtime:  docker run -e CSI_SOURCE=esp32 ...
 ENV CSI_SOURCE=auto
 
-ENTRYPOINT ["/bin/sh", "-c"]
-# Shell-form CMD allows $CSI_SOURCE to be substituted at container start.
-# The ENV default above (CSI_SOURCE=auto) applies when the variable is unset.
-CMD ["/app/sensing-server --source ${CSI_SOURCE} --tick-ms 100 --ui-path /app/ui --http-port 3000 --ws-port 3001"]
+ENTRYPOINT ["/app/sensing-server"]
+# Exec-form CMD allows arguments after the image name to reach the binary directly.
+# Use -e CSI_SOURCE=... to control the data source at runtime.
+# Examples:
+#   docker run ruvnet/wifi-densepose:latest                            # auto mode (default)
+#   docker run -e CSI_SOURCE=simulated ruvnet/wifi-densepose:latest  # synthetic data
+#   docker run -e CSI_SOURCE=esp32 ruvnet/wifi-densepose:latest       # real ESP32 hardware
+#   docker run ruvnet/wifi-densepose:latest --source simulated        # equivalent to above
+CMD []
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -18,8 +18,8 @@ services:
       #          wifi           — use host Wi-Fi RSSI/scan data (Windows netsh)
       #          simulated      — generate synthetic CSI data (no hardware required)
       - CSI_SOURCE=${CSI_SOURCE:-auto}
-    # command is passed as arguments to ENTRYPOINT (/bin/sh -c), so $CSI_SOURCE is expanded by the shell.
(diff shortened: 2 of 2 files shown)


proffesor-for-testing:
Thanks @fuleinist for spotting the shell-form ENTRYPOINT problem and writing a clear root-cause explanation — the PowerShell vs `/bin/sh -c "--"` breakdown is exactly right, and it was useful evidence when reproducing locally. The exec-form swap does unblock `docker run image --source wifi`, which I verified with a stub-binary Docker image.

proffesor-for-testing:
@ruvnet — recommend **closing in favor of #402**. Both PRs fix the shell-form ENTRYPOINT (issue #384), and that part of this PR is correct. However, after locally reproducing both against `upstream/main`:

1. **`CSI_SOURCE` env var is not actually read by the binary.** The PR description says "The server already reads CSI_SOURCE from the environment" — but in `sensing-server/src/main.rs:98` the `--source` arg has no clap `env = "…"` attribute. So `docker run -e CSI_SOURCE=esp32 image` is silently ignored under this PR.
2. **Binary defaults mismatch the Dockerfile.** With no CLI flags, clap's defaults take over:
   - `--http-port` → `8080` vs `EXPOSE 3000`
   - `--ws-port` → `8765` vs `EXPOSE 3001`
   - `--bind-addr` → `127.0.0.1` (won't accept Docker bridge traffic, so `-p 3000:3000` doesn't work)
   - `--ui-path` → `../../ui` (UI not served)

   Result: `docker run ruvnet/wifi-densepose:latest` still doesn't produce a usable server. `docker-compose up` works only because compose re-supplies every flag explicitly.
3. **Issue #399 (MODELS_DIR / discovered models empty when volume-mounted) is not addressed.**

#402 handles all of the above via a dedicated entrypoint script that injects the correct defaults while letting user flags override through clap last-wins.

Thanks again to @fuleinist — the root-cause write-up here is solid and wasn't wasted work; it aligns with the diagnosis in #402.

Issue #415, opened 2026-04-22
Build failure rust


ErfanDL:
I got this error when build

   Compiling openblas-src v0.10.15
   Compiling wasm-bindgen-shared v0.2.114
   Compiling anstyle-parse v0.2.7
   Compiling erased-serde v0.4.10
error: failed to run custom build command for `openblas-src v0.10.15`

Caused by:
  process didn't exit successfully: `C:\Users\Lumine\Desktop\RuView-0.6.2-esp32\RuView-0.6.2-esp32\rust-port\wifi-densepose-rs\target\release\build\openblas-src-83e779c108efea69\build-script-build` (exit code: 101)
  --- stderr

  thread 'main' (14152) panicked at C:\Users\Lumine\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\openblas-src-0.10.15\build.rs:129:13:
  Non-vcpkg builds are not supported on Windows. You must use the 'system' feature.
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...

Aleksahek:
debian forky
+

ruvnet:
This is a known Windows build issue. The `openblas-src` crate requires either vcpkg (with OpenBLAS installed) or the `system` feature on Windows for native builds.

**Quick fix — skip the optional BLAS dependency entirely:**
```bash
cargo build --release --no-default-features
```
This bypasses `openblas-src` and all other optional GPU/BLAS dependencies. The core sensing server and firmware tooling compile cleanly without them.

**If you need OpenBLAS on Windows** (for the training pipeline features):
```powershell
# Install vcpkg and OpenBLAS, then build with the system feature
vcpkg install openblas:x64-windows
cargo build --release --features system
```

See #366 for a similar report and the maintainer's response. The `--no-default-features` path is the recommended approach for Windows development and standard deployments.

---
_Generated by [Claude Code](https://claude.ai/code)_

ruvnet:
Update: the `--no-default-features` workaround I suggested earlier was actually broken on `main` until #425 merged just now (58a63d6).

[...]

PR #417, opened 2026-04-23
fix: add Windows ndarray-linalg openblas-system feature

D2758695161:
## Summary

Fix Windows build failure for `openblas-src` in the Rust port. On Windows, `openblas-src` does not support non-vcpkg builds — it requires either the `system` feature or vcpkg bootstrap. This PR adds target-specific ndarray-linalg configuration so Windows uses `openblas-system` feature (links against pre-installed system OpenBLAS) while other platforms continue using `openblas-static`.

## Root Cause

```
error: failed to run custom build command for `openblas-src v0.10.15`
Caused by:
  process didn't exit successfully: ...build-script-build (exit code: 101)
  thread 'main' (14152) panicked at ...openblas-src-0.10.15\build.rs:129:13:
  Non-vcpkg builds are not supported on Windows. You must use the 'system' feature.
```

`ndarray-linalg@0.16` with `features = ["openblas-static"]` uses `openblas-src` which only supports vcpkg builds on Windows.

## Fix

**Cargo.toml** — split `ndarray-linalg` into:
- Default (Linux/macOS): `features = ["openblas-static"]` (unchanged)
- Windows target: `features = ["openblas-system"]` (links against system-installed OpenBLAS)

```toml
ndarray-linalg = { version = "0.16", default-features = false, features = ["openblas-static"] }

[target.'cfg(windows)'.dependencies]
ndarray-linalg = { version = "0.16", default-features = false, features = ["openblas-system"] }
```

**build-guide.md** — added Windows section explaining vcpkg installation and the `openblas-system` option.

## Testing

The fix was verified by checking the Cargo.toml resolves correctly for both Windows and non-Windows targets. Full CI build on Windows requires OpenBLAS installed via vcpkg or system package manager.

## Related Issue

- Reported in https://github.com/ruvnet/RuView/issues/415

### DIFF
--- a/docs/build-guide.md
+++ b/docs/build-guide.md
@@ -186,6 +186,11 @@ A high-performance Rust port with ~810x speedup over the Python pipeline for the
 
   # macOS
   brew install openblas
+
+  # Windows (via vcpkg)
+  # Install vcpkg: git clone https://github.com/microsoft/vcpkg.git
+  # Then run: vcpkg install openblas:x64-windows
+  # Or use `openblas-system` feature (default on Windows) which links against system OpenBLAS
   ```
 
 ### Build
--- a/rust-port/wifi-densepose-rs/Cargo.toml
+++ b/rust-port/wifi-densepose-rs/Cargo.toml
@@ -50,7 +50,11 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 
 # Signal processing
 ndarray = { version = "0.15", features = ["serde"] }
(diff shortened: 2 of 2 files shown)


ruvnet:
Closing — this PR touches `rust-port/wifi-densepose-rs/`, which was renamed to `v2/` long ago. The diff cannot apply against current main. The underlying intent (Windows openblas / macOS Rust build / Confidence trit_signal / node_id NVS) may still be valuable as a clean PR against `v2/` — happy to re-review if rebased.

PR #434, opened 2026-04-26
docs(adr): ADR-086 — edge novelty gate (proposed)

ruvnet:
## Summary

Pushes the [ADR-084](docs/adr/ADR-084-rabitq-similarity-sensor.md) novelty sensor down into the ESP32 sensor MCU's Layer 4 (On-device Feature Extraction) of [ADR-081](docs/adr/ADR-081-adaptive-csi-mesh-firmware-kernel.md)'s 5-layer kernel. When novelty falls below the gate threshold, the firmware *suppresses the UDP send* — saving bandwidth, RF airtime, and TX energy in steady-state quiet rooms.

## What this changes (and what it does NOT)

| Layer | Change |
|---|---|
| **Sensor MCU firmware** | New `rv_sketch.{h,c}` (no_std 1-bit sketch + 32-slot IRAM ring bank), three Kconfig flags, `rv_feature_state_t` magic bumps to `0xC5110007` |
| **Wire format** | `0xC5110007` adds `suppressed_since_last: u16` + `gate_version: u8` *without growing payload* (narrows existing 16-bit `quality_flags` to 8-bit; only 8 bits were ever defined). v6 receivers fall back gracefully. |
| **Cluster-Pi receiver** | Reconciliation: when `suppressed_since_last > 0`, treat the gap as low-novelty contribution to the bank rather than missing data |
| **ADR-084 server work** | Unchanged. ADR-086 is the firmware twin of the cluster-Pi novelty gate. |

## Decision summary

- **Gating policy**: suppress UDP send when novelty < `CONFIG_RV_EDGE_NOVELTY_THRESHOLD` (basis-point integer, default 500 = 0.05).
- **Stuck-gate self-heal**: never suppress more than `CONFIG_RV_EDGE_MAX_CONSEC_SUPPRESS` (default 50 frames ≈ 10 s at 5 Hz).
- **Default-off Kconfig** (`CONFIG_RV_EDGE_NOVELTY_GATE_ENABLE=n`) so existing deployments behave identically until opted in.
- **Computation budget**: ≤ 200 µs sketch insert+score per frame on Xtensa LX7 at 240 MHz.

## Validation commitments

| Metric | Target |
|---|---:|
| Sketch insert + score on Xtensa LX7 | ≤ 200 µs |
| UDP TX-energy reduction (steady-state quiet rooms) | ≥ 30% |
| Cluster-Pi novelty top-K coverage drop vs unsuppressed | ≤ 5 pp |
| Bandwidth reduction (stable-room scenarios) | ≥ 50% |
| Stuck-gate self-heal | exercised by unit test, ≤ 10 s suppression cap |

## Six-pass implementation plan

1. **`rv_sketch.{h,c}` no_std primitive** — 1-bit sign quantization, packed-byte storage, 8-bit-table popcount for hamming. QEMU validation first.
2. **IRAM ring bank** — `RV_EDGE_BANK_SIZE` × `RV_EDGE_VECTOR_DIM_BYTES` bytes. FIFO eviction.
3. **Kconfig flags** — three, all default-off / safe.
4. **`rv_feature_state_t` v7 wire format** — magic + new fields, v6 fallback on receiver.
5. **Cluster-Pi reconciliation** — `suppressed_since_last` interpolation in the novelty bank.
6. **QEMU + COM7 hardware-in-loop validation** — measure suppression rate on stable-room CSI; assert acceptance numbers.

## Open questions explicitly flagged

1. **Xtensa LX7 POPCNT** — claim "no hardware POPCNT" is **conjecture** (no primary-source check yet). Pass 2 bench is the falsifier.
2. **PSRAM-backed bank** when the board has it (8 MB on the Waveshare AMOLED).
3. **`gate_version: u8` provenance** — Kconfig-pinned, NVS-stored, or per-firmware-release embedded?
4. **(Likeliest weak point)** Interaction with [ADR-082](docs/adr/ADR-082-pose-tracker-confirmed-output-filter.md) Tentative→Active gate — the edge gate may suppress legitimate first-detection frames the cluster-Pi tracker needs for promotion. Mitigations are speculative.

## Reserved follow-up ADR slots (pointer stubs in §"Related ADR slots")

- **ADR-087 (prospective)** — Pass-4 mesh-exchange scope clarification: locks ADR-084 Pass 4's cluster↔cluster reading; reserves the slot for sensor→Pi UDP compression if scope shifts.
- **ADR-088 (prospective)** — Firmware-release coordination policy: when does a server-side change require a firmware release vs not.

## Why this lands as its own ADR (not folded into ADR-084)

ADR-084 explicitly states "Sensor MCU is unchanged by this ADR; sketches happen at the cluster Pi." Pushing the gate down to the MCU is a *different* engineering decision with its own cost/benefit, its own threat model (firmware Secure-Boot V2 + signed images per ADR-028), and its own release cycle. Keeping it separate makes the cluster-Pi work shippable independently of the firmware release cycle.

## Test plan
- [x] ADR text references existing ADRs correctly (ADR-018, ADR-028, ADR-066, ADR-081, ADR-082, ADR-083, ADR-084, ADR-085)
- [x] All path references on `v2/` and `archive/v1/` (post-rename)
- [x] Cites primary sources where claims are real; marks conjecture explicitly where they're not
- [x] Six-pass plan is independently shippable per pass
- [ ] Reviewer accepts the v7 wire format (magic `0xC5110007`, narrowed `quality_flags`, two new fields)
- [ ] Reviewer accepts the conjecture flag on Xtensa LX7 POPCNT (Pass 2 bench will resolve)

### DIFF
--- a/docs/adr/ADR-086-edge-novelty-gate.md
+++ b/docs/adr/ADR-086-edge-novelty-gate.md
@@ -1,0 +1,423 @@
+# ADR-086: Edge Novelty Gate — Push the RaBitQ Sensor Down to the Sensor MCU
+
+| Field          | Value                                                                                                                                        |
+|----------------|----------------------------------------------------------------------------------------------------------------------------------------------|
+| **Status**     | Proposed                                                                                                                                     |
+| **Date**       | 2026-04-26                                                                                                                                   |
+| **Authors**    | ruv                                                                                                                                          |
+| **Refines**    | ADR-081 (5-layer adaptive CSI mesh firmware kernel — Layer 4 / On-device feature extraction), ADR-084 (RaBitQ similarity sensor) |
+| **Touches**    | ADR-018 (binary CSI frame magic discipline), ADR-028 (capability audit / witness verification), ADR-082 (confirmed-track output filter), ADR-085 (RaBitQ pipeline expansion) |
+| **Companion**  | `firmware/esp32-csi-node/main/rv_feature_state.h` (current `0xC5110006` v6 wire format), `docs/research/architecture/three-tier-rust-node.md` (BQ24074 power budget context), `vendor/ruvector/crates/ruvector-core/src/quantization.rs::BinaryQuantized` (std reference implementation that this ADR will not directly reuse on-MCU) |
+
+## Context
+
+ADR-081's 5-layer firmware kernel today emits one `rv_feature_state_t`
+packet per node every 100–1000 ms (1–10 Hz, default 5 Hz on COM7),
+60 bytes payload, magic `0xC5110006`, regardless of how interesting
+the underlying CSI window was. At a 5 Hz baseline the per-node steady-
+state load is ~300 B/s of UDP plus the radio TX duty that emits it.
+Across a 12-node deployment the cluster Pi sees ~3.6 kB/s of
+feature-state — not a bandwidth crisis on its own, but every one of
+those packets also costs sensor-MCU radio TX energy, every one
+contends for ESP-WIFI-MESH airtime per ADR-081 Layer 3, and every one
+runs through the cluster-Pi novelty bank ADR-084 Pass 3 only to be
+classified as "nothing new" most of the time in a quiet room.
+
+ADR-084 made novelty cheap on the cluster-Pi side. The same novelty
+sensor is structurally local: a sketch, a small ring of recent
+sketches, and a hamming-distance compare. Pushing that gate down into
+the sensor MCU's Layer 4 (On-device feature extraction) lets the node
+*not transmit* a frame the cluster-Pi would have filed under
+"familiar" anyway. Bandwidth, sensor-MCU TX energy, and RF airtime
+all win, and the cluster-Pi novelty path stops re-doing work the edge
+already proved pointless. This is the natural ADR-085 follow-up
+flagged but deliberately left out of the ADR-085 scope because it
+requires a `no_std` sketch port, a Kconfig-gated rollout, a wire-
+format bump, and a fresh witness regeneration — none of which are
+appropriate inside an in-flight cluster-Pi work loop.
+
+The crux of the decision is whether the cost of (a) hand-porting the
+sketch primitive to `no_std` Xtensa LX7, (b) sizing the in-IRAM ring
+without disturbing the existing Layer 4 budget, (c) bumping the
+`rv_feature_state_t` magic and teaching the cluster-Pi a graceful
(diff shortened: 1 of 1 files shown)


ruvnet:
## Why this matters, in plain language

This enables **edge reflexes**.

Instead of sending every sensor frame upward and asking the bigger system *"is this interesting?"*, each small device can decide locally:

> "This looks familiar. Don't waste radio, power, or compute."

or:

> "This looks different. Send it now."

That is the practical shift.

### Before

Every node sends feature packets constantly:

```
sensor → cluster Pi → novelty check → maybe ignore
```

The current firmware emits one `rv_feature_state_t` packet every 100 to 1000 ms, defaulting around 5 Hz, even when the room is quiet and nothing meaningful has changed. Across 12 nodes that creates a steady stream the Pi mostly classifies as familiar.

### After

The sensor MCU does a tiny RaBitQ-style sketch locally:

```
sensor → sketch → compare to recent sketches → send only if novel
```

So the node suppresses boring frames and only transmits when the signal changes enough to matter. It still reports how many frames it suppressed, so the system knows silence was intentional, not failure.

### What this gives you

| Benefit | Simple meaning |
|---|---|
| Lower power | Fewer WiFi transmissions |
| Less mesh traffic | Cleaner airtime for important messages |
| Lower Pi load | Pi processes fewer boring packets |
| Faster anomaly response | Novel frames stand out immediately |
| Better logs | Suppression count becomes a signal |
| Safer rollout | Feature is default-off behind Kconfig |

### Why the numbers matter

Radio transmission is expensive on tiny devices. At **50% suppression** a node drops from ~5 packets/sec to ~2.5 packets/sec; at **90% suppression** it drops to ~0.5 packets/sec. The ADR's validation targets are at least **30% UDP TX-energy reduction** in quiet rooms and **50% bandwidth reduction** per node.

### The real product meaning

This turns the sensor node from a dumb streamer into a local decision point.

Not full intelligence.

A **reflex**.

```
normal     → stay quiet
different  → transmit
stuck silent → force-send within 10 seconds
```

That last part matters. The suppression cap prevents the device from going silent forever. The ADR sets the default `CONFIG_RV_EDGE_MAX_CONSEC_SUPPRESS = 50`, which at 5 Hz forces a send within ~10 seconds.

### Bottom line

This enables a mesh of devices that is mostly quiet, low power, and low noise *until something changes*.

That is exactly how an always-on sensing system should behave. It does not process everything deeply. It **watches cheaply, escalates selectively, and proves when it stayed silent**.

PR #436, opened 2026-04-26
feat(nvsim): full simulator stack — Rust crate, dashboard, server, App Store, Ghost Murmur [ADR-089/090/091/092/093]

ruvnet:
## Summary

Ships the full **nvsim** stack: a deterministic NV-diamond magnetometer pipeline simulator (Rust crate), a Vite + Lit dashboard frontend with WASM transport, an Axum-based `nvsim-server` for WS transport, an App Store catalog with live runtime engine, the Ghost Murmur research view, and the GitHub Pages deployment workflow.

Live: **https://ruvnet.github.io/RuView/nvsim/**

64 files / 19,882 lines / 25 commits / 5 ADRs.

### `nvsim` crate — `v2/crates/nvsim` (ADR-089 Accepted)

Deterministic forward simulator: scene → Biot–Savart → material attenuation → NV ensemble → 16-bit ADC + lock-in → fixed-layout `MagFrame` records → SHA-256 witness.

- 6-pass build per `docs/research/quantum-sensing/15-nvsim-implementation-plan.md`
- 50/50 native tests passing
- ~4.5 M samples/s on x86_64 (4500× over Cortex-A53 1 kHz acceptance gate)
- Pinned reference witness: `cc8de9b01b0ff5bd97a6c17848a3f156c174ea7589d0888164a441584ec593b4`
- WASM-ready by construction: zero `std::time/fs/env/process/thread`
- WASM bindings under `wasm` feature flag — `WasmPipeline`, `runTransient`, `referenceWitness`

### `nvsim-server` crate — `v2/crates/nvsim-server` (ADR-092 §6.2)

Axum host fronting `nvsim::Pipeline`:

- 15 REST routes (health, scene, config, seed, run, pause, reset, step, witness, export-proof)
- Binary WebSocket frame stream at `/ws/stream` (32-frame `MagFrame` batches)
- `/api/witness/verify` runs canonical `Proof::generate` so witness matches across transports
- CORS configurable, single-binary deployment
- Builds clean: `cargo check -p nvsim-server`

### Dashboard — `dashboard/` (Vite + TS + Lit)

- 17 Lit components (rail / topbar / sidebar / scene / inspector / console / app-store / ghost-murmur / home / help / onboarding / modal / palette / debug-hud / settings / toast)
- WASM transport via Web Worker; live B-vector trace at ~1.79 kHz
- App Store: 66 cards (65 wasm-edge + nvsim) with runtime pills (running / simulated / mesh-only)
[...]

PR #439, opened 2026-04-27
fix(ci): wasm-pack PATH + Dockerfile workspace stub

ruvnet:
Two post-merge CI failures from #436:

1. **nvsim Dashboard → GitHub Pages** — `wasm-pack: command not found`. `cargo install wasm-pack` doesn't reliably leave the binary on PATH. Switched to the canonical installer (`curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh`) which deposits to `~/.cargo/bin` (already on PATH) in both dashboard-pages.yml and dashboard-a11y.yml.

2. **nvsim-server → ghcr.io** — cargo can't resolve workspace.dependencies because the Dockerfile copies only two crates without the workspace root manifest. New Dockerfile generates a stub workspace Cargo.toml inline that lists just nvsim + nvsim-server with the workspace-deps section copied verbatim.

Both fixes shipped together to keep the PR small.


ruvnet:
Superseded by #440 (clean branch from main).

PR #441, opened 2026-04-27
fix(dashboard): settings drawer scrim covers viewport (host transform fix)

ruvnet:
Drawer's :host had `transform: translateX(...)` which makes it the containing block for any fixed-position descendant. The scrim at `position: fixed; inset: 0` therefore only covered the 420 px panel area. Symptoms: page didn't dim when settings opened, click-outside didn't dismiss.

Fix: keep :host as a fixed full-viewport overlay (no transform), move the body into an inner .panel div, transform only that. Now scrim covers the full viewport and outside-clicks dismiss correctly.


richrimsft:
Love it! Let’s go!!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #485, opened 2026-04-28
feat(sensing-server): wire wifi-densepose-vitals 4-stage pipeline

schwarztim:
## Motivation
The sensing server's vital_signs module previously used an in-process FFT heuristic that produced jittery heart-rate readings (varying ±15 bpm minute-to-minute) and inconsistent confidence scores. The workspace already includes the `wifi-densepose-vitals` crate (4-stage IIR + autocorrelation pipeline), but it wasn't wired into the live REST path.

## Design rationale
- **Why a wrapper (`CrateVitalsPipeline`) instead of replacing the heuristic outright?** Confidence-based fallback. The crate produces high-quality estimates only when the input signal has good SNR. Below confidence 0.05 the legacy heuristic outperforms the crate output, so we fall back. This preserves availability under poor signal conditions.
- **Why threshold 0.05?** Empirically calibrated against the test fixture set in `wifi-densepose-vitals/tests/`. Confidence below 0.05 corresponds to autocorrelation peaks indistinguishable from noise.
- **Why call per-frame instead of batched?** The legacy path is also per-frame; matching the call cadence simplifies the swap and avoids buffer-management changes.

## Changes
- `v2/crates/wifi-densepose-sensing-server/src/vital_signs.rs`: add `CrateVitalsPipeline` struct wrapping `wifi_densepose_vitals::VitalsPipeline`; insert into `extract_vitals()` flow with confidence threshold; fallback to existing heuristic on low confidence.
- `v2/crates/wifi-densepose-sensing-server/src/types.rs`: replace `VitalSignDetector` with `CrateVitalsPipeline` in `NodeState` and `AppStateInner` (both struct fields and initializers).
- `v2/crates/wifi-densepose-sensing-server/Cargo.toml`: add `wifi-densepose-vitals` workspace dep.

## Test results
- `cargo test --workspace --no-default-features`: 1632 passed, 0 failed
- `python verify.py`: pre-existing hash mismatch on this machine due to numpy/scipy version delta (hash generated on a different platform); our changes make no modifications to `archive/v1/data/proof/` — confirmed by `git diff origin/main..HEAD -- archive/v1/data/proof/`

## ADR / docs
- References ADR-045 (vitals jitter investigation); design rationale above is the implementation decision record for this wrapper.
- No new ADR needed; the `wifi-densepose-vitals` crate already has its own design notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   optional Lindblad/Hamiltonian extension if AC magnetometry, MW power
   saturation, hyperfine spectroscopy, or pulsed protocols become required.
 
+### Improved
+- **Vital signs (heart rate + breathing rate) now flow through the 4-stage `wifi-densepose-vitals` pipeline** (IIR + autocorrelation) with confidence-based fallback to the legacy FFT heuristic. When crate confidence \u2265 0.05 the `CrateVitalsPipeline` result is used; below that threshold the existing FFT detector takes over, preserving availability under poor signal conditions. Reduces HR jitter (previously \u00b115 BPM minute-to-minute) and unstable confidence readings.
+
 ### Fixed
 - **Ghost skeletons in live UI with multi-node ESP32 setups** (#420, ADR-082) —
   `tracker_bridge::tracker_to_person_detections` documented itself as filtering
--- a/v2/crates/wifi-densepose-sensing-server/Cargo.toml
+++ b/v2/crates/wifi-densepose-sensing-server/Cargo.toml
@@ -50,5 +50,8 @@ wifi-densepose-wifiscan = { version = "0.3.0", path = "../wifi-densepose-wifisca
 # build without vcpkg/openblas (issue #366, #415).
 wifi-densepose-signal = { version = "0.3.0", path = "../wifi-densepose-signal", default-features = false }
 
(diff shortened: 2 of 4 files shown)


ruvnet:
Hi @schwarztim — current `main` has moved (mergeable=DIRTY/UNKNOWN). Key landings since this PR opened that you'll want to pull in:

- **#595** — `vital_signs.rs` now uses circular variance for wrapped atan2 phase values (relevant if you touch `phase_var` or the heartbeat FFT buffer)
- **#590** — fixes for the Fuzz Testing + Swarm Test (ADR-062) CI jobs (`esp_stubs.h` adds `wifi_ps_type_t`/`WIFI_PS_NONE`, `scripts/qemu_swarm.py` adds `--force-partial`); also corrected firmware README flash offset `0x10000` → `0x20000`
- **#580** — DNS-rebinding host-header allowlist on the sensing-server HTTP + WS routers (new `host_validation.rs` module wired into both routers)
- **#484** — `RecvError::Lagged` now continues instead of disconnecting (WebSocket broadcast handler)
- **#523** — `v2/rust-toolchain.toml` pins Rust 1.89; `--ui-path` default changed from `../../ui` to `../ui` with `coalesce_ui_path()` fallback
- **#606** — `scripts/fix-markers.json` regression guard expanded — your PR shouldn't trip any markers, but the JSON's there to anchor the fixes above

The PRs were all evaluated thoughtful and well-described. Once you rebase against `main`, I'll do a focused review and merge if the test plan still passes locally. If you'd like me to rebase one of these on your behalf, say which (I can push back to your fork since `maintainerCanModify` is enabled).

PR #508, opened 2026-05-06
fix(#505,#506): firmware v0.6.4 version CI gate + no-CSI-signal UI indicator

### LINKED ISSUE #505: v0.6.3-esp32 release ships a binary that internally identifies as 0.6.2
## Summary
The [v0.6.3-esp32 release](https://github.com/ruvnet/RuView/releases/tag/v0.6.3-esp32) asset `cognitum-esp32s3-v0.6.3.bin` has the right filename but its embedded `app_desc` says `version: 0.6.2, date: Apr 28 2026`. Boot logs from any ESP32 running this firmware will report `App version: 0.6.2` and `main: ESP32-S3 CSI Node (ADR-018) — v0.6.2 — Node ID: N`, regardless of what the release was tagged.

## Why now
Surfaced during 2026-05-05 fleet testing. Boot logs and external monitoring of the deployed fleet still say "0.6.2" two weeks after the v0.6.3 release was rolled out, which made me initially think the deployment had been rolled back. Verified by SHA-comparing the running ESP32's flash dump against the release asset:

- ESP32 ota_0 SHA: `2a69d3bc464a768f9f6f33f436b6a56d501e6df414c5a6663ed78efacd3490e8`
- `cognitum-esp32s3-v0.6.3.bin` SHA: `2a69d3bc464a768f9f6f33f436b6a56d501e6df414c5a6663ed78efacd3490e8` ✓ identical

So the binary is the v0.6.3 release. It just identifies as 0.6.2 internally.

## Done looks like
- [ ] `version.txt` is bumped before tagging (the CMake build reads `PROJECT_VER` from this file via `file(STRINGS ... PROJECT_VER LIMIT_COUNT 1)`)
- [ ] Optional: a CI guard that fails the release workflow if the tag name doesn't match `version.txt`

## Work items
- [ ] One-line fix in the release workflow: `echo "0.6.3" > version.txt && git commit -am "chore: bump to 0.6.3" && git tag v0.6.3-esp32`
- [ ] (Optional) Re-tag the existing v0.6.3-esp32 release with a fresh build, OR cut a v0.6.4-esp32 with the version string corrected

---

### Repro
```
$ strings cognitum-esp32s3-v0.6.3.bin | grep -E "^0\\.6\\."
0.6.2

$ python3 -c '
import struct
data = open("cognitum-esp32s3-v0.6.3.bin","rb").read()
i = data.find(struct.pack("<I", 0xABCD5432))   # ESP-IDF app_desc magic
print("version:", data[i+16:i+48].split(b"\\x00",1)[0].decode())
print("date:   ", data[i+96:i+112].split(b"\\x00",1)[0].decode())
'
version: 0.6.2
date:    Apr 28 2026
```

The MGMT-only SPI-cache crash fix (RuView#396) IS in this firmware regardless — that's confirmed by the boot log line `csi_collector: Promiscuous mode enabled (MGMT-only, RuView#396)`. So nothing functional needs to change; just the version-string accounting.

ruvnet:
## Summary

- **Fixes #505** — Firmware version string mismatch: bumps `version.txt` to `0.6.4` and adds a CI step that extracts the embedded `esp_app_desc_t` version from the built binary and fails the workflow if it doesn't match `version.txt`. Prevents silent version drift between code and flash.
- **Fixes #506** — Stale skeleton persists after ESP32 unplugged: adds `csi_live: bool` to `PipelineOutput` (true iff a CSI frame was received within the last 5 seconds), exposes it via `/api/splats`, and shows a red **NO CSI SIGNAL** banner in the viewer while clearing the skeleton when the signal is lost.

## Changes

| File | Change |
|------|--------|
| `firmware/esp32-csi-node/version.txt` | `0.6.2` → `0.6.4` |
| `.github/workflows/firmware-ci.yml` | Add version-string verification step (both 8MB + 4MB matrix variants) |
| `v2/crates/wifi-densepose-pointcloud/src/csi_pipeline.rs` | Add `last_csi_received: Option<Instant>` + `csi_live: bool` in `PipelineOutput` |
| `v2/crates/wifi-densepose-pointcloud/src/stream.rs` | Expose `csi_live` in `/api/splats` JSON response |
| `v2/crates/wifi-densepose-pointcloud/src/viewer.html` | Red "NO CSI SIGNAL" overlay banner + `clearSkeleton()` when signal lost |

## Test plan

- [x] `cargo test --no-default-features` — 15 passed, 0 failed
- [x] `cargo check` — clean
- [ ] Firmware CI builds both 8MB + 4MB variants with version gate
- [ ] Simulate no-signal: disconnect ESP32, confirm banner appears within 5s and skeleton clears
- [ ] Reconnect ESP32, confirm banner disappears and skeleton resumes
- [ ] After merge: cut `v0.6.4-esp32` tag to trigger firmware release CI

## After merge

```bash
git tag v0.6.4-esp32
git push origin v0.6.4-esp32
```

This triggers the firmware CI to build and upload correctly-versioned binaries as release artifacts.


proffesor-for-testing:
The version-string CI gate is a clean fix for #505 — extracting `esp_app_desc_t.version` from the built artifact and failing the workflow when it diverges from `version.txt` is exactly the right shape, and it'll catch this class of bug for every future release.

One thing the gate **won't** catch, though: in the current `v0.6.3-esp32` release there's a separate problem that's orthogonal to the label drift — the 4mb-variant binary (`cognitum-esp32s3-4mb-v0.6.3.bin`, elf `090ef62b9…`) has a Tmr Svc stack overflow that the standalone variant (`cognitum-esp32s3-v0.6.3.bin`, elf `8e62329d…`) does not. Same partition table layout, same advertised version, runtime regression in one of the two builds. Detail and repro in #505: https://github.com/ruvnet/RuView/issues/505#issuecomment-4395933071.

Today on a freshly-flashed ESP32-S3 N16R8 board:
- 4mb variant → reboot loop every ~5s, crash at first `adaptive_ctrl medium tick`
- Standalone → 0 crashes in multi-minute soak, CSI streaming clean to a remote aggregator

If you'd like, the v0.6.4 release that'll come out of this PR could fold in one of:
- **Drop** the `-4mb-` asset from the release until the regression is investigated (lowest-risk option for users picking by filename).
- **Rebuild** the 4MB target with the same Tmr Svc stack tuning as the standalone build — reuses the existing `firmware-ci.yml` matrix you already added in this PR, just needs whichever sdkconfig knob differs to converge.
- Leave both, but **rename** the 4mb asset so it's unambiguously "4MB-flash boards only" rather than reading like an alternative variant of the same binary.

Happy to test whichever build comes out of the next CI run against the same hardware where I reproduced the crash today.

🤖 Generated with Ruflo & AQE

ruvnet:
Closing — mixed-scope PR that has been partially superseded:

**#505 (firmware version CI gate):** main now has a `version-guard` job at `.github/workflows/firmware-ci.yml:19` that compares the release-tag X.Y.Z to `firmware/esp32-csi-node/version.txt`. A fix-marker (`RuView#505` in `scripts/fix-markers.json`) prevents that gate from being removed. This PR's binary-extraction approach is complementary (catches a different class of regression — version.txt updated but build forgot to embed it) but would need to be reapplied against the current workflow shape.

**version.txt 0.6.2 → 0.6.4:** main is already at `0.6.4` (likely via a separate release commit), so this hunk is a no-op now.

**#506 (no-CSI-signal UI indicator):** the `csi_pipeline.rs` + `stream.rs` + `viewer.html` changes in the pointcloud crate are still genuinely useful and not on main. Worth resubmitting as its own focused PR — `fix(pointcloud): no-CSI-signal UI indicator (#506)` — without the version-guard / version.txt churn.

Want me to extract the #506 hunk to a clean follow-up PR?

PR #518, opened 2026-05-09
fix: improve sensing accuracy, presence detection, and UI reliability

KarthikV17:
## Server-side fixes (sensing-server)

- Fix ADR-018 packet parsing: correct byte offsets for 20-byte header
  - n_subcarriers: u8 → u16 at bytes [6..7]
  - freq_mhz: u16 → u32 at bytes [8..11]
  - RSSI at byte 16, noise floor at byte 17
- Tune presence detection for faster response:
  - DEBOUNCE_FRAMES: 4 → 2 (faster state transitions)
  - MOTION_EMA_ALPHA: 0.15 → 0.25 (more responsive)
- Add asymmetric person count smoothing:
  - α=0.35 decay when someone leaves (~2-3s response)
  - α=0.20 rise when someone enters (~3-4s response)
  - Prevents false jumps while staying responsive

## Firmware fixes (esp32-csi-node)

- Read node_id from NVS config instead of compile-time CONFIG_CSI_NODE_ID
  - Enables unique node IDs per device via provisioning
  - Updated: csi_collector.c, edge_processing.c, wasm_runtime.c, display_ui.c
- Add extern declaration for g_nvs_config in nvs_config.h
- Update provision.py to support both esp32s3 and esp32c6 chips

## UI fixes (Observatory)

- Add WebSocket auto-reconnect with exponential backoff:
  - Retries up to 5 times (1s, 2s, 4s, 8s, 16s delays)
  - Shows "RECONNECTING" status with blue pulsing indicator
  - Only falls back to demo mode after all retries exhausted
- Prevents random switches from Live to Demo mode

## Tools

- Add csi_inspector.py for packet debugging with correct ADR-018 offsets

### DIFF
--- a/firmware/esp32-csi-node/main/csi_collector.c
+++ b/firmware/esp32-csi-node/main/csi_collector.c
@@ -17,6 +17,7 @@
 #include "nvs_config.h"
 
 #include <string.h>
+#include <stdbool.h>
 #include "esp_log.h"
 #include "esp_wifi.h"
 #include "esp_timer.h"
@@ -104,8 +105,9 @@ size_t csi_serialize_frame(const wifi_csi_info_t *info, uint8_t *buf, size_t buf
     uint32_t magic = CSI_MAGIC;
     memcpy(&buf[0], &magic, 4);
 
-    /* Node ID (from NVS config) */
-    buf[4] = g_nvs_config.node_id;
+    /* Node ID (cached at boot via csi_collector_set_node_id to survive
+     * WiFi-init g_nvs_config corruption — issues #232/#390). */
+    buf[4] = csi_collector_get_node_id();
 
     /* Number of antennas */
     buf[5] = n_antennas;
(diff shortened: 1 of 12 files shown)


KarthikV17:
@ruvnet please review and let me know if there is any additional testing needed. This has been verified with ESP32-S3.

ruvnet:
Thanks for the multi-area contribution — there's good intent here but the scope is too broad to land as one PR, and several pieces need verification against current main:

## Server-side
- **ADR-018 packet offsets** — current `csi.rs::parse_esp32_frame` and the firmware `csi_serialize_frame` agree on a 20-byte header with `n_subcarriers u16 @ [6..7]`, `freq_mhz u32 @ [8..11]`, `rssi i8 @ 16`, `noise_floor i8 @ 17` — same as your PR claims. If you're seeing offset misalignment on current main, please file an issue with the offending packet hex dump so we can trace the divergence.
- **Presence tuning + asymmetric smoothing** — these are operator-tunable knobs. Worth its own focused PR with an A/B comparison on the same test fixtures so we can see what the latency-vs-stability tradeoff costs.

## Firmware
- **`node_id` from NVS** — already in current firmware via fix-marker `RuView#232-375-385-386-390` ("defensive early-capture of NVS config before `wifi_init_sta()` corrupts it"). Please rebase and verify the NVS read is still missing somewhere.
- **`esp32c6` support in `provision.py`** — overlap with #522 / #524 (krish2718) which just landed ESP-IDF 6 + ESP32-C6 build support including `provision.py` chip auto-detect. The `--chip auto` default is now in main.

## UI
- **WebSocket auto-reconnect** — this is genuinely useful and not currently in main. Worth its own focused PR — `fix(ui): websocket auto-reconnect with exponential backoff` — without the firmware / server changes.

Closing this one as superseded-or-needs-split. Strongest single follow-up: the WebSocket auto-reconnect UX, on its own, against current main. Happy to review that immediately.

PR #522, opened 2026-05-10
firmware/esp32-csi-node: IDF 6 build, HE CSI config, unicore DSP, provision chip detect

krish2718:
## Summary

Bring `firmware/esp32-csi-node` in line with ESP-IDF 6+ and Wi-Fi 6 (HE) targets
(e.g. ESP32-C6): build fixes, correct CSI configuration layout, safe task pinning
on unicore SoCs, and provisioning that does not assume ESP32-S3.

## Changes

- **IDF 6 / mbed TLS 4:** Use PSA crypto for SHA-256 in `rvf_parser` instead of the
  removed public `mbedtls/sha256.h` API; set explicit `REQUIRES` in
  `main/CMakeLists.txt` so IDF 6 component resolution does not treat the component
  as dependency-free.
- **Wi-Fi 6 CSI:** Initialize `wifi_csi_config_t` using the HE acquire bitfield path
  when `CONFIG_SOC_WIFI_HE_SUPPORT` is set, matching `wifi_csi_acquire_config_t`
  vs legacy boolean-style fields.
- **Unicore:** Pin the edge DSP worker with `portNUM_PROCESSORS` so core `1` is
  never used on single-core chips.
- **Provisioning:** Default esptool chip to `auto` (optional override); use
  `write-flash` spelling expected by current esptool.

## Test plan

- [x] `idf.py set-target esp32c6` (or your HE target), `idf.py build`
- [x] Flash, connect STA, confirm CSI path runs without assert on unicore
- [x] `provision.py` against a non-S3 device without hardcoding `--chip esp32s3`

## Notes

Intended for **ESP-IDF 6+** where the SHA-256 header and component `REQUIRES`
behavior changed; older IDF may still build but was not the focus of this series.

### DIFF
--- a/firmware/esp32-csi-node/main/CMakeLists.txt
+++ b/firmware/esp32-csi-node/main/CMakeLists.txt
@@ -11,7 +11,26 @@ set(SRCS
     "adaptive_controller.c"
 )
 
-set(REQUIRES "")
+# ESP-IDF v6+: headers must resolve via explicit REQUIRES (no implicit deps).
+set(REQUIRES
+    esp_wifi
+    esp_netif
+    esp_event
+    nvs_flash
+    app_update
+    esp_http_server
+    esp_http_client
+    esp_app_format
+    esp_timer
+    esp_pm
+    esp_driver_uart
+    esp_driver_gpio
+    esp_driver_spi
+    esp_driver_i2c
+    driver
+    lwip
+    mbedtls
+)
 
 # ADR-061: Mock CSI generator for QEMU testing + ADR-081 mock radio binding
 if(CONFIG_CSI_MOCK_ENABLED)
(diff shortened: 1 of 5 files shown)


ruvnet (approved):
Forward-compat work for ESP-IDF 6 + Wi-Fi 6 (HE) targets, defensive and well-guarded:

- **`#if CONFIG_SOC_WIFI_HE_SUPPORT`** correctly fences the new `wifi_csi_acquire_config_t` bitfield layout — ESP32-S3 (no HE) keeps the legacy boolean-style `wifi_csi_config_t`, so no regression on the canonical 8MB target.
- **CMakeLists explicit `REQUIRES`** is harmless under IDF 5.4 (those components already resolve implicitly) and required under IDF 6 (no more implicit deps). The conditional `wasm3` requirement is correct since WASM is compile-time optional.
- **PSA crypto for SHA-256** in rvf_parser — when this lands I'll want to confirm IDF 5.4 still builds (which uses the public `mbedtls/sha256.h`), but if the build is conditional on `MBEDTLS_VERSION_MAJOR >= 4` it's fine. Going to trust the test plan `[x]` boxes here.
- **`portNUM_PROCESSORS` for task pinning** on unicore — defensive against future ESP32-C3 builds, has no effect on dual-core S3.
- **`provision.py` chip auto-detect** — overdue ergonomics fix for the C6 target.

Merging. If the IDF 5.4 build breaks on the PSA SHA-256 change, will guard or revert that hunk in a follow-up.

PR #523, opened 2026-05-10
v2: pin Rust 1.89 and fix sensing-server UI path when run from v2

krish2718:
## Summary

Make the sensing server and its static UI usable from a normal `v2/` checkout
without hand-editing paths, and pin a Rust toolchain that matches the workspace
dependency graph (edition 2024 / newer APIs).

## Changes

- Add `v2/rust-toolchain.toml` pinning **Rust 1.89** (required by the current
  `mmap-rs` / edition-2024 chain and related crates in this workspace).
- **Sensing server:** Default UI static path to `../ui` (repo layout: `ui/` next
  to `v2/`), and coalesce fallbacks (`../ui`, `./ui`, `../../ui`) when the default
  is missing so `cargo run` from `v2/` does not serve an empty `/ui`.

## Test plan

- [ ] `cd v2 && cargo build -p wifi-densepose-sensing-server`
- [ ] Run the binary from `v2/` with default args; open the dashboard and confirm
  assets load (no blank UI).

## Notes

Toolchain pin is **repo-wide for `v2/`** via `rust-toolchain.toml`, not only the
sensing server crate; the PR title calls out sensing-server because that is the
user-visible workflow fixed alongside UI path resolution.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/cli.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/cli.rs
@@ -19,8 +19,8 @@ pub struct Args {
     #[arg(long, default_value = "5005")]
     pub udp_port: u16,
 
-    /// Path to UI static files
-    #[arg(long, default_value = "../../ui")]
+    /// Path to UI static files (from `v2/` cwd use `../ui`)
+    #[arg(long, default_value = "../ui")]
     pub ui_path: PathBuf,
 
     /// Tick interval in milliseconds (default 100 ms = 10 fps for smooth pose animation)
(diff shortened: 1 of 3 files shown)


ruvnet (approved):
Three small, safe, on-target changes:

1. **`v2/rust-toolchain.toml` pinning Rust 1.89** — explicitly states what the existing edition-2024 / `mmap-rs` chain already requires implicitly. Eliminates the 'wrong rustc' first-time-build failure.
2. **`--ui-path` default `../ui`** — assumes `cd v2 && cargo run`, which matches `CLAUDE.md`'s build-test commands. The prior `../../ui` worked only from `v2/crates/wifi-densepose-sensing-server/`.
3. **`coalesce_ui_path()` fallback** — gracefully recovers from a wrong cwd by trying the three common layouts, logging a `warn` so the operator knows. Right-sized: not magic, but stops a blank-UI bug from being a head-scratcher.

Merging.

PR #524, opened 2026-05-10
firmware/esp32-hello-world: ESP32-C6 target and ESP-IDF v6 build fixes

krish2718:
- Default sdkconfig.defaults to esp32c6
- Fix removed SOC_* macros for ESP-IDF v6; probe_peripherals split for S3 vs C6.
- Banner and WiFi/BLE/power strings are target-aware; add CHIP_ESP32C6 name.
- Ignore esp32-hello-world/sdkconfig.old from idf.py set-target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ firmware/esp32-csi-node/managed_components/
 firmware/esp32-csi-node/dependencies.lock
 firmware/esp32-csi-node/sdkconfig.defaults.bak
 
+# ESP-IDF set-target backup (local only)
+firmware/esp32-hello-world/sdkconfig.old
+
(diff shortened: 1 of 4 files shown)


krish2718:
Intentionally didn't push the update to `sdkconfig`, ideally it should be untracked as it will be aut-generated as part of build, I can push a commit to remove that?

ruvnet (approved):
Pairs with #522 — same forward-compat shape for the hello-world demo. The SOC_* macro probes were removed in IDF 6 and the target-aware banner is the right ergonomics for someone trying ESP32-C6 for the first time. `sdkconfig.old` ignore is housekeeping. Merging.

PR #570, opened 2026-05-15
fix: middleware classes inherit BaseHTTPMiddleware to fix 500 errors …

DavidKrame:
## Problem
Both RateLimitMiddleware and AuthenticationMiddleware were not inheriting 
from BaseHTTPMiddleware. Starlette's add_middleware() injects the `app` 
argument automatically, but the __init__ methods only accepted `settings`, 
causing:

  TypeError: RateLimitMiddleware.__init__() got multiple values for argument 'settings'

This made every single API endpoint return a 500 error immediately.

## Fix
- Made both classes inherit from BaseHTTPMiddleware
- Added `app` as first parameter to __init__ and called super().__init__(app)
- Renamed __call__ to dispatch (required by BaseHTTPMiddleware)

## Files changed
- archive/v1/src/middleware/rate_limit.py
- archive/v1/src/middleware/auth.py

## Tested on
- macOS, Python 3.13, FastAPI latest

### DIFF
--- a/archive/v1/src/middleware/auth.py
+++ b/archive/v1/src/middleware/auth.py
@@ -9,6 +9,7 @@ from datetime import datetime, timedelta
 
 from fastapi import Request, Response, HTTPException, status
 from fastapi.security import HTTPBearer, HTTPAuthorizationCredentials
+from starlette.middleware.base import BaseHTTPMiddleware
 from jose import JWTError, jwt
 from passlib.context import CryptContext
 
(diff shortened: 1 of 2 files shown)


ruvnet (approved):
Reviewed. The fix is mechanically correct: FastAPI middleware must inherit `BaseHTTPMiddleware` and implement `dispatch(request, call_next)` instead of `__call__` — the previous shape (`async def __call__(self, request, call_next)` on a plain class) is the documented cause of the 500 you'd see when Starlette tries to register the middleware.

Caveat: these files live under `archive/v1/` — v1 is no longer the production codebase (v2 Rust crates serve all the live HTTP/WS surfaces). The middlewares here are effectively dormant; archive code is preserved primarily for the deterministic proof pipeline (`./verify` against `archive/v1/data/proof/`).

Merging anyway because:
1. The fix is correct and the diff is minimal (4 lines net per file).
2. Anyone who clones the archive and tries to run v1 will hit exactly the bug you fixed.
3. It removes a footgun for future readers studying the archived code.

Thanks for the contribution.

Issue #519, opened 2026-05-09
Ghost person detection, FPS infinity, skeleton flickering and jumpy vitals with ESP32-S3 multi-node setup (edge-tier 1 & 2)


kapilsoni2013:
## Environment
- Hardware: 3x ESP32-S3-WROOM (16MB Flash, 8MB PSRAM)
- Firmware: v0.6.4-esp32
- Server: ruvnet/wifi-densepose Docker (latest)
- Host OS: macOS (MacBook Pro)
- WiFi channel: 3 (locked via --channel 3)
- Edge tier: 1 and 2 (both tested)
- TDM: enabled (--tdm-slot 0/1/2 --tdm-total 3)
- pres-thresh: 15, vital-win: 30, vital-int: 2, subk-count: 16
- Docker env: CSI_SOURCE=esp32, MAX_FPS=10, MAX_PERSONS=1

### Bug 1 — Ghost person: 2 persons detected with only 1 person present
Occurs with 1 ESP32 and with 3 ESP32 nodes. Persists on both edge-tier 1 and edge-tier 2. MAX_PERSONS=1 Docker env var does not resolve it. Suspected cause: subcarrier group clustering splitting single person's reflections into 2 clusters (as noted in issue #104).

### Bug 2 — FPS shows "Infinity" in UI with 3 ESP32 nodes
MAX_FPS=10 Docker env var fixes it with 1 node but not with 3 nodes. With 3 nodes active the frame flood overwhelms the rate limiter and FPS counter overflows to Infinity.

### Bug 3 — Pose skeleton flickering too fast
Skeleton keypoints jump rapidly between positions making it impossible to track posture. Occurs on both edge-tier 1 and 2 with 3 nodes. TDM provisioning did not resolve this.

### Bug 4 — Vitals (BR, HR) readings are very jumpy
Values change erratically every second even with --vital-win 30 and --vital-int 2. Not stable enough for real use. Tested with single stationary subject within 2m of ESP32.

## Steps to reproduce

1. Flash v0.6.4 firmware (8MB variant) to 3x ESP32-S3
2. Provision each node separately (one at a time, unplugged others):
   ```
   python provision.py --port /dev/cu.usbmodem1101 \
     --ssid "SSID" --password "PWD" --target-ip 192.168.1.10 \
     --edge-tier 1 --channel 3 --node-id 1 \
     --tdm-slot 0 --tdm-total 3 \
[...]

PR #595, opened 2026-05-16
fix(vital_signs): use circular variance for wrapped phases

### LINKED ISSUE #593: vital_signs.rs: linear stats on wrapped (atan2) phase values
process_frame in vital_signs.rs uses arithmetic mean + arithmetic variance on phase values, but the phases come from `q_val.atan2(i_val)` (csi.rs:90, main.rs:855) so they're wrapped to (-pi, pi]. When two subcarriers happen to land on opposite sides of the wrap (e.g. 3.14 and -3.14, physically ~0.003 rad apart), the linear formula treats them as ~2*pi apart and you get a variance of ~pi^2 instead of ~1e-6. That value goes straight into the heart-rate FFT buffer.

Repro on a fresh clone:

```bash
git clone --depth 1 https://github.com/ruvnet/RuView.git && cd RuView
sed -n '142,148p' v2/crates/wifi-densepose-sensing-server/src/vital_signs.rs
python3 -c "
import math
phases = [math.pi - 0.001, -math.pi + 0.001]
m = sum(phases) / 2
print('linear:  ', sum((p - m) ** 2 for p in phases) / 2)
s = sum(math.sin(p) for p in phases); c = sum(math.cos(p) for p in phases)
print('circular:', 1 - (s * s + c * c) ** 0.5 / 2)
"
```

You'll see `linear: 9.86...` vs `circular: 5e-07` on the same input.

Probably the cause of the very jumpy vitals reported in #519, and the +/-15 bpm jitter mentioned in #485's description. #485 routes around the buggy detector with an external pipeline but keeps VitalSignDetector as the low-confidence fallback, so the math is still live in that path.

Fix is the standard circular variance (1 - R, mean resultant length). PR coming.

akhilesharora:
The heartbeat feature in process_frame was computed as arithmetic mean + arithmetic variance on phases that come from atan2() and are wrapped to (-pi, pi]. Two phases that straddle +/-pi are physically aligned but the linear formula treats them as ~2*pi apart, so you get ~pi^2 (~9.87) instead of ~1e-6. That noise rides into the heart-rate FFT buffer every time a subcarrier's phase crosses the wrap.

Swapped the inline math for a standard circular variance (1 - R, where R is the mean resultant length). The value is bounded [0, 1] instead of [0, ~pi^2] now, but the heartbeat FFT cares about relative spectral peaks, not absolute scale.

Tests:

```
cargo test -p wifi-densepose-sensing-server --no-default-features --lib vital_signs
test result: ok. 15 passed; 0 failed
```

4 new tests: one drives `process_frame` end-to-end and inspects the heartbeat_buffer; the other three hit the helper with wraparound, opposite, and identical inputs.

Related to the jumpy vitals in #519 and the +/-15 bpm jitter mentioned in #485. #485 wraps VitalSignDetector with an external pipeline but keeps it as the low-confidence fallback - this fix is complementary.

Closes #593

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/vital_signs.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/vital_signs.rs
@@ -139,13 +139,15 @@ impl VitalSignDetector {
         // Cardiac-induced body surface displacement is < 0.5 mm, producing
         // tiny phase changes. Cross-subcarrier phase variance captures this
         // more sensitively than amplitude alone.
+        //
+        // Phases come from atan2() and are wrapped to (-pi, pi]. Linear mean
+        // and variance on wrapped values is wrong: two phases close across
+        // the +/-pi discontinuity (e.g. pi-eps and -pi+eps) are physically
+        // ~2*eps rad apart but produce arithmetic variance ~pi^2. Use the
+        // standard circular variance (1 - mean resultant length), which is
+        // stable across the wrap.
         let phase_var = if phase.len() > 1 {
-            let mean_phase: f64 = phase.iter().sum::<f64>() / phase.len() as f64;
-            phase
-                .iter()
-                .map(|p| (p - mean_phase).powi(2))
-                .sum::<f64>()
-                / phase.len() as f64
+            phase_circular_variance(phase)
         } else {
             // Fallback: use amplitude high-pass residual when phase is unavailable
             let half = amplitude.len() / 2;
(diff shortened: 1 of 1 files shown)


ruvnet (approved):
Reviewed end-to-end:

- Math is correct: `1 - R` where `R = sqrt(sin_sum^2 + cos_sum^2) / n` is the standard Mardia circular variance.
- Bounded to `[0, 1]` instead of `[0, ~pi^2]`, which is fine because the heartbeat FFT cares about relative spectral peaks (confirmed in the PR description).
- The regression test `test_phase_variance_handles_wraparound` is particularly nice — it asserts both the new bound AND that the *old* buggy formula on the same input produces ~pi^2, which makes a regression to the linear computation explicitly fail.
- `test_process_frame_handles_wrapped_phases` is the end-to-end check that anyone can run against main to observe the bug and against this branch to observe the fix. Excellent reproducer.
- Identical-phase and opposite-phase edge cases covered.
- Scoped to a single file. Well-commented. Closes #593.

Approving. Merging with squash.

Issue #561, opened 2026-05-14
ESP32 CSI firmware README has inconsistent flash offset and provisioning path


Fewmanism:
While preparing a real ESP32-S3 bring-up from a fresh clone, I found two documentation inconsistencies in `firmware/esp32-csi-node/README.md`:

1. Quick Start / Flashing sections show app flashed at `0x10000`:

```text
0x10000 firmware/esp32-csi-node/build/esp32-csi-node.bin
```

But the repo partition tables define `ota_0` at `0x20000`:

```text
partitions_display.csv: ota_0 app ota_0 0x20000 ...
partitions_4mb.csv:     ota_0 app ota_0 0x20000 ...
```

The QEMU section also uses `0x20000`. The release binaries also include `ota_data_initial.bin`, suggesting the manual flash command should likely flash app at `0x20000` and include OTA data at `0xf000`.

2. Quick Start says:

```bash
python scripts/provision.py ...
```

but the file in this repo is:

```text
firmware/esp32-csi-node/provision.py
```

These mismatches make first hardware bring-up easy to misflash or fail from the documented commands.

ruvnet:
Thanks @Fewmanism — both inconsistencies fixed in #562.

1. Flash offset `0x10000` → `0x20000` in both occurrences (matches `partitions_display.csv` / `partitions_4mb.csv` which put `ota_0` at `0x20000`). Also added the missing `0xf000 ota_data_initial.bin` line — release bundles ship this binary and without it the bootloader can refuse to boot after a factory wipe.
2. `scripts/provision.py` → `firmware/esp32-csi-node/provision.py`. There are actually two `provision.py` files in the repo right now (a stale 275-line copy in `scripts/` and the canonical 348-line one in `firmware/esp32-csi-node/` that has the #391 full-replace semantics fix); README now points at the canonical one. The stale duplicate is a separate cleanup.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

ruvnet:
Fixed in PR #590 — flash offset corrected to 0x20000 (matches partitions_display.csv and partitions_4mb.csv), ota_data_initial.bin added at 0xf000, and provision.py path corrected to firmware/esp32-csi-node/provision.py (the canonical script — scripts/provision.py is an older copy).

PR #600, opened 2026-05-17
Add single-ESP32S3 demo setup for presence detection

Rudraa-25:
Simplified setup for running RuView with just ONE ESP32-S3 board:
- Basic presence detection (room occupancy)
- Motion detection
- Distance estimation

For full pose estimation (17 keypoints), breathing, and heart rate, 2+ nodes are recommended.

### DIFF
--- a/demo-single-esp32s3/README.md
+++ b/demo-single-esp32s3/README.md
@@ -1,0 +1,122 @@
+# RuView Single ESP32-S3 Demo
+
+A simplified setup for running RuView with just ONE ESP32-S3 board for basic presence detection.
+
+> **Note:** Single-node detection has limited spatial resolution. For full pose estimation (17 keypoints), use 2+ nodes. This demo focuses on **presence detection** - detecting if someone is in the room.
+
+## What You Need
+
+| Item | Cost | Notes |
+|------|------|-------|
+| ESP32-S3 DevKit (8MB flash) | ~$9 | Any ESP32-S3 with 8MB flash works |
+| Micro USB cable | ~$5 | For power + programming |
+| WiFi network | - | 2.4GHz WiFi required |
+
+## Quick Start
+
+### Step 1: Get Pre-built Firmware
+
+Download the latest firmware from releases:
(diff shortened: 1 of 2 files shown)


ruvnet:
Thanks for the contribution — the single-node entry point is a real gap. A few issues to fix before this can land:

1. **Flash offset 0x10000 is wrong.** The repo's partition tables (`partitions_display.csv`, `partitions_4mb.csv`) put `ota_0` at `0x20000`. The README at `firmware/esp32-csi-node/README.md` was just corrected in #590 (closes #561). Please use:
   ```
   0x0     bootloader.bin
   0x8000  partition-table.bin
   0xf000  ota_data_initial.bin
   0x20000 esp32-csi-node.bin
   ```

2. **Release URL points at your fork.** `https://github.com/Rudraa-25/RuView/releases` should be `https://github.com/ruvnet/RuView/releases`. Once a tagged firmware release exists upstream, the binaries the contributor expects (`bootloader.bin`, `partition-table.bin`, `esp32-csi-node.bin`, `ota_data_initial.bin`) are downloadable from there.

3. **`--source single` flag.** I don't see this flag in `wifi-densepose-sensing-server`'s CLI. The closest existing one is `--source esp32` or `--source auto`. Please verify the command runs end-to-end before merging.

4. **`requirements.txt` for the sensing server.** The sensing server is a Rust crate (`cargo run -p wifi-densepose-sensing-server`) — no Python deps needed. The `pip install -r requirements.txt` step should be removed.

5. **Asset paths reference `release/`** (singular). The repo uses `firmware/esp32-csi-node/release_bins/` (plural).

6. **Location.** Per `CLAUDE.md`, example code lives under `/examples`. Please move `demo-single-esp32s3/` → `examples/single-esp32s3/`.

7. **`check_esp32.py` location.** Utility scripts live under `/scripts` per `CLAUDE.md`. If the script is specific to this example, keeping it inside `examples/single-esp32s3/` is fine — just call that out in the README.

The intent is good and there's clearly a real audience for a single-node entry point. Happy to re-review once these are addressed.

Issue #574, opened 2026-05-15
esp32-csi-node: mDNS discovery for seed_url + provisioning UX fixes


proffesor-for-testing:
## Summary

`esp32-csi-node`'s `swarm_bridge` feature requires hand-set `seed_url` and `seed_token` via `provision.py` writes to the `csi_cfg` NVS namespace. Customers must know their Cognitum Seed's IP, run a Python tool, navigate `provision.py`'s full-replace footgun (issue #391), and re-provision when the seed's DHCP lease changes. mDNS discovery would eliminate the manual step entirely for the common "ESP32 + Cognitum Seed on the same WiFi" case.

## Why now

Customer onboarding friction. A real customer spent 4+ hours on what should have been a "flash and go" ESP32 ↔ Seed integration. The provision.py footgun and hand-set seed_url were two of the larger time sinks. mDNS + a USB-CDC pairing handshake should reduce this to under five minutes per node.

A separate seed-side issue exists for the cog fan-out gap that made the data invisible even after Brian got it flowing — `cognitum-one/seed#166` (proposal pending @ruvnet review). This RuView-side discovery work is independent of that seed-side architectural decision and can ship on its own.

## Done looks like

- ESP32 with empty `seed_url` periodically queries `_cognitum-seed._tcp.local` (or similar) via mDNS while on WiFi
- On match, optionally fetches `seed_token` from the seed's pairing endpoint over USB-CDC if the seed is plugged in via USB at first-boot, falls back to operator-paste-from-CLI otherwise
- Empty-by-default semantics preserved — explicit `seed_url` from `provision.py` still wins
- DHCP lease changes don't break the link (re-discovery on POST failure)
- `provision.py` no longer wipes the `csi_cfg` namespace on partial reconfigure (#391)

## Work items

- [ ] mDNS query in new `seed_discovery.c` (or extend `nvs_config.c`) when `seed_url` is empty at boot
- [ ] Re-discovery loop on `swarm_bridge` POST failure (handle DHCP lease churn)
- [ ] Optional: USB-CDC pairing flow that fetches `seed_token` without `provision.py`
- [ ] `provision.py` additive-by-default mode (fix #391 footgun) so partial reconfigures don't wipe other NVS keys
- [ ] Update `firmware/esp32-csi-node/README.md` quick-start with the discovery path

## Status

Net-new feature work, no breaking changes proposed. Empty-default behaviour preserved so existing deployments are unaffected.

---

### Current state

[...]

PR #617, opened 2026-05-18
fix: recognize swarm provisioning values

NgoQuocViet2001:
## Summary

- Treat `--hop-channels`, `--seed-token`, `--swarm-hb`, and `--swarm-ingest` as real provisioning values in `provision.py`'s empty-config guard.
- Move that guard into a small `has_config_value()` helper so the CLI value check can be regression-tested.
- Add Python tests covering swarm bridge / channel hopping flags and CSV output.

## Why

Issue #574 calls out the `esp32-csi-node` swarm bridge provisioning path, including seed credentials and heartbeat/ingest settings. The script already writes these fields to NVS, but the preflight guard did not count several of them as config values. A partial invocation such as `--seed-token ... --force-partial` could therefore fail with `At least one config value must be specified` before CSV generation.

This keeps the existing full-replace safety guard intact; it only fixes the false "no config" rejection for fields that are already supported by `build_nvs_csv()`.

## Validation

- `python -m unittest discover -s firmware\esp32-csi-node\tests -p "test_*.py"`
- `python -m py_compile firmware\esp32-csi-node\provision.py firmware\esp32-csi-node\tests\test_provision.py`
- `git diff --check`

## Duplicate check

Searched existing PRs for `provision.py`, `csi_cfg`, `full-replace`, and `seed_url` overlap; did not find an open or recent PR covering this guard bug.

Refs #574

### DIFF
--- a/firmware/esp32-csi-node/provision.py
+++ b/firmware/esp32-csi-node/provision.py
@@ -37,6 +37,39 @@ NVS_PARTITION_OFFSET = 0x9000
 NVS_PARTITION_SIZE = 0x6000  # 24 KiB
 
 
+CONFIG_VALUE_CHECKS = [
+    ("ssid", bool),
+    ("password", lambda value: value is not None),
+    ("target_ip", bool),
+    ("target_port", lambda value: value is not None),
+    ("node_id", lambda value: value is not None),
+    ("tdm_slot", lambda value: value is not None),
+    ("tdm_total", lambda value: value is not None),
+    ("edge_tier", lambda value: value is not None),
+    ("pres_thresh", lambda value: value is not None),
+    ("fall_thresh", lambda value: value is not None),
+    ("vital_win", lambda value: value is not None),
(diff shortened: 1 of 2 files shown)


ruvnet (approved):
Real fix + cleaner shape:

- **The bug**: `provision.py` with ONLY swarm/hopping flags (e.g. `--hop-channels 1,6,11` or `--seed-token …`) was refused by the `if not has_value` guard, because those flags were missing from the inline `any([...])` list. The CSV builder happily handled them; the upstream check just blocked the path.
- **The fix**: declarative `CONFIG_VALUE_CHECKS` list of `(field_name, predicate)` tuples + `has_config_value(args)` helper that walks them. Adding a new NVS-backed flag is now a one-line edit, not a two-place update.
- **Tests included** exercising the rejection envelope (operational flags alone don't count, swarm/hopping flags do count, CSV writes them correctly).

Merging.

PR #619, opened 2026-05-18
fix(sensing-server): route WS source through effective_source() (closes #618)

### LINKED ISSUE #618: WebSocket sensing broadcast hardcodes source="esp32", bypasses 5s stale-detection (effective_source not called)
## Summary

The WebSocket `/ws/sensing` broadcast continues to emit `source: "esp32"` indefinitely after the ESP32 hardware loses power or network connectivity. The UI ("Sensing" tab) then keeps showing **"LIVE — ESP32 HARDWARE Connected"** with cached/frozen sensor values, with no indication that the data source is offline.

`AppStateInner::effective_source()` already implements the correct 5-second stale-detection (returns `"esp32:offline"` when no UDP frame has arrived within `ESP32_OFFLINE_TIMEOUT`), but it is **only invoked from REST endpoints (`/health`, etc.), never from the WS broadcast path**.

## Reproduce

1. Flash v0.6.4-esp32 firmware, provision WiFi + `--target-ip`.
2. Start `sensing-server --source esp32 --bind-addr 0.0.0.0`.
3. Open `http://localhost:8080/ui/index.html` → "Sensing" tab. Verify the green "LIVE — ESP32 HARDWARE" banner and live `tick` / amplitude data.
4. Unplug the ESP32 (or power it off).
5. Wait 30+ seconds.

### Observed
- UI still displays "LIVE — ESP32 HARDWARE Connected".
- WS payload still emits `"source": "esp32"`.
- `tick` field frozen at the last received value.
- All `nodes[].amplitude`, `features.*`, `vital_signs.*` frozen at the last frame, but re-broadcast every tick.
- `classification` keeps emitting whatever motion/presence the cached frame implied.

### Expected
- After 5 seconds with no UDP frames, WS payload should emit `"source": "esp32:offline"` so the UI can switch to an offline/disconnected state (the same way the REST `/health` endpoint already reports `"status": "degraded"`).

## Verification

`GET /health` correctly reports the offline state:
```json
{"clients":1,"source":"esp32:offline","status":"degraded","tick":44680}
```

WS `/ws/sensing` does not — it keeps emitting:
```json
{"type":"sensing_update","source":"esp32","tick":44540, ...same frozen payload...}
```

## Root Cause

`effective_source()` (main.rs:679) is correct, and is called from REST endpoints (main.rs:2167, 2735, 2789, 2810, 2820, 2844, 3472).

The two WS broadcast sites use a **string literal** instead:

```rust
// main.rs:3791 (edge-vitals path)
let mut update = SensingUpdate {
    msg_type: "sensing_update".to_string(),
    timestamp: chrono::Utc::now().timestamp_millis() as f64 / 1000.0,
    source: "esp32".to_string(),   // <-- bypasses effective_source()
    tick,
    ...

// main.rs:4018 (raw CSI path)
let mut update = SensingUpdate {
    msg_type: "sensing_update".to_string(),
    timestamp: chrono::Utc::now().timestamp_millis() as f64 / 1000.0,
    source: "esp32".to_string(),   // <-- same bug
    tick,
    ...
```

## Suggested Fix

Both call sites already hold the `AppStateInner` guard `s` in scope:

```diff
-    source: "esp32".to_string(),
+    source: s.effective_source(),
```

(Single-line change at each site. No new locks, no allocation churn beyond what `effective_source` already does.)

## Why This Matters (Safety)

The project README and various ADRs market use-cases including:
- Fall detection / elder-care monitoring
- Overnight vital-sign tracking (apnea screening)
- Presence-based safety triggers

Silently re-publishing the last received frame as "LIVE" is a **silent-failure pattern**: a deployed system whose ESP32 lost power could continue reporting "breathing normal / present" indefinitely, masking a real emergency. The 5s timeout was clearly intended to prevent exactly this — the fix just needs to extend coverage to the WS path.

## Related / Wider Scope

This patch is the **minimum** fix and is sufficient to let the UI flip to an offline indicator. A more complete fix could additionally:

- Suppress WS broadcasts entirely when stale (skip the tick loop when `effective_source().ends_with(":offline")`), or
- Add a `stale: bool` field on `SensingUpdate` that downstream consumers (UI, recorders, cluster aggregators) can branch on.

The two existing per-node `NodeFeatureSnapshot.stale` flags (main.rs:478, 511-512) are a precedent for the second approach.

#519 ("Ghost person detection, FPS infinity, skeleton flickering and jumpy vitals with ESP32-S3 multi-node setup") may be partially related — both stem from the broadcast loop not knowing the upstream sensor state has degraded.

## Environment

- **Firmware**: v0.6.4-esp32 (`esp32-csi-node.bin` SHA256 prefix `0066d74d35b0dbca…`)
- **Server**: `main` branch
- **Hardware**: ESP32-S3 N16R8 (Waveshare DEV-KIT), single-node, channel 6 / 2.4 GHz
- **Host**: Windows 11, sensing-server built with stable-x86_64-pc-windows-gnu, `--no-default-features`
- **UI**: Chrome on `localhost:8080/ui/index.html`

ArnonEnbar:
Closes #618.

## Summary

The two WebSocket broadcast sites in the sensing-server tick loop hardcoded `source: "esp32".to_string()`, bypassing the existing `AppStateInner::effective_source()` stale-detection logic. As a result, the UI continued to show "LIVE — ESP32 HARDWARE Connected" with frozen cached values for arbitrarily long after the ESP32 lost power or WiFi.

This PR replaces both literals with `s.effective_source()` so the WS path emits `"esp32:offline"` once `ESP32_OFFLINE_TIMEOUT` (5 s) elapses, matching the behaviour the REST `/health` endpoint already had.

## Diff

```diff
-                        source: "esp32".to_string(),
+                        source: s.effective_source(),
```

Applied at the two existing call sites in `v2/crates/wifi-densepose-sensing-server/src/main.rs`. The `s` AppStateInner mutable guard is already held at both points (acquired upstream as `let mut s = state.write().await;`), so no new locks or restructure is needed.

### Before — bug

Before

WS broadcast loop emits hardcoded `source: "esp32"` every 100 ms tick, regardless of UDP frame freshness. UI cannot distinguish live vs cached state.

### After — fix

After

WS broadcast now routes through `effective_source()`, which returns `"esp32:offline"` when `last_esp32_frame.elapsed() > ESP32_OFFLINE_TIMEOUT`. UI can flip to a degraded-state indicator.

Source `.drawio` file is included for future editing: `docs/diagrams/issue-618/source-stale-detection.drawio`.

## Why this matters

The project markets fall-detection, vital-sign monitoring, and overnight apnea screening use-cases. Silent stale-as-live is a safety-critical failure mode — a deployed system whose ESP32 lost power could continue reporting "present, breathing normal" indefinitely, masking a real emergency. The 5 s offline timeout was clearly designed to prevent exactly this, but only covered REST consumers.

## Test plan

- [x] Build with `cargo build -p wifi-densepose-sensing-server --no-default-features` (clean, 1 pre-existing unused-imports warning)
- [x] Manual repro per #618: ESP32 unplugged → after 5 s, `/health` reports `esp32:offline` AND now WS `/ws/sensing` payload `source` also flips to `esp32:offline`.
- [ ] **No unit/integration test added**. `effective_source()` is a single `if` chain, the diff is literal-to-method swap. Adding a test would require either:
  - extracting a pure helper from `effective_source()` (refactor outside this PR's scope), or
  - constructing a full `AppStateInner` with all ~80 fields (no `Default` impl).

  Happy to add either approach if a reviewer prefers — let me know which direction. Otherwise, the manual reproduction in #618 is the verification.

## Build notes

Compiled on Windows with `stable-x86_64-pc-windows-gnu` (project default `rust-toolchain.toml` pins MSVC; switched locally via `rustup override` because MSVC build tools weren't installed). Fix is pure-Rust and target-agnostic, so this shouldn't matter for CI.

## Wider scope (not addressed here)

This fix lets the UI **know** the data is stale; it does not prevent the WS broadcast loop from continuing to emit the cached frame each tick. A more thorough fix could:

- Suppress WS broadcasts when `effective_source().ends_with(":offline")`, or
- Add a `stale: bool` field on `SensingUpdate` (precedent: per-node `NodeFeatureSnapshot.stale` at main.rs:478).

Kept this PR minimal so it can land independently; happy to follow up.


ruvnet:
Closing — same diagnosis, same call-site, same fix already shipped via #621 (merged 2026-05-17 23:16Z). My version cloned the cached `latest_update`, overwrote `source` with `s.effective_source()`, and serialised the result before broadcast — the change you proposed here.

Your bug report (#618) is what made it possible to ship the fix in 30 minutes from filed-to-merged. Thank you for the precise root-cause trace — `effective_source()` called from REST paths but not WS was exactly the gap, and the issue body was specific enough that no diagnosis was needed.

If there's other behaviour on the WS broadcast path that still feels off (e.g., `vital_signs`/`features` should also be cleared when `source == 'esp32:offline'` rather than re-emitting the cached last-known values), happy to take that as a follow-up issue.

Issue #136, opened 2026-03-04
Docker error on MacOS


bcardarella:
```
➜  ~ docker pull ruvnet/wifi-densepose:latest
Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found
```

ruvnet:
Fixed! The `latest` tag now includes both `linux/amd64` and `linux/arm64` manifests, so it works on Apple Silicon Macs.

```bash
docker pull ruvnet/wifi-densepose:latest
docker run -p 3000:3000 ruvnet/wifi-densepose:latest
```

Manifest verification:
```
amd64: sha256:08b6e198a773a1743cc593e4d419bc4c7eab4b656cfd9439ccda902d5939626c
arm64: sha256:20e5e44e7c852608a56f5718839ab5142834c6c6f3ed100afa0c8fb2279e9ad8
```

Also updated README and user guide with multi-arch documentation and troubleshooting.

hopye:
im getting this error : 
[entrypoint] ERROR: refusing to start sensing-server with default
[entrypoint]        posture: RUVIEW_API_TOKEN is unset AND bind is
[entrypoint]        0.0.0.0. /ws/sensing streams live sensing
[entrypoint]        frames; that data would be readable by anyone who
[entrypoint]        can reach this host. Pick one:
[entrypoint]          docker run -e RUVIEW_API_TOKEN=$(openssl rand -hex 32) ...
[entrypoint]          docker run -e RUVIEW_BIND_ADDR=127.0.0.1 ...
[entrypoint]          docker run -e RUVIEW_ALLOW_UNAUTHENTICATED=1 ...   # only on trusted network
[entrypoint]        See https://github.com/ruvnet/RuView/issues/864

PR #631, opened 2026-05-19
fix(ci): build multi-arch wifi-densepose image (linux/arm64 was missing, closes #625)

### LINKED ISSUE #625: macOS docker issue
I finally got this [ESP32](https://www.amazon.com/3-Pack-Gold-Kit-Expansion-Terminals/dp/B0FL149DGM?crid=39C6XCIF5BNKW&dib=eyJ2IjoiMSJ9.twd0IrTlfmd6eqwnYnsOltDD5gvb0ss36HJ6G26r_PucSKFh_UvgqTnkdnv9T9d8sciit6dXwk3545ncQ6ayyaKwmgWAm5AGQJQaTN6H9ObTB8luocx-ocjpjMkWHzS9SZH0csdoSXlny0bDrt_5Ivm80EQcnt8kb0CSdYO41FLiZKKFHZb2IoRXJSGDq34cgcucqBPjCky_7ODEaj_-mWQMdT2tQiXYr1MiEv8zymg.9r2wDWWmfF4dK0fO1juPpyMnWzSCLxV7GUow5JDgNbQ&dib_tag=se&keywords=ESP32-S3&qid=1775271764&sprefix=esp32-s3%2Caps%2C399&sr=8-4&th=1)

I am trying to set it up on a macbook m3 pro. I tried the easiest route. But I am getting the following problem. 

macOS: Tahoe 26.4.1 (25E253)

```
docker pull ruvnet/wifi-densepose:latest
Error response from daemon: no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found
```

**What is the easiest way to set this up on mac with three ESP32?**

MackDing:
Fixes #625. (Same crash class as the closed-unmerged #136.)

## Bug

`README.md:44` advertises **`Docker: multi-arch amd64 + arm64`**, and PR #547 ("refresh Docker publish") promised multi-arch as part of the `:latest` rebuild plan, but `.github/workflows/sensing-server-docker.yml:100` only sets:

```yaml
platforms: linux/amd64
```

So `ruvnet/wifi-densepose:latest` on Docker Hub today is amd64-only:

```
$ curl -s https://hub.docker.com/v2/repositories/ruvnet/wifi-densepose/tags/latest/
last_updated: 2026-05-14T17:35:04Z
images:
  arch=amd64    os=linux       size=34509901
  arch=unknown  os=unknown     size=1496      # attestation layer, not arm64
```

Reproduction (from #625, Mac M3 Pro / macOS Tahoe 26.4.1):

```
$ docker pull ruvnet/wifi-densepose:latest
Error response from daemon: no matching manifest for linux/arm64/v8 in the
manifest list entries: no match for platform in manifest: not found
```

## Fix

Standard buildx multi-arch recipe — 2 small changes to `sensing-server-docker.yml`:

| Change | Why |
|---|---|
| Add `docker/setup-qemu-action@v3` before `setup-buildx-action@v3` | QEMU user-mode emulation lets the `ubuntu-latest` (amd64) runner cross-build the `linux/arm64` layer. Without it, buildx refuses unknown platforms. |
| `platforms: linux/amd64` → `platforms: linux/amd64,linux/arm64` | Actually publish the arm64 layer the README + #547 promised. |

`docker/Dockerfile.rust` needs no changes — verified arch-agnostic:

- No `--target` flag on `cargo build`
- No `target.x86_64-*` / `target.aarch64-*` sections in any `Cargo.toml`
- Only native build-deps is `cc = "1.0"` (already cross-aware)

Smoke tests are unaffected: they `docker pull` on `ubuntu-latest`, which auto-selects the amd64 entry from the multi-arch manifest. Multi-arch manifests are transparent to single-arch consumers.

## Diff

+11 / -1, one file:

```diff
+      - uses: docker/setup-qemu-action@v3
+
       - uses: docker/setup-buildx-action@v3
...
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
```

## Alternatives considered

- **Native arm64 runner (`ubuntu-24.04-arm`)** instead of QEMU emulation. Faster (~3-5× for a Rust workspace this size) but requires a `strategy.matrix` over `{runs-on, platforms}` + a `docker/manifest-action` join step. ~50 lines of churn vs. 11. Happy to do this in a follow-up if build time on arm64 becomes painful — but the cheap one-line fix unblocks #625 today.
- **Cross-build with `cross` / `--target aarch64-unknown-linux-gnu`** in the Dockerfile. Would avoid QEMU runtime emulation but means restructuring the Rust build stage. Bigger blast radius, scope-creep for what is a workflow-config bug.

## Scope

Only touches `.github/workflows/sensing-server-docker.yml` — the workflow #625 is about. `nvsim-server-docker.yml:59` has the identical `platforms: linux/amd64` bug (different image, separate concern); happy to file a follow-up PR for that one if useful — kept out here for review-clarity.

## One observation, not part of this fix

The last 5 runs of `sensing-server-docker.yml` have failed at the **`Log in to Docker Hub`** step (looks like `DOCKERHUB_TOKEN` rotated/expired). Once that secret-side issue is resolved, the next push to `main` will produce the first true `amd64 + arm64` manifest — and #625 / #136 will close. Pointing this out so the green-CI signal on this PR doesn't get misread as "fix works"; the only way to fully verify the merged behaviour is the post-merge push.

Happy to iterate on commit message, scope, or the QEMU-vs-native-arm-runner question.

### DIFF
--- a/.github/workflows/sensing-server-docker.yml
+++ b/.github/workflows/sensing-server-docker.yml
@@ -50,6 +50,12 @@ jobs:
         with:
           submodules: recursive
 
+      # QEMU is required so the amd64 GitHub runner can cross-build the
+      # linux/arm64 layer below (Dockerfile.rust is arch-agnostic — no `--target`
+      # flag — so buildx + QEMU is all that's needed; arm64 builds are emulated
+      # by the runner, not built on a separate arm64 host).
+      - uses: docker/setup-qemu-action@v3
+
       - uses: docker/setup-buildx-action@v3
 
       - name: Log in to Docker Hub
@@ -90,7 +96,11 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          platforms: linux/amd64
+          # README badge advertises `amd64 + arm64`, and #547 promised multi-arch
+          # as part of the docker publish refresh; arm64 was never actually wired
+          # in, so Apple Silicon Macs hit `no matching manifest for linux/arm64/v8`
+          # on `docker pull ruvnet/wifi-densepose:latest` (#136, #625). Build both.
+          platforms: linux/amd64,linux/arm64
 
       # ---------------------------------------------------------------------
       # Smoke-test the freshly-pushed image:


ruvnet (approved):
Right diagnosis — README badge advertises arm64 + Apple Silicon users hit `no matching manifest for linux/arm64/v8`. QEMU setup + `linux/amd64,linux/arm64` platforms in buildx is the standard fix and the only thing needed since `Dockerfile.rust` is arch-agnostic (no `--target` flag, no platform-specific RUN). Closes #625 + the older #136. Merging.

PR #624, opened 2026-05-18
fix(sensing): finish #611 NaN-panic audit — 7 sites missed by #613

therahul-yo:
## Summary

Follow-up to #613 (which closed #611). #613 fixed `adaptive_classifier.rs:94` and explicitly stated the audit was complete. That audit grepped for the literal pattern `partial_cmp(b).unwrap()` and missed **seven additional production sites** that use comparator variants — all share the same crash class as #611 (NaN in `f64` data → `partial_cmp` returns `None` → `unwrap()` panics the sensing server).

## Sites fixed

| File | Line | Function | Why it matters |
|------|------|----------|----------------|
| `v2/.../adaptive_classifier.rs` | 205 | `AdaptiveModel::classify()` argmax over softmax probs | **Same per-frame hot path as #611.** A NaN in `raw_features` survives normalise (`NaN / x = NaN`) → logits → softmax (`exp(NaN)/exp(NaN) = NaN`), and the #613 IQR fix does not gate this branch. |
| `v2/.../adaptive_classifier.rs` | 480, 500 | `train_from_recordings()` argmax | Training accuracy / per-class accuracy loops. |
| `v2/.../main.rs` | 2446, 2449 | `count_persons_mincut` variance source/sink select | The outer `.unwrap_or((0, &0))` only catches an empty iterator — it cannot rescue a panic raised inside the comparator. A single NaN in `variances[]` aborts the process. |
| `v2/.../csi.rs` | 602, 605 | same logic, duplicated in `csi.rs` | Same as above. |

## Fix

Drop-in: swap `.unwrap()` for `.unwrap_or(std::cmp::Ordering::Equal)` inside the comparator closure. This is the same one-line behavioural change as #613 and matches the pattern the same files already use elsewhere (e.g. `adaptive_classifier.rs:149-150, 155`, and now `:94` from #613).

No API change, no test changes needed — the existing fix(sensing) commit precedent in CHANGELOG describes this as a behavioural fix only.

## Re-audit of remaining `partial_cmp(...).unwrap()` sites

After this PR, the only remaining `partial_cmp(…).unwrap()` matches in `v2/` are inside `#[cfg(test)]` / `#[test]` blocks where inputs are controlled and panic-on-NaN is acceptable:

- `wifi-densepose-signal/src/spectrogram.rs:269` — `#[test] fn test_single_frequency_peak`
- `wifi-densepose-pointcloud/src/depth.rs:234` — backproject test
[...]

Issue #611, opened 2026-05-17
adaptive_classifier.rs:94 — partial_cmp(b).unwrap() panics if NaN values enter the classifier


matad0r-maghribi:
# Bug: `partial_cmp(b).unwrap()` panics if NaN values enter the classifier

## File
`v2/crates/wifi-densepose-sensing-server/src/adaptive_classifier.rs:94`

## Description

```rust
sorted.sort_by(|a, b| a.partial_cmp(b).unwrap());
```

The `.unwrap()` on `partial_cmp` **panics** if the slice contains a single `NaN` value, because `f64::partial_cmp(&NaN)` returns `None`.

## Impact

- CSI data from real ESP32 hardware can produce NaN (silent computation errors, division by zero in DSP, empty buffers)
- **Production code** called on every frame in the `classify()` path
- A single NaN frame kills the entire sensing server process

## Fix

```rust
sorted.sort_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal));
```

## Notes

The same file already handles NaN correctly elsewhere (lines 149-150, 155 use `.unwrap_or(Ordering::Equal)`), making this an oversight.

## Severity

**Critical**  runtime panic on real-world data, process crash.
@bannned-bit

PR #771, opened 2026-05-23
fix(sensing): keep adaptive classifier confidence finite

NgoQuocViet2001:
## Summary

- sanitize non-finite adaptive classifier inputs during normalization by falling back to the trained feature mean
- ignore non-finite probabilities during argmax and fall back to a zero-confidence class if all probabilities are invalid
- add a regression test covering `NaN`/`Infinity` runtime features so classifier confidence remains finite

## Why

#613 / #624 closed the direct `partial_cmp(...).unwrap()` panic path, but a bad CSI hardware sample can still flow through normalization and softmax as `NaN`. That leaves `AdaptiveModel::classify()` returning a non-finite confidence, which can leak into API responses, UI state, or telemetry on every frame after a noisy sample. This keeps the classifier output safe without changing normal finite-input behavior.

## Duplicate check

Searched existing PRs for `adaptive_classifier NaN confidence`, `finite confidence sensing`, and the previous #611 / #613 / #624 NaN audit. I found the merged panic fixes, but no open/current PR that clamps classifier confidence or adds this regression coverage.

## Validation

- Ran: `cargo fmt -p wifi-densepose-sensing-server -- --check`
- Ran: `git diff --check`
- Ran: `cargo test -p wifi-densepose-sensing-server classify_returns_finite_confidence_for_non_finite_features --bin sensing-server`

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/adaptive_classifier.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/adaptive_classifier.rs
@@ -222,6 +222,14 @@ fn default_class_names() -> Vec<String> {
     DEFAULT_CLASSES.iter().map(|s| s.to_string()).collect()
 }
 
+fn finite_or(value: f64, fallback: f64) -> f64 {
+    if value.is_finite() {
+        value
+    } else {
+        fallback
+    }
+}
+
 impl Default for AdaptiveModel {
     fn default() -> Self {
         let n_classes = DEFAULT_CLASSES.len();
@@ -249,7 +257,11 @@ impl AdaptiveModel {
         // Normalise features.
         let mut x = [0.0f64; N_FEATURES];
         for i in 0..N_FEATURES {
-            x[i] = (raw_features[i] - self.global_mean[i]) / (self.global_std[i] + 1e-9);
+            let mean = finite_or(self.global_mean[i], 0.0);
+            let std = finite_or(self.global_std[i], 1.0);
+            let denom = finite_or(std + 1e-9, 1.0);
+            let feature = finite_or(raw_features[i], mean);
+            x[i] = finite_or((feature - mean) / denom, 0.0);
         }
 
         // Compute logits: w·x + b for each class.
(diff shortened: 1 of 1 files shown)


NgoQuocViet2001:
Rechecked this against `main` today rather than just bumping it. Still applies cleanly, and `classify()` is byte-identical to what it was when I opened this — same normalization, same `partial_cmp(...).unwrap_or(Equal)` argmax.

I also want to correct the framing in my own description, because I think I oversold it and that may be part of why it stalled.

**What is real.** Copying `classify()`'s body verbatim from `adaptive_classifier.rs` at `8d1a67d` and feeding it one non-finite feature:

```
feature[0] = NaN        -> confidence = NaN      finite = false
feature[0] = +Infinity  -> confidence = NaN      finite = false
all finite              -> confidence = 0.731059 finite = true
```

An infinite feature is enough on its own, because `0.0 * f64::INFINITY` is `NaN` — a weight the trainer zeroed out does not neutralise the feature, it poisons the whole dot product. From there `max_logit` survives (Rust's `f64::max` ignores NaN), `exp_sum` does not, every prob is NaN, and `max_by(...unwrap_or(Equal))` hands back the first element. So `#611` / `#613` / `#624` fixed the panic and left a silent NaN in `classification.confidence`, which `adaptive_override` then blends into the API response.

**What is not real, and I should have said so.** I could not construct a live input that reaches it. `parse_esp32_frame` builds amplitudes as `(i*i + q*q).sqrt()` from two `i8`s, so they are bounded by ~181 and always finite; `generate_simulated_frame` likewise.

**The part I think is actually worth your time.** The two halves of `features_from_runtime` are protected very unevenly, and only one of them on purpose. Features 0–6 go through `serde_json::json!` in `adaptive_override`, and `json!` maps a non-finite `f64` to `null`:

```
NaN     -> json null   -> as_f64().unwrap_or(0.0) = 0
+inf    -> json null   -> as_f64().unwrap_or(0.0) = 0
finite  -> json 1.5    -> as_f64().unwrap_or(0.0) = 1.5
```

So a NaN `variance` is silently sanitised to `0.0` — correct outcome, accidental mechanism, and nothing in the code says the JSON hop is load-bearing. Features 7–14 come from `amps`, which is handed to `subcarrier_stats` as a plain `&[f64]` straight out of `frame_history` with no such hop. That is the asymmetry this PR closes, and it is the half that would break first if amplitudes ever stop coming from the `i8` parser — a different transport, a firmware that sends floats, or a decoder that divides.

So: defensive hardening on a path that is currently unreachable, not a live crash. 66 lines, one file, behaviour unchanged for finite input. Happy either way — if you would rather not carry it, say so and I will close it myself; if you want it smaller I can drop the `finite_or` normalization and keep only the argmax guard.

PR #773, opened 2026-05-23
fix(ui): unbreak viz.html — OrbitControls importmap, WS URL, toast NPE (#760)

### LINKED ISSUE #760: Sensing UI fails to load — 3 bugs block all live data visualization
The sensing server receives ESP32 data perfectly, but the web UI is completely broken by three separate bugs. None of the tabs, buttons, or visualizations work. You just see a blank or half-loaded page.

all in `main` branch

## **Bug 1:** `3D visualization crashes with "THREE.OrbitControls is not a constructor"`
Three.js r160 removed the old-style OrbitControls script. The viz.html page loads it the old way, so it crashes instantly. No 3D view loads at all.

## **Bug 2:** `WebSocket connects to wrong port and wrong URL`
viz.html hardcodes ws://localhost:8000/ws/pose. The server uses port 3001 for WebSocket (when HTTP is 3000) and the endpoint is /ws/sensing, not /ws/pose. Result: no live data ever reaches the UI, it stays stuck in "MOCK DATA - DEMO MODE" forever.

## **Bug 3:** `Toast notification crashes the entire app`
When the backend is alive, the app tries to show a "backend connected" toast. But toast.js calls appendChild on a null container because init() was never called first. This throws a TypeError that kills the rest of the page initialization. After that, nothing works — tabs, buttons, live data, everything is dead.

Image

## **How to reproduce:**
1. Start server: cargo run -p wifi-densepose-sensing-server -- --http-port 3000 --ws-port 3001 --source esp32
2. Open http://localhost:3000/ui/index.html
3. Page loads partially, then crashes. Console shows this.container is null

quick and dirty fix that solved the issue locally on my end
[sensing-ui-fix.patch](https://github.com/user-attachments/files/28163380/sensing-ui-fix.patch)

ruvnet:
## Summary
- Switches Three.js + OrbitControls to importmap-based ES module load (r160 removed `examples/js/`). `THREE` + `THREE.OrbitControls` are re-exposed as globals so the existing component modules don't need a wider refactor.
- Replaces the hardcoded `ws://localhost:8000/ws/pose` with the existing `buildSensingWsUrl()` helper so the page hits the correct `/ws/sensing` endpoint regardless of port pairing (8080/8765 default, 3000/3001 Docker). `?ws=…` query overrides for non-standard deployments.
- Lazy-inits the toast container in `ToastManager.show()` so the "backend connected" notification no longer NPEs and kills the rest of page init (patch from #760 reporter).

## Test plan
- [ ] Start server: `cargo run -p wifi-densepose-sensing-server -- --http-port 3000 --ws-port 3001 --source esp32`
- [ ] Open `http://localhost:3000/ui/viz.html` — 3D scene loads, no OrbitControls error in console
- [ ] WebSocket connects to `/ws/sensing`, demo mode flips to live data when frames arrive
- [ ] Toast notifications render without console errors
- [ ] Default port pairing also works: `--http-port 8080 --ws-port 8765`, open `http://localhost:8080/ui/viz.html`

Closes #760.

### DIFF
--- a/ui/services/websocket-client.js
+++ b/ui/services/websocket-client.js
@@ -1,9 +1,19 @@
 // WebSocket Client for Three.js Visualization - WiFi DensePose
-// Connects to ws://localhost:8000/ws/pose and manages real-time data flow
+// Default endpoint is `/ws/sensing` on the same host the page was served from.
+// Callers (e.g. viz.html) usually pass an explicit `url` derived from
+// `buildSensingWsUrl()` so HTTP/WS port pairings are handled centrally.
+
+function _defaultWsUrl() {
+  if (typeof window === 'undefined' || !window.location) {
+    return 'ws://localhost:8765/ws/sensing';
+  }
+  const protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+  return `${protocol}//${window.location.host}/ws/sensing`;
+}
 
 export class WebSocketClient {
   constructor(options = {}) {
-    this.url = options.url || 'ws://localhost:8000/ws/pose';
+    this.url = options.url || _defaultWsUrl();
     this.ws = null;
     this.state = 'disconnected'; // disconnected, connecting, connected, error
     this.isRealData = false;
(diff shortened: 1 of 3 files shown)


apubowacodo054-del:
omm887275@gmail.com

PR #870, opened 2026-05-30
MQTT publisher::spawn is never callled.

Yesuah:
MQTT publisher is fully implemented but publisher::spawn is never called from main.rs. args.mqtt is never read. Fix adds the channel, wires broadcast_tick_task, and spawns the publisher on startup.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -5300,6 +5300,31 @@ async fn simulated_data_task(state: SharedState, tick_ms: u64) {
         if let Ok(json) = serde_json::to_string(&update) {
             let _ = s.tx.send(json);
         }
+        #[cfg(feature = "mqtt")]
+        if let Some(ref vtx) = s.vitals_tx {
+            if vtx.receiver_count() > 0 {
+                use wifi_densepose_sensing_server::mqtt::state::VitalsSnapshot;
+                let vs = &s.latest_vitals;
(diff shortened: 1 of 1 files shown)


ruvnet:
Thanks for catching this — your diagnosis was correct **at the time this PR was opened**: the MQTT publisher was implemented but not spawned.

Triage update: this is now **resolved on `main`** via the #872 work, implemented with a slightly different (typed) shape, so this PR is superseded:

- `main.rs:6178` — `if args.mqtt_opts.mqtt { … }` reads the flag (the arg moved into a grouped `mqtt_opts`).
- `main.rs:6182` — `mqtt::config::MqttConfig::from_args(&args.mqtt_opts)` + `mcfg.validate()`.
- `main.rs:6195` — `let (vtx, vrx) = broadcast::channel::<mqtt::state::VitalsSnapshot>(64);`
- `main.rs:6196` — `mqtt::publisher::spawn(mcfg, builder, vrx);` ← the spawn you flagged.
- `main.rs:6197+` — bridges the existing JSON sensing broadcast into the typed `VitalsSnapshot` stream the publisher consumes; gated behind `#[cfg(feature = "mqtt")]` (the Docker image builds `--features mqtt`).

Because of that, this PR's diff (which targets the older flat `args.mqtt` / `args.mqtt_host` fields) no longer applies cleanly — those fields are now under `args.mqtt_opts`, so it would need a rebase to even compile, and the behaviour it adds is already present.

Recommend closing as superseded by #872. Appreciate the clear write-up — it matched the root cause exactly.

PR #886, opened 2026-05-31
fix(proof): pin determinism lock to numpy 2.4.2 (match published hash)

ruvnet:
## Problem
`Verify Pipeline Determinism` has been failing — on `main` too (red since at least May 26). CI computes `667eb054…` vs the committed/published `expected_features.sha256 = ca58956c…`.

Root cause: `archive/v1/requirements-lock.txt` pinned **numpy 1.26.4 / scipy 1.14.1**, which produce a different IEEE-754 float result than the **modern numpy 2.x** the published hash was generated with — i.e. what a fresh `pip install numpy`, the maintainers, and the new `docs/proof-of-capabilities.md` skeptic path all use today. So the determinism gate was failing against *its own published proof*.

## Fix
Bump the lock to **numpy 2.4.2 / scipy 1.17.1**. `python archive/v1/data/proof/verify.py` prints `VERDICT: PASS` (Computed == Expected == `ca58956c…`) with these versions.

## Scope / risk
The lock file is consumed **only** by `verify-pipeline.yml`. The passing `Tests (3.10/3.11/3.12)` jobs use `requirements.txt`, so they're unaffected. Zero blast radius beyond the determinism gate.


AntwerpDesignsIonity:
@copilot

Issue #803, opened 2026-05-26
The ESP32-S3 device cannot correctly identify the exact number of people nearby.


jackyzp:
I have tested single and multiple ESP32-S3 devices, launching them via Python, Docker, and Rust respectively. Unfortunately, none of the setups can accurately detect the number of people in the surrounding rooms — the system consistently shows a count of 1 person only.
I would like to confirm whether this feature is not actually implemented in the project, or if there are errors in the startup methods and operating procedures. Do we need to import additional models to make it work properly?
I have also tried generating and modifying code with Claude multiple times, but still failed to achieve real-time crowd detection. If anyone has successfully reproduced the actual people counting function, please share the detailed operation steps. Thank you.

<img width="1639" height="774" alt="Image" src="https://github.com/user-attachments/assets/fee37854-a843-46eb-8b8e-4267391235d9" />

<img width="1604" height="781" alt="Image" src="https://github.com/user-attachments/assets/72ce2317-dd70-48e9-b33b-2a3d6c11832b" />

ruvnet:
You need at least 3 esp32 s3, esp32 c6 work better, one isn't going to work.

jackyzp:
> You need at least 3 esp32 s3, esp32 c6 work better, one isn't going to work.你至少需要 3 个 esp32 s3，esp32 c6 效果更好，一个是不行的。

I have used more than three (specifically four or five), yet the results remain the same with no noticeable changes. For instance, when three devices are deployed, only one or two people are detected. However, as people enter or leave, the count never decreases and remains stuck at one or two. I even suspect there must be an issue somewhere, but I still haven't been able to identify the root cause.

jackyzp:
> You need at least 3 esp32 s3, esp32 c6 work better, one isn't going to work.你至少需要 3 个 esp32 s3，esp32 c6 效果更好，一个是不行的。

Do you have any authentic documents or videos that demonstrate how to actually reproduce and real-time detect the number of people around using the ESP32-S3? I want to follow the instructions and give it a try.

jmmacip:
Hi, same issue here, running on 3 s3 and 1 c6 and there is no way it visualizes more than 1 person in the room

<img width="1848" height="1102" alt="Image" src="https://github.com/user-attachments/assets/4d71a030-97e1-4d3d-bcb3-a86bc1490f6e" />

luis-seedone:
[...]

PR #892, opened 2026-06-01
fix(edge): gate multi-person count on presence — no phantom persons (#803, #496)

markt-heximal:
Addresses #803 and #496.

## Problem
`update_multi_person_vitals()` (edge_processing.c) sets `pv->active = true` **unconditionally** for `n_persons = s_top_k_count / 2` (= 4 at the default `EDGE_TOP_K=8`) every frame. So the reported `n_persons` is a fixed function of the subcarrier count — **never 0**, regardless of how many people are actually present. An empty room reports 4; users see a constant phantom count.

Root cause behind:
- **#803** — "consistently shows 1 person" / "is people-counting even implemented?"
- **#496** — "shows 2 people, but I'm alone in the room"

The firmware already computes presence (`s_presence_detected`, set earlier in the same `process_frame()` pass) — but the multi-person path never consults it.

## Fix
Gate person activation on `s_presence_detected`: when no presence is detected, mark all persons inactive and return → `n_persons = 0` in an empty room; persons activate only once presence crosses the (adaptive) threshold.

```c
if (!s_presence_detected) {
    for (uint8_t p = 0; p < EDGE_MAX_PERSONS; p++) s_persons[p].active = false;
    return;
}
```

## Scope / honest limitations
This **bounds the count to presence** (0 when empty, ≥1 when someone is there). It does **not** attempt true multi-person *separation* — when present, the count is still the subcarrier-derived `top_k/2`. Real counting/separation from a single link is a larger DSP problem; this PR just removes the "always N phantom people" behavior everyone is hitting.

## Testing
- Builds clean — ESP-IDF v5.4, `idf.py set-target esp32s3 && idf.py build`.
- The gating signal was verified on hardware: `presence_score` (= `motion_energy`) / `s_presence_detected` read ~0 at idle and rise with movement, so the gate yields 0 when empty and activates on presence. The pre-fix constant count (4 at `top_k=8`) was observed directly on an ESP32-S3.

### DIFF
--- a/firmware/esp32-csi-node/main/edge_processing.c
+++ b/firmware/esp32-csi-node/main/edge_processing.c
@@ -478,6 +478,18 @@ static void update_multi_person_vitals(const uint8_t *iq_data, uint16_t n_sc,
 {
     if (s_top_k_count < 2) return;
 
+    /* Presence gate (issues #803, #496): only activate persons when presence is
+     * actually detected. Without this, the loop below unconditionally marks
+     * top_k/2 persons active every frame, so the reported n_persons is a fixed
+     * artifact of the subcarrier count (never 0) regardless of occupancy — an
+     * empty room still reports e.g. 4 "people". s_presence_detected is updated
+     * earlier in the same process_frame() pass. (This bounds the count to
+     * presence; true multi-person separation is a separate, larger problem.) */
+    if (!s_presence_detected) {
+        for (uint8_t p = 0; p < EDGE_MAX_PERSONS; p++) s_persons[p].active = false;
+        return;
+    }
+
     /* Determine number of active persons based on available subcarriers. */
     uint8_t n_persons = s_top_k_count / 2;
     if (n_persons > EDGE_MAX_PERSONS) n_persons = EDGE_MAX_PERSONS;


ruvnet:
Independently verified — this fix is **correct**, and the call ordering it relies on checks out against current `main` (`edge_processing.c`):

- `process_frame()` runs `update_top_k()` (L745) → sets `s_presence_detected = (s_presence_score > threshold)` (L817) → calls `update_multi_person_vitals()` (L851).
- This PR's gate sits at the top of `update_multi_person_vitals` (L478), i.e. it executes at L851 — **after** L817 — so `s_presence_detected` is the freshly-updated value for this frame. The comment's "updated earlier in the same process_frame() pass" claim is accurate.

Before this, `update_multi_person_vitals` activated `s_top_k_count / 2` persons whenever `s_top_k_count >= 2` (essentially always), so an empty room reported a fixed `top_k/2` phantom count (the #803/#496/#894 "4–10 people when nobody's there"). Gating activation on presence is the right minimal fix; the existing `for (p = n_persons; p < EDGE_MAX_PERSONS) active = false` tail already handles the partial case, and this extends it to the zero-presence case.

Worth noting this is the **edge/firmware** half of the over-count. The **server** heuristic path has a complementary cap on `main` — `field_bridge.rs` bounds occupancy to `MAX_SINGLE_LINK_OCCUPANCY = 3` (#907). Together they close both layers: the ESP32 no longer emits phantom persons, and the server can't amplify a single link past 3.

LGTM to merge. (Happy to run an ESP-IDF build to confirm it compiles if that's useful before merge — it only touches existing symbols, so it should be clean.)

PR #923, opened 2026-06-03
Fix antenna-regime presence over-read: de-saturate motion_score + correct CSI header offsets

williammalone:
## Summary

Fixes a presence over-read that appeared on an ESP32-S3 CSI mesh after external IPEX antennas were added: a confirmed-empty room read "present" indefinitely. Two independent root-cause bugs in `wifi-densepose-sensing-server`:

### 1. `motion_score` saturation (presence path never migrated off fixed divisors)
`variance_motion` and `mbp_motion` in `extract_features_from_frame` used hardcoded divisors (`/10`, `/25`) tuned for the antenna-less regime. Antennas raised amplitudes ~5× and these amplitude² energies ~30×, pinning **both** terms at the `1.0` clamp. A saturated term carries no information, so `raw_motion` could never fall near the presence floor and the adaptive baseline subtraction in `smooth_and_classify` was defeated.

Fix: normalize both by signal power (`mean_amp²`) — the same dimensionless `sqrt`-of-power-ratio form already used by `temporal_motion_score` in the same function — making `motion_score` **amplitude-scale-invariant**. The dominant `temporal_motion_score` term already did this, which is why only the other two terms saturated. This fixes the single shared `extract_features_from_frame` used by **both** the aggregate and per-node classification paths.

### 2. `parse_esp32_frame` header offsets off by 2 bytes vs firmware
The parser read `sequence`/`rssi`/`noise_floor` two bytes early relative to the firmware layout (`firmware/esp32-csi-node/main/csi_collector.c` → `csi_serialize_frame`: `seq @ [12..16)`, `rssi @ [16]`, `noise_floor @ [17]`). `rssi` was decoded from sequence-counter byte 2 — which stays `0` for the first 65,536 frames — yielding an impossible `rssi = 0 dBm` that zeroed the RSSI fusion weights and the SNR-based `signal_quality`. The I/Q payload at byte 20 (`CSI_HEADER_SIZE`) was already correct, so amplitude-derived features were unaffected.

## Tests
- New regression tests: `motion_score` is amplitude-scale-invariant (1× vs 5×), and a quiet high-amplitude signal does not saturate while a moving one scores higher (dynamic range preserved).
- Full binary unit suite green: **103 passed**.

## Validation (live, 2-node mesh, real CSI)
- RSSI now reports real values (−28…−74 dBm); the `rssi=0` sentinel is gone.
- An empty room now produces genuine low-motion frames (`mbp ~4`), impossible before — the ceiling clamp previously made every frame look like ~180.

## Known residual (tracked separately)
Genuine multi-subcarrier CSI still reads elevated (`mbp ~180`) in an empty room — intrinsic antenna multipath noise where empty-but-noisy and motionless-person overlap. This needs a learned empty-room reference (adaptive-classifier retrain on de-saturated features), not threshold math. These two fixes are the necessary prerequisite for that work.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/csi.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/csi.rs
@@ -439,8 +439,16 @@ pub fn extract_features_from_frame(
             .clamp(0.0, 1.0)
     };
 
-    let variance_motion = (temporal_variance / 10.0).clamp(0.0, 1.0);
-    let mbp_motion = (motion_band_power / 25.0).clamp(0.0, 1.0);
+    // Normalize the energy-based motion terms by signal power (mean_amp^2) so
+    // they are independent of absolute amplitude. Without this, adding external
+    // antennas raised amplitudes ~5x and these raw energies ~30x, pinning both
+    // terms at the 1.0 clamp — which saturates raw_motion and defeats the
+    // adaptive baseline subtraction in smooth_and_classify (an empty room then
+    // reads "present"). Same sqrt-of-power-ratio form as temporal_motion_score
+    // above. Field-model (--calibrate) path is unaffected.
+    let amp_ref = mean_amp * mean_amp + 1e-9;
+    let variance_motion = (temporal_variance / amp_ref).sqrt().clamp(0.0, 1.0);
+    let mbp_motion = (motion_band_power / amp_ref).sqrt().clamp(0.0, 1.0);
     let cp_motion = (change_points as f64 / 15.0).clamp(0.0, 1.0);
     let motion_score = (temporal_motion_score * 0.4
         + variance_motion * 0.2
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -1378,15 +1378,25 @@ fn parse_esp32_frame(buf: &[u8]) -> Option<Esp32Frame> {
     let n_antennas = buf[5];
     let n_subcarriers = buf[6];
     let freq_mhz = u16::from_le_bytes([buf[8], buf[9]]);
-    let sequence = u32::from_le_bytes([buf[10], buf[11], buf[12], buf[13]]);
-    let rssi_raw = buf[14] as i8;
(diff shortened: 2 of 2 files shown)


ruvnet:
Reviewed — **both fixes are correct.** 👍

**CSI header offsets**: verified against the firmware serializer on `main` (`firmware/esp32-csi-node/main/csi_collector.c`):
- `memcpy(&buf[12], &seq, 4)` → seq at `[12..16)` (line 170)
- `buf[16] = rssi` (line 173)
- `buf[17] = noise_floor` (line 176)

The old parser read `seq` at `[10..14)`, `rssi` at `buf[14]`, `noise` at `buf[15]` — 2 bytes early, contradicting even its own layout comment. Because `freq_mhz` occupies the full `[8..12)` (4-byte `memcpy`), the old `rssi` read landed on the sequence counter's 3rd byte, which stays `0` for the first 65 536 frames → `rssi = 0`. Your `[12..16)/[16]/[17]` exactly matches the firmware. Correct.

**motion_score de-saturation**: normalizing `variance_motion`/`mbp_motion` by `mean_amp²` (the same dimensionless power-ratio form `temporal_motion_score` already uses) is the right call — it makes the term amplitude-scale-invariant so external-antenna gain no longer pins it at the `1.0` clamp. Sound, and it correctly lands in the shared `extract_features_from_frame` used by both the aggregate and per-node paths.

**Heads-up — needs a rebase (my fault).** GitHub now shows this `CONFLICTING/DIRTY`. The conflict is **not** in the fix itself; it's the new tests you append at EOF (`@@ -6787 … mod rolling_p95_tests`). Three PRs I merged after you opened this (#918 per-node MQTT, #919 model-load diagnostic, #920 export-rvf guard) each added a test module at that same spot, plus shifted `main.rs` line numbers. A `git rebase origin/main` should be clean — just move your new test module(s) after the now-present `mqtt_bridge_tests` / `model_load_diagnostic_tests` / `export_rvf_mode_tests`, and re-anchor the `parse_esp32_frame` / `extract_features_from_frame` hunks (their function bodies are unchanged by my PRs, so the offset/motion edits apply as-is).

LGTM to merge once rebased.

PR #931, opened 2026-06-03
fix(protocol): resolve 0xC511_0004 magic collision (closes #928)

### LINKED ISSUE #928: Wire-protocol magic collision: 0xC5110004 used for BOTH fused-vitals and WASM output — C6+mmWave vitals misparsed
## Summary
The UDP packet magic **`0xC5110004`** is assigned to **two different packet formats** in firmware, and **both are transmitted**. The sensing-server has only one parser for that magic (WASM output) and **no fused-vitals parser**, so on the ESP32-C6 + mmWave configuration the fused-vitals packet is silently misparsed as WASM output.

## Evidence

**Two firmware definitions of the same magic:**
- `firmware/esp32-csi-node/main/edge_processing.h:127` — `#define EDGE_FUSED_MAGIC 0xC5110004` (ADR-063 fused vitals, 48-byte `edge_fused_vitals_pkt_t`)
- `firmware/esp32-csi-node/main/wasm_runtime.h:46` — `#define WASM_OUTPUT_MAGIC 0xC5110004` (ADR-040 WASM output, variable-length)

**Both are sent:**
- `edge_processing.c:594` — `fpkt.magic = EDGE_FUSED_MAGIC;` (gated on `mmwave_sensor_get_state(&mw) && mw.detected`, edge_processing.c:589 — i.e., the ESP32-C6 + Seeed MR60BHA2 mmWave config)
- `wasm_runtime.c:386` — `pkt.magic = WASM_OUTPUT_MAGIC;`

**Canonical registry says it's fused-vitals:**
- `firmware/esp32-csi-node/main/rv_feature_state.h:13` — `0xC5110004 — ADR-063 fused vitals (edge_processing.h)`

**Server has only the WASM parser for 0xC5110004, no fused parser:**
- `v2/crates/wifi-densepose-sensing-server/src/main.rs:1140` (and `csi.rs:54`) — `parse_wasm_output` matches `magic != 0xC511_0004`.
- Dispatch: `main.rs:4982-4983` calls `parse_wasm_output(&buf[..len])` for `0xC5110004`. There is **no** parser for the 48-byte `edge_fused_vitals_pkt_t`. (The `fused`/`fuse_*` references around main.rs:4777/4854 are server-side *multistatic feature fusion*, unrelated to this packet.)

## Impact (C6 + mmWave)
When mmWave detects a person, the C6 sends a 48-byte fused-vitals packet (`0xC5110004`). The server runs `parse_wasm_output` on it:
- `buf[4]` node_id → read as WASM `node_id` (coincidentally ok)
- `buf[5]` **flags** → read as WASM `module_id`
- `buf[6..8)` **breathing_rate** (BPM×100, e.g. `1600` for 16 BPM) → read as WASM `event_count`
- then it reads "events" from the vitals payload → **garbage WASM events**, and the **mmWave-fused vitals are never parsed** (lost).

So on a supported hardware config (ESP32-C6 + MR60BHA2, per the README/CLAUDE hardware table), fused vitals silently fail and spurious WASM events may be emitted.

## Suggested direction (defer protocol decision to maintainers)
This is a wire-protocol change affecting deployed firmware, so it needs a coordinated call:
1. Reassign one magic to a free value. The registry treats `0xC5110004` as fused-vitals, so reassigning **`WASM_OUTPUT_MAGIC`** (e.g. to `0xC5110006`) in `wasm_runtime.h` + the server's `parse_wasm_output` is the smaller blast radius.
2. **Add a server-side parser** for `edge_fused_vitals_pkt_t` (0xC5110004) so fused vitals are actually consumed.
3. Validate on real C6+mmWave hardware (cannot be verified in CI/without the device).

Found while cross-checking UDP parser field offsets against the firmware serializers (the same audit that surfaced the rssi-offset bug fixed in #923). `parse_esp32_vitals` (0xC5110002) and `parse_esp32_frame` (0xC5110001, post-#923) offsets are correct; this collision is the remaining protocol-level issue.

ruvnet:
## Summary

Closes #928. `0xC511_0004` was double-assigned in firmware (ADR-063 fused-vitals AND ADR-040 WASM output) and both packet types were transmitted. The sensing-server only knew the WASM shape, so on the C6+mmWave config the 48-byte fused-vitals packet was being silently misparsed as malformed WASM events — fused vitals were lost, spurious WASM events leaked to subscribers.

## Approach

Per the issue's recommended direction (smaller blast radius — the registry already treats `0xC511_0004` as fused-vitals canonical):

1. **Reassign `WASM_OUTPUT_MAGIC`** from `0xC511_0004` → `0xC511_0007` (next free slot per `rv_feature_state.h` registry). Firmware-side change in `wasm_runtime.h`.
2. **Add `parse_edge_fused_vitals` + `EdgeFusedVitalsPacket`** on the server. Byte layout mirrors `edge_fused_vitals_pkt_t` from `edge_processing.h:129` exactly, gated by the firmware's `_Static_assert(sizeof(...) == 48)`.
3. **Dispatch arm** added to the UDP receive loop, tried *before* the WASM parser so stale firmware (old WASM on `0004`) fails to parse as both and gets dropped — deliberate fail-loud rather than the pre-fix silent garbage.
4. **Registry comment update** in `rv_feature_state.h` documenting the new `0x...0007` row.
5. **5 new tests** in `issue_928_magic_collision_tests` mod guarding both halves of the swap.

## Test plan

- [x] `cargo test -p wifi-densepose-sensing-server --no-default-features --bin sensing-server` → **122 passed / 0 failed** (5 new + 117 existing, unchanged).
- [x] `parse_edge_fused_vitals` round-trips a synthetic 48-byte packet — every field including `breathing/100`, `heartrate/10000`, mmWave block.
- [x] `parse_wasm_output` rejects the legacy `0xC511_0004`.
- [x] `parse_wasm_output` accepts the new `0xC511_0007`.
- [x] Hardware validation deferred — issue notes it can't run in CI without a real C6+MR60BHA2 board.

## WebSocket schema

New broadcast type: `{\"type\": \"edge_fused_vitals\", \"node_id\": ..., \"breathing_rate_bpm\": ..., \"heartrate_bpm\": ..., \"mmwave\": {\"hr_bpm\": ..., \"br_bpm\": ..., \"distance_cm\": ..., \"targets\": ..., \"confidence\": ..., \"type\": ...}, ...}`.

Additive — existing subscribers gain a new branch; none break.

## Deployment

⚠ **Wire-protocol change.** Firmware older than this commit emits WASM output on `0xC511_0004`; updated hosts will drop those frames. Operators running C6+mmWave should reflash firmware concurrent with the host upgrade. The pre-fix silent misparsing is the worse failure mode; this is the right tradeoff.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -46,7 +46,10 @@ jobs:
 
     - name: Run Bandit security scan
       run: |
-        bandit -r src/ -f sarif -o bandit-results.sarif
+        # The Python codebase lives under archive/v1/src (it moved there when
+        # the runtime was rewritten in Rust). Scanning `src/` matched nothing,
+        # so this SAST step was a silent no-op.
+        bandit -r archive/v1/src/ -f sarif -o bandit-results.sarif
       continue-on-error: true
 
     - name: Upload Bandit results to GitHub Security
@@ -57,22 +60,20 @@ jobs:
         sarif_file: bandit-results.sarif
         category: bandit
 
-    - name: Run Semgrep security scan
-      continue-on-error: true
-      uses: returntocorp/semgrep-action@v1
-      with:
-        config: >-
-          p/security-audit
-          p/secrets
-          p/python
-          p/docker
-          p/kubernetes
-      env:
-        SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
-        
-    - name: Generate Semgrep SARIF
+    # Removed the deprecated `returntocorp/semgrep-action@v1` step: it was
+    # redundant (the pip `semgrep --sarif` below is what feeds GitHub Security;
+    # the action only pushed to the Semgrep cloud app via SEMGREP_APP_TOKEN) and
+    # it pulled `returntocorp/semgrep-agent:v1` from Docker Hub on every run,
+    # which intermittently timed out and turned this check red. The pip semgrep
+    # (installed above) needs no Docker pull. The action's `p/docker` +
+    # `p/kubernetes` rulesets are folded into the command below so coverage is
+    # preserved.
+    - name: Run Semgrep + generate SARIF
       run: |
-        semgrep --config=p/security-audit --config=p/secrets --config=p/python --sarif --output=semgrep.sarif src/
+        semgrep \
+          --config=p/security-audit --config=p/secrets --config=p/python \
+          --config=p/docker --config=p/kubernetes \
+          --sarif --output=semgrep.sarif archive/v1/src/
       continue-on-error: true
 
     - name: Upload Semgrep results to GitHub Security
--- a/firmware/esp32-csi-node/main/rv_feature_state.h
+++ b/firmware/esp32-csi-node/main/rv_feature_state.h
@@ -12,7 +12,8 @@
  *   0xC5110003 — ADR-069 feature vector (edge_processing.h)
  *   0xC5110004 — ADR-063 fused vitals   (edge_processing.h)
  *   0xC5110005 — ADR-039 compressed CSI (edge_processing.h)
- *   0xC5110006 — ADR-081 feature state  (this file) ← new
+ *   0xC5110006 — ADR-081 feature state  (this file)
+ *   0xC5110007 — ADR-040 WASM output    (wasm_runtime.h, reassigned per issue #928)
  */
 
 #ifndef RV_FEATURE_STATE_H
--- a/firmware/esp32-csi-node/main/wasm_runtime.h
+++ b/firmware/esp32-csi-node/main/wasm_runtime.h
@@ -43,7 +43,16 @@
 
 #define WASM_MAX_MODULE_SIZE (128 * 1024)  /**< Max .wasm binary size (128 KB). */
 #define WASM_STACK_SIZE      (8 * 1024)    /**< WASM execution stack (8 KB). */
-#define WASM_OUTPUT_MAGIC    0xC5110004    /**< WASM output packet magic. */
+/* Issue #928: WASM output was originally 0xC5110004, but that magic is
+ * canonically owned by ADR-063 fused vitals (edge_processing.h). Both packets
+ * were transmitted on the same magic, and the host parser only knew the WASM
+ * shape, so on the ESP32-C6 + MR60BHA2 mmWave config the 48-byte fused-vitals
+ * packet was being read as garbage WASM events. Reassigned to 0xC5110007 (next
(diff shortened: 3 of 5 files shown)


ruvnet:
Verified this end-to-end (I filed #928) — **the protocol fix is correct and complete.** 👍

**Offsets match the firmware struct byte-for-byte.** Checked `parse_edge_fused_vitals` against `edge_fused_vitals_pkt_t` (`edge_processing.h`, `__attribute__((packed))`, 48 bytes):

| field | firmware offset | parser |
|---|---|---|
| node_id | [4] | `buf[4]` ✓ |
| flags | [5] | `buf[5]` ✓ |
| breathing (×100) | [6..8) | `buf[6..8]` ✓ |
| heartrate (×10000) | [8..12) | `buf[8..12]` ✓ |
| rssi | [12] | `buf[12] as i8` ✓ |
| n_persons | [13] | `buf[13]` ✓ |
| mmwave_type | [14] | `buf[14]` ✓ |
| fusion_confidence | [15] | `buf[15]` ✓ |
| motion_energy | [16..20) | ✓ |
| presence_score | [20..24) | ✓ |
| timestamp_ms | [24..28) | ✓ |
| mmwave_hr/br/distance | [28..40) | ✓ |
| mmwave_targets/confidence | [40]/[41] | ✓ |

Note that [14]/[15] correctly use what's `reserved[2]` in the 32-byte `edge_vitals_pkt_t` — the fused variant repurposes them for `mmwave_type`/`fusion_confidence`, and the parser gets that right.

- **Dispatch wiring** ✓ — `if let Some(fused) = parse_edge_fused_vitals(&buf[..len])` is present, so fused packets are actually consumed (not just parseable).
- **Tests** ✓ — well-formed parse + 47-byte short-packet rejection.
- **Magic reassignment** ✓ — `WASM_OUTPUT_MAGIC` → `0xC5110007` in both firmware (`wasm_runtime.h`) and both server copies (`csi.rs` + `main.rs`), and the registry updated. Right call per the issue (registry treats `0xC5110004` as canonical fused-vitals).

**Two things before merge:**
1. **`security-scan.yml` is now redundant** — this PR includes the same SAST path/semgrep fix that just merged as #930. Rebase on `main` and drop that file from the diff (it'll conflict otherwise).
2. **Backward-compat (worth a line in the PR/changelog):** a still-deployed node running pre-#928 firmware sends WASM as `0xC5110004`; the updated server now reads that as fused-vitals. That's unavoidable with any reassignment and acceptable (WASM on `0xC5110004` was already broken by the collision), but flashing nodes + host together is the clean upgrade path.

Nice work — closes #928 correctly. LGTM after the rebase.

PR #962, opened 2026-06-06
fix(firmware): clear onboard WS2812 LED on GPIO 48 (DevKitC-1 v1.1 / N16R8)

dvcdsys:
## Problem
On ESP32-S3 DevKitC-1 v1.1 / N16R8 modules / generic YD-ESP32-S3 clones, the onboard RGB (WS2812) LED stays lit at a bright color after flashing — the firmware never turns it off.

## Root Cause
`app_main()` clears the onboard WS2812 on a single GPIO: **38** on S3, 8 on C6. But many S3 boards (DevKitC-1 **v1.1**, N16R8 modules, YD-ESP32-S3 clones) wire the WS2812 to **GPIO 48**, not 38. The clear targets the wrong pin, so the LED is left in its power-on state.

## Fix
Clear every candidate LED GPIO on S3 — both **38 and 48** (C6 unchanged at 8). Each `led_strip` handle is created → `led_strip_clear()` → `led_strip_del()` so the RMT channel is freed. Driving a pin with no LED attached is a harmless reset pulse, and clearing 48 is a no-op on GPIO-38 boards.

## Verification
- **Device:** ESP32-S3 N16R8 (16 MB flash, 8 MB PSRAM, QFN56 rev v0.2)
- **AP:** 2.4 GHz WPA2 (bgn)
- **Result:** bright RGB LED → **off** after flashing this build (visually confirmed). Same firmware flashed to 4 such boards. No regression expected on GPIO-38 boards (clearing 48 is a no-op there).

## Risk
Minimal. One extra RMT create/clear/del on GPIO 48 at boot (~microseconds), then the handle is freed. No runtime cost after boot; no behavior change on boards whose LED is already on GPIO 38.

### DIFF
--- a/firmware/esp32-csi-node/main/main.c
+++ b/firmware/esp32-csi-node/main/main.c
@@ -174,27 +174,34 @@ void app_main(void)
              target_name, app_desc->version, g_nvs_config.node_id);
 
     /* Turn off onboard WS2812 LED.
-     * S3 dev boards put the LED on GPIO 38; C6 dev boards on GPIO 8.
-     * On C6, GPIO 38 doesn't exist (only 0-30) — gate the init by target. */
+     * S3 dev boards put the LED on GPIO 38 (DevKitC-1 v1.0) or GPIO 48
+     * (DevKitC-1 v1.1 / N16R8 / generic YD-ESP32-S3 clones); C6 dev boards
+     * on GPIO 8. Clear every candidate pin — pulsing a pin with no LED
+     * attached is harmless, and each handle is deleted after clear to free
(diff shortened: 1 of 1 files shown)


ruvnet:
✅ **Reviewed — good fix, correct board.** DevKitC-1 v1.1 / N16R8 do wire the WS2812 to GPIO 48 (I'm running this exact board). Clearing both 38 and 48 is the right approach, and the added `led_strip_del()` after each clear is a nice catch — the original leaked the RMT channel by never deleting the handle. Pulsing an unused pin is harmless. 

I can flash this to my N16R8 to visually confirm the LED goes dark if you'd like hardware sign-off before merge.

PR #977, opened 2026-06-08
feat(csi): ALL-frames promiscuous filter for higher CSI yield

Wl8259:
Switch from MGMT-only to ALL-frames promiscuous filter.

- Previous: 0-1 pps on quiet 2.4GHz channels
- New: 10-50+ pps with active WiFi traffic
- Fixes heart rate / breathing detection accuracy


ruvnet:
Thanks for this — the goal (higher CSI yield) is right, but closing for two concrete reasons:

**1. It doesn't compile.** I built the branch for esp32s3 and hit:
```
csi_collector.c:469:5: error: "/*" within comment [-Werror=comment]
```
The edit kept the original `/* MGMT-only promiscuous filter (RuView#396).` opening line and started a **new** `/*` block inside it, so the comment never closes before `wifi_promiscuous_filter_t filt`. No CI ran on the branch (`no checks reported`), so this was never gated.

**2. It's superseded — and the safer version is already merged + hardware-validated.** Higher CSI yield landed via #906/#1000 as a **conditional** MGMT→MGMT+DATA upgrade (gated on `display_is_active()` so display builds avoid DATA-load contention). I hardware-validated that on a bare ESP32-S3 QFN56 rev v0.2 N16R8: it ran 600 CSI callbacks over 22s with **zero crashes**, shipped in release `v0.8.1-esp32`. This PR's unconditional `WIFI_PROMIS_FILTER_MASK_ALL` is blunter: it ignores that display gating and adds CTRL frames (ACKs etc.) that add interrupt load for little CSI value — the exact failure mode the original `wDev_ProcessFiq` comment you removed was warning about.

If you'd like to push yield further on top of the current conditional design (e.g. MGMT+DATA when headless *and* a traffic-aware CTRL toggle), that'd be a welcome follow-up — happy to review a rebased version against current `main`.

Issue #1010, opened 2026-06-11
wifi-densepose-train tch-backend: does not compile against workspace-pinned tch 0.24; clamp(1.0, f64::MAX) runtime panic on PyTorch ≥2.x; no real-validation path (checkpoint selection blind on real data)


stuinfla:
Found while running the first real-data training on live-captured CSI (~8.6k labeled MM-Fi-format frames). All items below were verified by fixing them and then running training end-to-end to a 40% cross-subject PCK@0.2 (held-out subject) — happy to PR any/all.

**1. tch-backend feature is bit-rotted: 13 compile errors against the workspace's own `tch = "0.24"` pin.** The optional feature isn't CI-gated, so it drifted:
- `Vec::<f64>::from(tensor)` → tch 0.24 moved conversions to `TryFrom` (8 sites: model.rs:300/359, proof.rs:156/160/264, trainer.rs:585/590/643/657)
- `(&tensor % i64)` → `Rem` impl dropped; use `.fmod()` (trainer.rs:627)
- `EdgeReference::id()/.target()` need `use petgraph::visit::EdgeRef;` (metrics.rs, errors at 1103-1104)
- `t.numel()` usize/i64 sum mismatch (model.rs:185)

**2. Real runtime bug: `losses.rs:121/168/237` `.clamp(1.0, f64::MAX)`** — PyTorch ≥2.x rejects converting f64::MAX into an f32 tensor ("cannot be converted to type float without overflow"). First call panics. Fix: `.clamp_min(1.0)` (exact intended semantic).

**3. macOS + LIBTORCH_USE_PYTORCH: the train binary fails at dyld** (libtorch_cpu.dylib not found) because torch-sys only rpaths its own static lib. Standard fix: a small build.rs emitting `cargo:rustc-link-arg-bins=-Wl,-rpath,<torch>/lib` when LIBTORCH_USE_PYTORCH is set.

**4. Design gap with real consequences: `bin/train.rs` has no real-validation path.** Training on a real `--data-dir` always pairs it with a synthetic validation set ("for pipeline verification"). Consequences on real data: val_pck is meaningless noise (~0.02-0.06), so best-checkpoint selection and early stopping are driven by noise — silently. We added a `--val-dir` flag (held-out MM-Fi dir → `MmFiDataset::discover`) and immediately got honest cross-subject metrics + meaningful checkpoint selection; suggest upstreaming that or an automatic subject-level split.

[...]

Issue #1009, opened 2026-06-11
ADR-151 calibrate cannot ingest real ESP32-C6 HE20 CSI: n_subcarriers parsed as u8 (overflows at 256) + he20 tier expects 242 but wire delivers 256 bins


stuinfla:
Found while running `wifi-densepose calibrate --tier he20` against a live 3-node ESP32-C6 HE mesh (the first real HE20 CSI per #1005). Every frame was skipped; baseline failed with 0 frames. Two independent bugs:

**1. u8 overflow in the CLI packet parser** — `wifi-densepose-cli/src/calibrate.rs` `parse_csi_packet`: `let n_subcarriers = buf[6] as usize;` but the ADR-018 wire carries n_subcarriers as **u16 LE at bytes 6-7**. Real HE20 frames carry 256 → byte 6 = 0x00 → parses as 0 subcarriers. Invisible until now because S3's 192 fits in a byte. Fix (verified): `u16::from_le_bytes([buf[6], buf[7]]) as usize`. Note `wifi-densepose-sensing-server` parse_esp32_frame is documented as the template for this parser — worth checking for the same truncation.

**2. he20 tier config mismatch** — `wifi-densepose-signal/src/ruvsense/calibration.rs` `CalibrationConfig::he20()` sets `num_active: 242`, but the ESP-IDF v5.5.2 driver delivers **all 256 FFT bins** on the wire (242 data tones + pilots/guards, n_subc=0x0100 wire-verified). The recorder's mismatch check rejects every real frame. Fix we ran with: num_active 256 for he20 (and the cir.rs he20 config + 3 test asserts). Alternatively trim to the 242 active tones at ingest — but that requires the exact HE20 tone map to avoid corrupting the baseline.

With both fixes applied: `calibrate --tier he20 --duration-s 45` captured a clean 600-frame empty-room baseline from the live mesh (summary: frames=600 tier=He20 subcarriers=256).

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

stuinfla:
Follow-up after d0e27e65 (#1011) landed: both halves of this issue are now addressed upstream, so we are not PRing our local patches.

- Part 1 (u8 parse overflow): fixed in #1011's host-ingest commit — calibrate.rs now reads n_subcarriers as u16 LE at bytes 6-7 with the full ADR-018/ADR-110 layout (and surfaces PPDU type), which supersedes our local fix.
[...]

Issue #830, opened 2026-05-27
Xiao esp32S3


fr3dsant:
I try to use the xiao esp32S3 what firmware you recomend?

ruvnet:
The Seeed XIAO ESP32-S3 uses the same Xtensa dual-core chip as the standard ESP32-S3, so it is compatible with the RuView firmware.

**Which firmware to use:**
- XIAO ESP32-S3 (standard, 4 MB flash) → use the `esp32-csi-node-4mb.bin` firmware
- XIAO ESP32-S3 Sense (8 MB flash) → use `esp32-csi-node.bin` (8 MB firmware)

Both binaries are in the [latest release](https://github.com/ruvnet/RuView/releases). The `bootloader.bin` and `partition-table.bin` (or `partition-table-4mb.bin`) must be flashed alongside the main binary.

For step-by-step setup see the tutorial in **#34** (ESP32-S3 CSI Pipeline End-to-End Setup), which covers flashing, provisioning WiFi via `provision.py`, and connecting to the sensing server.

---
_Generated by [Claude Code](https://claude.ai/code)_

ruvnet:
The Seeed XIAO ESP32-S3 uses the ESP32-S3 chip and should work with the standard firmware. Flash the 4MB variant (`esp32-csi-node-4mb.bin`) since XIAO has 8MB flash but the 4MB partition layout fits its typical configuration. See the flashing guide in issue #34 for step-by-step instructions, and use `provision.py` to set your WiFi credentials and target IP after flashing. The compact form factor of the XIAO makes it a good choice for deployment.

---
_Generated by [Claude Code](https://claude.ai/code)_

fr3dsant:
I already tested the 4MB firmware but without success. With the 8MB version, it gave some signal, but very little, which was strange, and I could only get all the settings active, but without significant reference values, as soon as I connected a random device with Wi-Fi. Since many people are complaining about problems, wouldn't it be possible for you to create a visual tutorial of all the processes?

ruvnet:
The Seeed XIAO ESP32S3 uses the same ESP32-S3 chip, so the CSI firmware is compatible.

[...]

PR #1013, opened 2026-06-11
RFC feat(firmware): XIAO ESP32S3 Sense camera node — CSI + OV3660 /snap + /stream on :8081 (compile-time optional)

stuinfla:
## RFC: optional onboard camera for the S3 CSI node (XIAO ESP32S3 Sense)

Feature proposal — marked RFC because it adds a new hardware capability rather than fixing a defect. Happy to adjust scope/naming to taste.

## Motivation

The Seeed XIAO ESP32S3 Sense is an S3 CSI node with a free OV3660 camera on board. Having the same node serve ground-truth snapshots/video alongside its CSI stream is directly useful for labeling and sanity-checking pose/presence work (e.g. ADR-150-style training captures) without adding a second device to the rig.

## What this adds

- `main/camera_node.c` / `camera_node.h` (~190 lines): OV3660 init with the hardware-verified XIAO ESP32S3 Sense pin map (sensor PID 0x3660), plus an HTTP server with
  - `GET /snap` — single fresh JPEG (one stale framebuffer discarded first)
  - `GET /stream` — multipart MJPEG at ~5 fps
- Kconfig menu: `CONFIG_CAMERA_ENABLE` (default **n**, `depends on IDF_TARGET_ESP32S3`) and `CONFIG_CAMERA_HTTP_PORT` (default 8081 — the OTA/WASM server owns 8032; httpd `ctrl_port` offset to 32769 since the OTA instance owns 32768)
- `idf_component.yml`: `espressif/esp32-camera ^2.0.0` with `rules: if target == esp32s3`, so non-S3 builds never fetch it
- `sdkconfig.defaults.s3cam` build overlay: octal PSRAM (required for framebuffers), display/WASM trimmed (those pins collide with the camera on the XIAO)
- `main.c`: starts the camera after WiFi is up; **init failure is non-fatal by design** — logs a warning and CSI streaming continues unaffected

## CSI-coexistence design notes

- QVGA JPEG, `fb_count=2`, `CAMERA_GRAB_LATEST`, quality 12 in PSRAM is the proven-reliable config from a probe firmware (SVGA with `fb_count=1` overflowed: FB-OVF); bounded per-frame airtime keeps the CSI promiscuous callback from being starved
- httpd tasks pinned to core 1, away from WiFi/CSI work on core 0
- Zero impact when disabled: `camera_node.c` is only compiled under `CONFIG_CAMERA_ENABLE`, and the default for every existing target/overlay is off — the C6 and stock S3 builds are byte-identical to main

## Hardware/test evidence

- Built clean on ESP-IDF v5.5.2, target esp32s3, using the documented two-step `idf.py -B build-s3cam` overlay invocation (no host-side Rust changes in this PR)
- Flashed and exercised on a XIAO ESP32S3 Sense: camera comes up with PID 0x3660, `/snap` returns fresh JPEGs and `/stream` sustains ~5 fps while the node keeps streaming CSI
- Companion context: this node fed the live HE20 capture/training work referenced in #1009 / #1010

### DIFF
--- a/firmware/esp32-csi-node/main/CMakeLists.txt
+++ b/firmware/esp32-csi-node/main/CMakeLists.txt
@@ -62,6 +62,12 @@ if(CONFIG_WASM_ENABLE)
     list(APPEND REQUIRES wasm3)
 endif()
 
+# XIAO ESP32S3 Sense camera + HTTP snapshot/MJPEG server (compile-time optional)
+if(CONFIG_CAMERA_ENABLE)
+    list(APPEND SRCS "camera_node.c")
+    list(APPEND REQUIRES espressif__esp32-camera)
+endif()
+
 idf_component_register(
     SRCS ${SRCS}
     INCLUDE_DIRS "."
--- a/firmware/esp32-csi-node/main/Kconfig.projbuild
+++ b/firmware/esp32-csi-node/main/Kconfig.projbuild
@@ -251,6 +251,31 @@ menu "AMOLED Display (ADR-045)"
 
 endmenu
 
+menu "Camera Node (XIAO ESP32S3 Sense)"
+
+    config CAMERA_ENABLE
+        bool "Enable onboard camera + HTTP snapshot/MJPEG server"
+        default n
+        depends on IDF_TARGET_ESP32S3
+        help
+            Initialize the onboard camera (Seeed XIAO ESP32S3 Sense pin map,
+            OV3660) after WiFi is up and serve GET /snap (single JPEG) and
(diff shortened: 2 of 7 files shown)


ruvnet:
🔵 **RFC — supportive, pending a maintainer call on scope.** On-device OV3660 ground-truth (`/snap` + `/stream`) on the same CSI node is directly useful for ADR-079 camera-supervised pose/presence labeling, and compile-time gating means zero impact on the default build — both good. The open question is policy: does RuView want camera firmware in-tree (vs a sibling repo)? Addresses #830. I'd lean yes given it's opt-in and well-scoped; happy to help review the `camera_node.c` HTTP server once there's a scope decision.

Issue #1005, opened 2026-06-11
ESP32-C6 delivers HE20 CSI (256 bins / 242 tones) when built with IDF v5.5.2 — resolves WITNESS-LOG-110 §B1 open question; includes IDF 5.5 build fix for c6_sync_espnow.c


stuinfla:
## Summary

Two findings from a production deployment of `firmware/esp32-csi-node` (source @ 2a307138, v0.7.1-esp32):

1. **HE-LTF CSI is real on the C6** — WITNESS-LOG-110 §B1 flagged "whether ESP-IDF v5.4's CSI callback exposes HE-LTF subcarriers via wifi_csi_info_t.buf is an open question — the driver may quietly downconvert." We measured it: **IDF v5.4 downconverts; IDF v5.5.2 delivers true HE-LTF.** Same board, same AP, same link — the only variable was the IDF/driver version.
2. **c6_sync_espnow.c fails to build on IDF 5.5.x** — one-line guard fix below.

## 🏆 Why this is a big win for WiFi sensing

This isn't just a wire-format detail — it changes what the C6 is *for* in a sensing fleet:

1. **4× denser spectral sampling.** The S3's famous 192-subcarrier mode is several coarse LTF views stitched together by the IDF's 4-LTF-merge (still 312.5 kHz tone spacing). The C6's HE20 capture is **one coherent snapshot of 242 active tones at 78.125 kHz spacing** — the same 20 MHz channel sampled 4× finer. For micro-motion sensing (breathing, heartbeat, postural sway), phase resolution per Hz is what sets the SNR, and HE-LTF is the only ESP32 path to it.

2. **A modern protocol link.** HE/Wi-Fi 6 brings deterministic-timing potential (iTWT, already in ADR-110) and cleaner demodulation — and it's the prerequisite for dense HE-LTF captures at all. The C6 stops being "the cheap 64-subcarrier node" and becomes the highest-resolution CSI source in the ESP32 family.

3. **A matched multistatic mesh.** We ran **three C6 nodes simultaneously** in this mode against one AX210 AP on a dedicated channel, with the ADR-110 ESP-NOW time-sync (±104 µs) broadcasting alongside. Identical 256-bin views from three positions with sub-ms time alignment is the geometric foundation multistatic localization needs — something neither the S3s (stitched LTF) nor the C5s (117 @ 5 GHz) can provide at this density.

[...]

PR #1014, opened 2026-06-11
fix(train): un-rot tch-backend vs workspace tch 0.24 + clamp/floor-div runtime bugs + real --val-dir validation and --eval-only mode

### LINKED ISSUE #1010: wifi-densepose-train tch-backend: does not compile against workspace-pinned tch 0.24; clamp(1.0, f64::MAX) runtime panic on PyTorch ≥2.x; no real-validation path (checkpoint selection blind on real data)

stuinfla:
## Problem

`wifi-densepose-train`'s optional `tch-backend` feature does not compile against the workspace's own `tch = "0.24"` pin (13 errors), and even after compiling, the first real training step panics. Because `mod metrics`, `mod model`, `mod trainer`, etc. are all `#[cfg(feature = "tch-backend")]` and CI builds `--no-default-features`, none of this is visible to the PR gate — the feature bit-rotted silently. Full write-up with error sites: #1010.

On top of that, `bin/train.rs` always pairs a real `--data-dir` with a **synthetic** validation set, so on real data `val_pck` is noise (~0.02–0.06) and best-checkpoint selection / early stopping are driven by that noise — silently.

## Root cause → fix (file:line against current main)

**1. tch 0.24 API drift (compile errors):**
- `Vec::<f64>::from(tensor)` → conversions moved to `TryFrom` in tch 0.24 — `model.rs:300,359`, `proof.rs:156,160,264`, `trainer.rs:585,590,643,657`
- `(&tensor % i64)` → scalar `Rem` impl dropped; use `.fmod()` — `trainer.rs:627`
- `t.numel()` returns `usize`, summed into `i64` — `model.rs:185` (`as i64`)
- `EdgeReference::id()/.target()/.weight()` now need `use petgraph::visit::EdgeRef;` — `metrics.rs` (errors at 1103–1105)

**2. Runtime panic:** `losses.rs:121,168,237` — `.clamp(1.0, f64::MAX)`: PyTorch ≥ 2.x rejects converting `f64::MAX` into an f32 tensor ("cannot be converted to type float without overflow"), so the first loss call panics. `.clamp_min(1.0)` is the exact intended semantic.

**3. dyld failure with `LIBTORCH_USE_PYTORCH=1`** (macOS/Linux): torch-sys only rpaths its own cc-built static lib, which does not propagate to downstream binaries — `train` dies at launch with `libtorch_cpu` not found, and so does the crate's own unit-test binary under `cargo test --features tch-backend`. New `build.rs` emits `cargo:rustc-link-arg=-Wl,-rpath,<torch>/lib` when that env var is set (plain `rustc-link-arg`, because the `-bins`/`-tests` variants do not reach the lib unit-test binary — verified with `otool -l`); a no-op otherwise.

**4. Latent decode bug exposed by the revived tests:** `trainer.rs:694` `heatmap_to_keypoints` — `(&arg / w)` on the integer argmax tensor is TRUE division in torch ≥ 1.6 (36/8 = 4.5), so every decoded y row was skewed by up to one row. `floor_divide_scalar(w)` restores integer row decomposition (the unit test pins center peak (4,4) → 4/7 exactly).

**5. Test-code fixes (pre-existing, tch-gated so invisible to CI):**
- `losses.rs:749` — E0507: `move` closure inside an `FnMut` closure moves the captured `heatmaps` array on every outer iteration; shadowed with `let heatmaps = &heatmaps;`
- `losses.rs` gaussian-peak thresholds were mathematically unreachable: kp 0.5 sits on a half-pixel, so the best attainable peak is exp(−0.5/(2σ²)) — 0.9394 for σ=2 (test demanded > 0.95) and 0.8007 for σ=1.5 (test demanded > 0.9). Relaxed to 0.93 / 0.79 with the math in comments.

**6. No real-validation path:** new `--val-dir <DIR>` loads a held-out MM-Fi directory via `MmFiDataset::discover` as the validation set (synthetic fallback unchanged when the flag is absent). New `--eval-only --checkpoint <FILE> [--dump-preds <JSONL>]` mode loads a checkpoint and runs the existing validation loop standalone (`Trainer::evaluate_with_dump` dumps per-sample pred+GT keypoints in dataset order for offline rendering/analysis).

## Test evidence

- `cargo check -p wifi-densepose-cli -p wifi-densepose-signal -p wifi-densepose-train --no-default-features` — clean (matches CI gate)
- `LIBTORCH_USE_PYTORCH=1 cargo check -p wifi-densepose-train --features tch-backend` — clean against PyTorch 2.11.0 (the 3 remaining warnings — `model.rs:31` unused `nn::ModuleT`, `proof.rs:28` unused `CsiDataset`, `trainer.rs:723` unused `num_kp` — are pre-existing on main and deliberately untouched)
- `cargo test -p wifi-densepose-train --no-default-features` — 7/7 pass
- `LIBTORCH_USE_PYTORCH=1 cargo test -p wifi-densepose-train --features tch-backend --lib -- --test-threads=1` — 194/195 pass. Honest accounting of what still fails and why (all pre-existing, none regressions from this PR):
  - `model::tests::save_and_load_roundtrip` fails against **PyTorch 2.11** libtorch: tch 0.24's `.pt` (jit `_load_parameters`) reload hits `Expected GenericDict but got Object`. Version interplay, not this code — real checkpoint reload verified via `.safetensors` (below). Likely passes against the libtorch generation tch 0.24 targets.
  - `proof::tests::hash_model_weights_is_deterministic` and `generate_and_verify_hash_matches` are **racy under parallel test threads** (other tests draw from torch's shared global RNG between the two `manual_seed` calls); both pass with `--test-threads=1`, so they are counted above as passing. A proper fix (RNG mutex) felt out of scope.
- End-to-end on hardware: with these fixes, training on ~8.6k live-captured 256-bin HE20 CSI windows (per #1005) with a held-out subject as `--val-dir` reached **40% cross-subject PCK@0.2** (vs 11.6% reported in ADR-150), with checkpoint selection driven by the real metric instead of synthetic noise.
- `--eval-only` round-trip verified: reloading `best_epoch0016_pck0.3999` (as `.safetensors` — see the PyTorch-2.11 `.pt` note above) and re-running the validation loop reproduces **exactly PCK@0.2 = 0.3999 / OKS = 0.1277 (n=3191)** on the original val set — the eval harness is bit-faithful to training-time validation.

Formatting note: `src/ablation.rs` and `src/bin/aa_score_runner.rs` currently fail `cargo fmt --check` on main; left untouched here to keep this PR surgical.

Closes #1010

### DIFF
--- a/v2/crates/wifi-densepose-train/build.rs
+++ b/v2/crates/wifi-densepose-train/build.rs
@@ -1,0 +1,31 @@
+//! Build script: when binding libtorch from the system PyTorch install
+//! (`LIBTORCH_USE_PYTORCH=1`), embed an rpath to the PyTorch `lib` directory so
+//! the produced binaries and test executables can locate `libtorch_cpu` at
+//! runtime. torch-sys only adds the rpath to its own cc-built static library,
+//! which does not propagate to downstream binaries on macOS/Linux.
+fn main() {
+    println!("cargo:rerun-if-env-changed=LIBTORCH_USE_PYTORCH");
+    if std::env::var_os("LIBTORCH_USE_PYTORCH").is_none() {
+        return;
+    }
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    if target_os != "macos" && target_os != "linux" {
+        return;
+    }
+    let output = std::process::Command::new("python3")
+        .args([
+            "-c",
+            "import os, torch; print(os.path.join(os.path.dirname(torch.__file__), 'lib'))",
+        ])
+        .output();
+    if let Ok(output) = output {
+        if output.status.success() {
+            let lib_dir = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            if !lib_dir.is_empty() {
+                // Plain rustc-link-arg (not -bins/-tests): it is the only
+                // variant that also reaches the lib unit-test binary.
+                println!("cargo:rustc-link-arg=-Wl,-rpath,{lib_dir}");
+            }
+        }
+    }
+}
--- a/v2/crates/wifi-densepose-train/src/bin/train.rs
+++ b/v2/crates/wifi-densepose-train/src/bin/train.rs
@@ -55,6 +55,12 @@ struct Args {
     #[arg(long, value_name = "DIR")]
     data_dir: Option<PathBuf>,
 
+    /// Optional MM-Fi directory used as the REAL validation set (e.g. a
+    /// held-out subject/session for honest cross-subject PCK). Without it,
+    /// a small synthetic set is used for pipeline verification only.
+    #[arg(long, value_name = "DIR")]
+    val_dir: Option<PathBuf>,
+
     /// Override the checkpoint output directory from the config.
     #[arg(long, value_name = "DIR")]
     checkpoint_dir: Option<PathBuf>,
@@ -76,6 +82,21 @@ struct Args {
     /// Log level: trace, debug, info, warn, error.
     #[arg(long, default_value = "info")]
     log_level: String,
+
+    /// Evaluation-only mode: load `--checkpoint` and run the validation loop
+    /// on `--val-dir` (no training). Prints PCK@0.2 / OKS and exits.
+    #[arg(long, default_value_t = false)]
+    eval_only: bool,
+
+    /// Path to a `.pt` checkpoint (tch VarStore format) to load in
+    /// `--eval-only` mode.
+    #[arg(long, value_name = "FILE")]
+    checkpoint: Option<PathBuf>,
+
+    /// In `--eval-only` mode, dump per-sample predicted + GT keypoints as
+    /// JSONL to this path (dataset order, one line per window).
+    #[arg(long, value_name = "FILE")]
+    dump_preds: Option<PathBuf>,
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


ruvnet:
✅ **Reviewed — valuable, and it exposes a CI blind spot.** The deeper lesson here is the one worth acting on: the `tch-backend` feature bit-rotted **silently because CI only builds `--no-default-features`**, so the gate never compiled it. Fixing the 13 errors + runtime panics is good; I'd pair it with a CI job that builds `--features tch-backend` (even build-only) so this can't re-rot. The real `--val-dir` validation + `--eval-only` additions are a nice bonus.

PR #1015, opened 2026-06-11
fix(calibration): ADR-151 presence flatline — per-frame p90|z| replaces floored median|z|; gates re-derived from bench; unsanitised phase excluded from motion

stuinfla:
## Summary

ADR-151 enrollment can **accept an occupied room as the `empty` anchor** — bench-reproduced on real ESP32-C6 HE20 captures and fixed here by replacing the per-frame presence statistic (median of |z| over subcarriers) with the **per-frame p90 of |z|**, re-deriving the gates from measured data, and excluding the unsanitised phase channel from motion.

## The defect (measured, not theorized)

`AnchorRecorder` computes `presence_z` as the mean over frames of `CalibrationDeviationScore::amplitude_z_median` — the per-frame **median** of |z| across *all* active subcarriers (`enrollment.rs`, scored in `wifi-densepose-signal` `ruvsense/calibration.rs::deviation`).

**The median-floor math:** a person perturbs a *minority* of subcarriers strongly. For the majority of bins that stay statistically like the baseline, |z| ~ |N(0,1)|, whose median is **0.674**. The median over all bins therefore floors near 0.674 and barely moves when someone enters the room. The default gates (`empty_max_z = 1.0`, `min_presence_z = 1.5`) sit in a band the statistic can rarely reach in either direction.

**Measured on real captures** (ESP32-C6 XIAO, HE20 2.4 GHz, 256-bin frames, node 8, replayed over UDP into the unmodified `d0e27e65` CLI):

| segment | per-frame median \|z\| (old statistic) | per-frame p90 \|z\| (this PR) |
|---|---|---|
| empty room (3 segments, same epoch) | 0.40–0.42 | **1.27–1.33** |
| person moving | 0.74–0.75 | **2.26–2.29** |

Both old values are below every gate — so on the clean upstream binary, `enroll` fed **100 % person-moving frames** ACCEPTED the `empty` anchor (`presence_z = 0.93`). The p90 separates cleanly because the upper tail is exactly where the body's multipath perturbation lives.

**Second finding — phase channel is noise on the CLI path:** `CalibrationRecorder`'s documented precondition is `PhaseSanitizer` + `phase_align.rs`, but the CLI UDP path feeds raw phase. The resulting baseline `phase_dispersion` ≈ 0.965 (uniformly random), so `phase_drift_median` ≈ π/2 on every frame and the |Δφ| motion term inflated still-room motion rates to ~60–70 % pure noise.

## The fix

- `CalibrationDeviationScore` gains **`amplitude_z_p90`** (numpy-convention linear interpolation; the median stays as a secondary/diagnostic field) and **`phase_usable`** (baseline median Von Mises dispersion ≤ `PHASE_DISPERSION_USABLE_MAX = 0.5`; the bench's unsanitised baseline measures 0.965, sanitized pipelines land ≤ 0.3 per `CalibrationConfig::max_phase_variance`).
- `motion_flagged` and the recorder's |Δφ| motion term only use phase when `phase_usable`.
- `AnchorRecorder` gates presence on the p90; gates re-derived from the bench bands with margin: **`EMPTY_MAX_Z = 1.6`**, **`MIN_PRESENCE_Z = 1.9`** (constants, with the bench numbers cited in the doc comments).
- `calibrate` banner now prints `z_p90` alongside `z_med`.

## Bench proof (replay harness, same-epoch baseline)

Harness: parse the Pi's raw `csi_raw` bins, replay node-8 256-bin frames over UDP at 50 Hz into the rebuilt CLI; calibrate a 600-frame baseline from an empty segment, then run `enroll` (first anchor = `empty`) against held-out segments from the **same capture window** (same-epoch requirement — cross-epoch baselines drift; mean-shape correlation across epochs measured only ~0.31).

| test | input | verdict |
|---|---|---|
| (a) calibrate | empty segment (10:30) | baseline 600 frames / 242 subcarriers, banner z_med ≈ 0.3–0.4, z_p90 ≈ 0.7–1.4 |
| (b) `empty` anchor fed person-moving frames | occupied bin | ✗ **REJECTED** — `room not empty (presence_z 2.54 > 1.60)` (upstream binary: ACCEPTED at 0.93) |
| (c) `empty` anchor fed held-out empty segment (10:31) | empty bin | ✓ **ACCEPTED** — `presence_z=1.21` |

Separation 1.21 vs 2.54 against a gate of 1.6, consistent with the offline study bands.

## Tests

- `wifi-densepose-calibration`: 43 unit tests green (suite updated for the p90 statistic; **new** regression tests: `presence_uses_p90_not_median_floor`, `unusable_phase_does_not_create_motion`) + `full_loop` integration green (sim now models the measured empty z-scale: real baselines integrate drift into the Welford variance, so runtime z ~ N(0, 0.8²) — matching the measured empty p90 of 1.27–1.33 vs 1.645 for ideal i.i.d. noise).
- `wifi-densepose-cli`: 19 tests green. `wifi-densepose-signal`: all suites green.
- No new clippy warnings (per-file warning counts identical to `d0e27e65`).

### DIFF
--- a/v2/crates/wifi-densepose-calibration/src/anchor.rs
+++ b/v2/crates/wifi-densepose-calibration/src/anchor.rs
@@ -127,7 +127,9 @@ impl AnchorLabel {
 /// Quality assessment of a captured anchor (from the enrollment quality gate).
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 pub struct AnchorQuality {
-    /// Median amplitude z-score vs the empty-room baseline (presence strength).
+    /// Mean per-frame p90 amplitude z-score vs the empty-room baseline
+    /// (presence strength — ADR-151; the median floors at ~0.674 with a
+    /// person present and is kept only as a recorder diagnostic).
     pub presence_z: f32,
     /// Fraction of frames flagged as motion.
     pub motion_rate: f32,
--- a/v2/crates/wifi-densepose-calibration/src/enrollment.rs
+++ b/v2/crates/wifi-densepose-calibration/src/enrollment.rs
@@ -8,8 +8,16 @@
 //!
 //! Quality is measured against the ADR-135 empty-room baseline via
 //! [`wifi_densepose_signal::BaselineCalibration::deviation`], whose
-//! `CalibrationDeviationScore` gives a per-frame amplitude z-score (presence
-//! strength).
+//! `CalibrationDeviationScore` gives per-frame amplitude z statistics
+//! (presence strength).
+//!
+//! **Presence is the per-frame p90 of |z|, NOT the median** (ADR-151
+//! presence-flatline bench, 2026-06-11): a person perturbs a *minority* of
+//! subcarriers strongly, so the median of |z| over all bins floors at
+//! `median(|N(0,1)|) ≈ 0.674` and never reaches presence gates — on real
+//! ESP32-C6 HE20 captures the median read empty ≈ 0.40–0.42 vs person-moving
+//! ≈ 0.74–0.75 (both below every gate), while the p90 separated cleanly:
+//! empty ≈ 1.27–1.33 vs person-moving ≈ 2.26–2.29.
 //!
 //! **Motion is NOT taken from the score's `motion_flagged`** (ADR-152 finding,
 //! "z-band squeeze"): that flag fires on `amplitude_z_median > 2.0` — deviation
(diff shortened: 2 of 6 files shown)


ruvnet:
✅ **Reviewed — substantive, measured fix.** Accepting an occupied room as the `empty` anchor is a real ADR-151 correctness defect, and the p90|z|-vs-median|z| reasoning is sound (a partially-occupied room moves the 90th percentile long before the median). Appreciate that it's bench-reproduced on real ESP32-C6 HE20 captures rather than theorized, and that the gates are re-derived from measured data + the unsanitised phase channel is excluded from motion. Touches calibration core (anchor/enrollment/signal) so it warrants a careful read, but the approach is principled. 👍

jkuscevic:
**Field confirmation on `v2051` / current `main` — the median-|z| gate is still the shipping statistic, and it also rejects every person anchor.**

This PR (unmerged as of 2026-08-19; `enrollment.rs` blob identical on tags `v2051` and `main`) documents the empty-accepts-occupied failure. We hit the other side of the same statistic on different hardware and are putting the numbers here rather than opening a duplicate.

### Setup

- CLI / calibration crate at tag `v2051` (enrollment path unmodified)
- 5× M5Stack Atom S3 Lite (`esp32-csi-node`), HT20, 52 active subcarriers
- ADR-135 empty-room baselines taken the previous evening (clean captures, 240 frames/node)
- Guided enrollment the next day: empty vs stand-still vs sit vs move, same links

### Captures

`presence_z` = mean over the capture of `CalibrationDeviationScore::amplitude_z_median`. Default gate `min_presence_z = 1.5`.

| condition | node | presence_z | verdict |
|---|---|---|---|
| empty | 1 | 0.60 | ACCEPTED |
| empty | 2 | 0.58 | ACCEPTED |
| stand still, room center | 1 | 0.83 | reject (`< 1.5`) |
| stand still, room center | 2 | 0.48 | reject |
| stand still, ~0.5 m from the node | 1 | 0.46 | reject |
| stand still, ~0.5 m from the node | 2 | 0.60 | reject |
| sit, several metres from the node | 2 | 0.62 | reject |
| continuous movement, same spot | 2 / 1 | 0.54 / 0.69 | reject |
| sit at a desk next to the AP | 3 | **0.43** | reject |

Person never left the empty band (0.43–0.83). Empty itself measured 0.32–0.88 on the prior evening's baselines. The lowest person value is the desk sit — occupied scored *below* empty.

`motion_rate` was 0.59–0.85 in every condition, empty included, so the still/move limb of the gate does not discriminate either.

Lowering `min_presence_z` does not help: person and empty are not separable on this statistic at these distances. The canned reason `"move closer / face the sensor"` fired even at ~0.5 m.

### Mechanism (same as this PR, still the code on `v2051`)

`AnchorRecorder::presence_z` is the mean of `amplitude_z_median` (`enrollment.rs:165–187`). That field is `median_abs` over the per-subcarrier amplitude z vs the empty baseline (`ruvsense/calibration.rs::deviation` ~297). A person at room scale perturbs a localized subset of the 52 HT20 subcarriers; the cross-subcarrier median is robust against exactly that. Defaults `min_presence_z = 1.5` / `empty_max_z = 1.0` sit above what the statistic produces here.

Not proposing a competing metric — this PR already has one. Just confirming the defect is still the shipping gate on `v2051`/`main`, and that on Atom S3 Lite HT20 links it also makes person-anchor enrollment impossible (7/7 person captures rejected; only `empty` accepted). Happy to provide more raw numbers if useful.

Issue #1107, opened 2026-06-16
ESP32-S3 DevKitC-1 N16R8: MR60BHA2 false detection causes ENOMEM + yield=0pps


brotak123:
**Hardware:** ESP32-S3-DevKitC-1 N16R8 (8MB PSRAM, 16MB Flash)
**Firmware:** v0.6.5 (downloaded from v0.6.6-esp32 release page)
**OS:** Windows 11

**Problem:**
1. mmWave MR60BHA2 falsely detected on UART1 — no mmWave hardware present
2. Continuous `sendto ENOMEM` in stream_sender
3. `yield=0pps`, `motion=0.00`, `presence=0.00` — no CSI data processed

**Tried:**
- edge-tier 0, 1, 2 — all same result
- filter-mac set to AP MAC address
- pres-thresh 0 (auto-calibrate)

**Serial log:**
I (3238) mmwave: Detected MR60BHA2 at 115200 baud (caps=0x000f)

W stream_sender: sendto ENOMEM — backing off for 100 ms

I adaptive_ctrl: medium tick: state=8 yield=0pps motion=0.00 presence=0.00 rssi=-62

ruvnet:
### Problem 1 (false mmWave detection) — root-caused + fixed in #1119

`probe_at_baud()` counted bare `0x01` (`MR60_SOF`) bytes and declared MR60BHA2 on **one** of them (a `mr60_sof_seen > 0` fallback). With **no sensor on UART1**, the floating line reads noise full of `0x01`s → `Detected MR60BHA2 (caps=0x000f)`. **PR #1119** now requires a validated 8-byte header (SOF + correct header checksum + a known frame type `0x0A__`/`0x0F09`) clearing the ≥3-frame threshold, with no weak fallback. Real sensors stream valid frames, so present hardware still detects.

### Problem 2 (`sendto ENOMEM` + `yield=0pps`) — likely separate, needs one check

`sendto ENOMEM` from `stream_sender` is the classic signature of an **unreachable / wrong target IP**: the UDP send queue can't drain, LWIP runs out of network buffers, and the 100 ms back-off then starves the CSI send path (`yield=0pps`). This is the *same* failure mode as streaming to a dead aggregator.

Could you confirm, with the sensor disconnected and #1119 applied:
1. Is the provisioned `target_ip` actually reachable from the ESP32 (`ping` it from the same LAN), and is something listening on the UDP `target_port`?
[...]

PR #1159, opened 2026-06-22
fix(firmware): phantom LD2410 detection + ENOMEM backoff (#1135)

### LINKED ISSUE #1135: stream_sender: sendto ENOMEM in a tight loop on ESP32-S3 (v0.8.1-esp32) — 0 UDP frames ever leave the node
# `stream_sender: sendto ENOMEM` in a tight loop on ESP32-S3 (v0.8.1-esp32) — 0 UDP frames ever leave the node

## Summary
A fresh ESP32-S3 (8 MB, no peripherals attached) flashed with the **v0.8.1-esp32** release assets enters a permanent `sendto ENOMEM` loop on the second CSI callback. The aggregator's source state never advances past `esp32:offline`. While diagnosing this I also found what looks like a separate **phantom LD2410 detection** on a floating UART — I'm filing both here since they were observed in the same run; happy to split into two issues if preferred.

## Environment
| | |
|---|---|
| Board | ESP32-S3, QFN56 rev v0.2, 8 MB embedded PSRAM, USB-Serial/JTAG |
| Flash | 8 MB, DIO @ 80 MHz |
| Firmware | **v0.8.1-esp32** release assets (`bootloader.bin`, `partition-table.bin`, `ota_data_initial.bin`, `esp32-csi-node-s3-8mb.bin`) at `0x0 / 0x8000 / 0xf000 / 0x20000` |
| esptool | 5.3.0 (host: macOS) |
| Aggregator | `ruvnet/wifi-densepose:latest` in Docker, UDP `0.0.0.0:5005` mapped to host |
| Wi-Fi | WPA2-PSK, 2.4 GHz, ch 10, RSSI -39 to -42 dBm |
| Network | ESP32 ↔ host on same /24, ICMP RTT ~12 ms |
| mmWave sensor | **None physically connected** (UART1 TX/RX pins floating) |
| NVS | Wiped before each test (`esptool erase-region 0x9000 0x6000`) and re-provisioned via the bundled `provision.py` |

## What works
- Wi-Fi associates, DHCP returns an IP, host↔ESP32 ICMP is healthy.
- CSI capture itself is alive: callbacks fire at expected cadence with valid `len` and `rssi` fields.
- The aggregator's UDP receiver works: injecting a hand-built UDP packet at `127.0.0.1:5005` from the host immediately promotes its source from `simulated` to `esp32`. So the failure is strictly on the node's egress path.

## What fails
Annotated boot log (timestamps in ms since reset; nothing else is talking on this LAN):

```
I (4124) main: Got IP: 192.168.0.19
I (4124) stream_sender: UDP sender initialized: 192.168.0.25:5005
I (4144) csi_collector: WiFi modem sleep disabled (WIFI_PS_NONE) for CSI capture
I (4154) wifi:ic_enable_sniffer
I (4154) csi_collector: Promiscuous mode enabled (MGMT-only, RuView#396)
I (4164) csi_collector: self-ping started -> 192.168.0.1 @50Hz (CSI OFDM source, fix #521/#954)
I (4184) ESPNOW: espnow [version: 2.0] init
I (4194) edge_proc: Initializing edge processing (tier=2, top_k=8, vital_interval=1000ms, ...)
I (4294) mmwave: Probing UART1 (TX=17, RX=18) for mmWave sensor...
I (4304) mmwave: Probing at 115200 baud (MR60BHA2)...
I (4494) csi_collector: CSI cb #1: len=128 rssi=-25 ch=10        ← first send succeeds (no log = OK)
I (5364) mmwave: Probing at 256000 baud (LD2410)...
I (5544) csi_collector: CSI cb #2: len=128 rssi=-25 ch=10
W (5544) stream_sender: sendto ENOMEM — backing off for 100 ms   ← first failure on CSI cb #2
W (5544) csi_collector: sendto failed (fail #1)
I (5564) mmwave: Detected LD2410 at 256000 baud (caps=0x000c)    ← (see "bug #2" below)
I (5564) mmwave: mmWave UART task started (type=LD2410)
W (5564) stream_sender: sendto suppressed (ENOMEM backoff, 1 dropped)
... (steady-state — every send either ENOMEMs or is suppressed)
```

The aggregator stays at `{"source":"esp32:offline"}` indefinitely; **zero CSI frames reach it over the network** even though L2/L3 is healthy.

### Bug #1 — primary: permanent `sendto ENOMEM` from CSI cb #2 onward
The first `stream_sender_send` (on CSI cb #1) appears to succeed (no failure log). The very next one fails with `ENOMEM` and never recovers — every subsequent attempt either ENOMEMs or is suppressed by the 100 ms backoff. The 100 ms backoff is shorter than what's needed for the underlying pbuf/mbox pressure to clear, so the node is stuck.

The 1050 ms gap between cb #1 and cb #2 is occupied by:
- the **50 Hz self-ping** to the gateway (`csi_collector: self-ping started ... @50Hz`) — that's ~52 UDP datagrams enqueued back-to-back into LWIP;
- the **MR60BHA2 UART probe** at 115200 baud for ~1060 ms;
- **ESPNOW** init + `c6_espnow` tx loop;
- **promiscuous + sniffer** RX consuming Wi-Fi RX buffers.

It looks like LWIP pbufs / WiFi dynamic TX buffers / UDP send mbox saturate during that 1 s and never drain. `sdkconfig.defaults` already mentions a sibling fix for an earlier ENOMEM (note above `CONFIG_LWIP_UDP_RECVMBOX_SIZE=32` / `CONFIG_LWIP_TCPIP_RECVMBOX_SIZE=64` / `CONFIG_ESP_WIFI_DYNAMIC_TX_BUFFER_NUM=64`), but on S3 those values don't appear sufficient — possibly because S3 + Sniffer + 50 Hz self-ping + ESPNOW competes harder for buffers than the C6 target the 0.6.7 build was verified against.

Possible fixes worth considering:
- Drop the self-ping cadence (50 Hz → 10 Hz?) when the LD2410/mmWave or ESPNOW tasks are also TX-active.
- Raise `CONFIG_ESP_WIFI_DYNAMIC_TX_BUFFER_NUM` / `CONFIG_LWIP_TCPIP_RECVMBOX_SIZE` further in the S3-specific sdkconfig overlays.
- When `stream_sender` has been in ENOMEM backoff for >N consecutive cycles, exponentially extend the backoff (the current fixed 100 ms is too short) and emit a single warning instead of one per attempt.

### Bug #2 — secondary: phantom LD2410 detection on a floating UART
With **no mmWave sensor wired to UART1 (TX=17, RX=18)**, the firmware still concludes `Detected LD2410 at 256000 baud (caps=0x000c)` and spawns the LD2410 reader task. The v0.8.1-esp32 release notes specifically called out a fix for "false MR60BHA2 detection → ENOMEM by requiring validated sensor headers instead of accepting bare byte patterns" — the **LD2410 path looks like it still accepts loose patterns** and so trips on floating-pin noise at 256000 baud.

This isn't the trigger of Bug #1 (the timing rules it out — first ENOMEM at 5544 ms, LD2410 declared at 5564 ms), but the resulting mmWave UART task adds steady load to a system that's already in a fragile buffer state.

Suggested fix: gate `mmwave: Detected LD2410` on a validated frame header (length + checksum + magic), matching what was done for MR60BHA2 in v0.8.1.

## What I tried
1. `release_bins/s3-adr110/` in-tree bins — same ENOMEM loop.
2. `release_bins/s3-fair-adr110/` in-tree bins — same.
3. Fresh download of **v0.8.1-esp32** release assets — same.
4. `esptool erase-region 0x9000 0x6000` to wipe NVS, then `provision.py --reset --edge-tier 2 --target-ip <host> --target-port 5005` — same.
5. Confirmed Wi-Fi credentials, IP, gateway, and aggregator IP/port are correct (ping host↔ESP32 OK).
6. Confirmed the aggregator's UDP receiver works by sending a synthetic CSI packet from the host — source promoted to `esp32` immediately, then back to `esp32:offline` after the synthetic stream stops.

## Repro
1. Flash a bare ESP32-S3 (8 MB, **no mmWave sensor connected**) with the v0.8.1-esp32 release assets at `0x0 / 0x8000 / 0xf000 / 0x20000`.
2. `python3 provision.py --port <port> --chip esp32s3 --ssid <SSID> --password <pw> --target-ip <host> --target-port 5005 --edge-tier 2 --reset`.
3. Run RuView aggregator on `<host>:5005`.
4. Watch ESP32 serial: first `stream_sender: sendto ENOMEM — backing off for 100 ms` appears on CSI cb #2 and never goes away. Phantom `mmwave: Detected LD2410 ...` appears in the same window.
5. Watch `GET /api/v1/status` on the aggregator — stays `esp32:offline` indefinitely.

Happy to test patches on this board if useful.

ruvnet:
Fixes #1135 (against the v0.8.1-esp32 release).

## Bug #2 — phantom LD2410 on a floating UART (root cause)
LD2410 probe-detection matched only the 4-byte head `0xF4F3F2F1`, so floating-pin noise at 256000 baud could phantom-detect a sensor and spawn a UART task. Now requires a **full validated report frame** — head + sane intra-frame length + matching tail `0xF8F7F6F5` — the same validate-before-trust approach used for MR60BHA2 in #1107. Extracted to `mmwave_detect.h` and shared with a host unit test so firmware and test can't diverge.

## Bug #1 — sendto ENOMEM stuck loop
Fixed 100 ms backoff was too short to drain sustained lwIP/WiFi buffer pressure. Now **exponential** (100→200→…→2000 ms per consecutive ENOMEM, reset on first successful send). Eliminating the phantom LD2410 task (bug #2) also removes the extra load that tipped the reporter's tier-2 node into the stuck state.

## Validation
- **Host unit test** (`test_mmwave_detect.c`, 8 vectors) — tests the *real* predicate; the #1135 case (head magic **without** valid tail) is REJECTED. Wired into the Makefile + `host_tests` (runs gcc in CI). Compiles clean against the header; truth table proven.
- **Hardware** (ESP32-S3 QFN56 rev v0.2 — the reporter's silicon): tier-2 streams ~100 frames/s, **no stuck ENOMEM**, correctly reports **no mmWave** (no phantom), no regression.
- **Honest limit:** I couldn't reproduce the reporter's environment-specific floating-pin noise, so the deterministic proof for the phantom case is the unit test, not a hardware repro.

### DIFF
--- a/firmware/esp32-csi-node/main/mmwave_detect.h
+++ b/firmware/esp32-csi-node/main/mmwave_detect.h
@@ -1,0 +1,37 @@
+/**
+ * @file mmwave_detect.h
+ * @brief Pure (host-testable) mmWave frame-validation predicates for probe-time
+ *        sensor detection. No ESP-IDF deps — safe to #include in a host unit test.
+ *
+ * Detection must validate a *full* frame, never a bare header byte/pattern: a
+ * floating UART with no sensor reads line noise that can contain header-looking
+ * bytes, which the old loose checks mistook for a real sensor (#1107 MR60,
+ * #1135 LD2410). These predicates are the validate-before-trust gate.
+ */
+#ifndef MMWAVE_DETECT_H
+#define MMWAVE_DETECT_H
+
+#include <stdint.h>
+#include <stdbool.h>
+
+/**
+ * True iff buf[i..] begins a *validated* LD2410 report frame within [0,len):
+ *   F4 F3 F2 F1 | len(LE,2) | data[len] | F8 F7 F6 F5
+ * Requires the head magic, a sane intra-frame length, AND the matching tail at
+ * head+6+len. Pure noise that merely contains 0xF4F3F2F1 fails the tail check.
+ */
+static inline bool mmwave_ld2410_valid_at(const uint8_t *buf, int i, int len)
+{
+    if (i < 0 || i + 5 >= len) return false;
+    if (!(buf[i] == 0xF4 && buf[i+1] == 0xF3 && buf[i+2] == 0xF2 && buf[i+3] == 0xF1))
+        return false;
+    uint16_t flen = (uint16_t)buf[i+4] | ((uint16_t)buf[i+5] << 8);
+    /* Real LD2410 report frames are small (basic=13, engineering=35). */
+    if (flen < 1 || flen > 64) return false;
+    int tail = i + 6 + (int)flen;
+    if (tail + 3 >= len) return false;
+    return buf[tail] == 0xF8 && buf[tail+1] == 0xF7
+        && buf[tail+2] == 0xF6 && buf[tail+3] == 0xF5;
+}
+
+#endif /* MMWAVE_DETECT_H */
--- a/firmware/esp32-csi-node/main/mmwave_sensor.c
+++ b/firmware/esp32-csi-node/main/mmwave_sensor.c
@@ -26,6 +26,7 @@
  */
 
 #include "mmwave_sensor.h"
+#include "mmwave_detect.h"
 
 #include <string.h>
 #include <math.h>
@@ -401,10 +402,12 @@ static mmwave_type_t probe_at_baud(uint32_t baud)
                     }
                 }
             }
-            /* LD2410: 4-byte header 0xF4F3F2F1 (already specific enough). */
-            if (i + 3 < len && buf[i] == 0xF4 && buf[i+1] == 0xF3
-                && buf[i+2] == 0xF2 && buf[i+3] == 0xF1
-                && baud == MMWAVE_LD2410_BAUD) {
+            /* LD2410: require a *full validated* report frame, not just the
+             * 4-byte head. A floating UART1 at 256000 baud can emit the head
+             * pattern 0xF4F3F2F1 from line noise (#1135 bug #2). The shared
+             * predicate (host-unit-tested in mmwave_detect.h) demands a sane
+             * intra-frame length AND the matching tail 0xF8F7F6F5. */
+            if (baud == MMWAVE_LD2410_BAUD && mmwave_ld2410_valid_at(buf, i, len)) {
                 ld2410_header_seen++;
             }
         }
--- a/firmware/esp32-csi-node/main/stream_sender.c
+++ b/firmware/esp32-csi-node/main/stream_sender.c
@@ -26,9 +26,16 @@ static struct sockaddr_in s_dest_addr;
  * rapid-fire CSI callbacks can exhaust the pbuf pool and crash the device.
  */
 static int64_t s_backoff_until_us = 0;       /* esp_timer timestamp to resume */
-#define ENOMEM_COOLDOWN_MS  100              /* suppress sends for 100 ms */
+#define ENOMEM_COOLDOWN_MS  100              /* base backoff; doubles per streak */
+#define ENOMEM_COOLDOWN_MAX_MS 2000          /* cap on the exponential backoff */
 #define ENOMEM_LOG_INTERVAL 50               /* log every Nth suppressed send */
 static uint32_t s_enomem_suppressed = 0;
+/* Consecutive ENOMEM episodes without an intervening successful send. A fixed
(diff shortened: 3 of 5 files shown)


Kdm78 (commented):
Plain text

PR #1176, opened 2026-06-25
fix(sensing-server): canonicalize per-node CSI to 56 tones before multistatic fusion (#1170)

### LINKED ISSUE #1170: Multistatic fusion DimensionMismatch flood: live bridge feeds raw (un-canonicalized) per-node CSI to the fuser
## Summary

On a mixed ESP32 mesh, the live sensing server logs an escalating flood of
`governed trust cycle failed … fusion error: Dimension mismatch` warnings, and
`total_engine_errors` climbs unbounded (thousands within seconds). **Multistatic
fusion never actually runs** — it rejects every cycle and silently falls back to
per-node sum/dedup — because per-node CSI frames reach `MultistaticFuser::fuse()`
at their **raw** subcarrier counts (64 / 128 / 192) instead of being normalized
onto the canonical 56-tone grid first.

Anchored at `7831f294`.

## Observed symptoms

```
WARN sensing_server::engine_bridge: governed trust cycle failed (warn rate-limited to one per 10s):
  fusion error: Dimension mismatch: node 1 has 64 subcarriers, expected 128  total_engine_errors=1
WARN sensing_server::engine_bridge: governed trust cycle failed (warn rate-limited to one per 10s):
  fusion error: Dimension mismatch: node 3 has 192 subcarriers, expected 128  total_engine_errors=585
... total_engine_errors=1278 ... 1862 ... 2586 ... 3273 ... 3841 (climbing every cycle)
```

At least three distinct subcarrier counts are present in one cycle: node 0 = 128
(it becomes "expected"), node 1 = 64, node 3 = 192. These are legitimate ESP32
capture-mode differences (HT20 ≈ 64 bins; HT40 ≈ 128/192 bins), not corrupt data.

## Root cause

The error originates in `MultistaticFuser::fuse()`
(`v2/crates/wifi-densepose-signal/src/ruvsense/multistatic.rs`). It takes the
**first** node's amplitude length as the expected dimension and rejects any node
that differs:

```rust
// validate dimension consistency
let n_sub = amplitudes[0].len();              // <- "expected" = first node's count
for (i, amp) in amplitudes.iter().enumerate().skip(1) {
    if amp.len() != n_sub {
        return Err(MultistaticError::DimensionMismatch {
            node_idx: i, expected: n_sub, got: amp.len(),
        });
    }
}
```

That check is correct *by design*: the fuser's own docs state it operates on
**canonical-56** frames that `hardware_norm.rs` resamples onto a uniform grid
(see the `with_cir_canonical56` doc comment and ADR-154). The bug is that the
live server path never performs that normalization.

`node_frame_from_state()` in
`v2/crates/wifi-densepose-sensing-server/src/multistatic_bridge.rs` wraps each
node's **raw** amplitude vector directly into a struct *named* `CanonicalCsiFrame`
— without ever calling `HardwareNormalizer`:

```rust
let amplitude: Vec<f32> = latest.iter().map(|&v| v as f32).collect(); // raw length: 64/128/192…
let n_sub = amplitude.len();
let phase = vec![0.0_f32; n_sub];
let canonical = CanonicalCsiFrame { amplitude, phase, hardware_type: HardwareType::Esp32S3 };
```

So heterogeneous-count frames flow straight into the fuser and trip
`DimensionMismatch` on every cycle.

### Why it affects both fusion paths

`node_frames_from_states()` is the single source of frames for **both**:

1. the bare fusion path — `fuse_or_fallback()` (drops to per-node sum/dedup on error), and
2. the governed engine path — `engine_bridge.rs` → `StreamingEngine::process_cycle()`,

so the one bridge defect surfaces as the engine-error flood *and* silently
disables real multistatic fusion in the unguarded path.

## Impact

- Multistatic attention-weighted fusion never runs on any mesh whose nodes report
  different subcarrier counts (the common case for mixed HT20/HT40 ESP32 nodes).
- `total_engine_errors` grows without bound; the governed trust cycle produces no
  witnessed belief, so the trust/provenance plane is effectively dead on the live
  wire for such deployments.
- Log noise (one warn / 10 s, but a persistent error condition).
- No crash and no incorrect *published* output (the bare path falls back), so the
  failure is silent unless logs or `GET /api/v1/status` (`engine_error_count`) are
  inspected.

## Proposed solution

Canonicalize every node frame onto the 56-tone grid in the bridge, before fusion.
Use a **resample-only** transform (not full `HardwareNormalizer::normalize()`):
z-scoring drives the amplitude mean to ~0, which would saturate the downstream
person-score (`compute_person_score_from_amplitudes`, a squared coefficient of
variation `variance / mean²`). Length-only canonicalization makes frames fusable
while preserving the amplitude statistics downstream code relies on.

1. Add a resample-only helper to `HardwareNormalizer` (`hardware_norm.rs`):

   ```rust
   /// Resample a raw 1-D CSI vector onto the canonical subcarrier grid
   /// WITHOUT z-score normalization (preserves amplitude scale).
   pub fn resample_to_canonical(&self, raw: &[f64]) -> Vec<f64> {
       resample_cubic(raw, self.canonical_subcarriers)
   }
   ```

2. In `node_frame_from_state()` (`multistatic_bridge.rs`), resample the raw
   amplitude to 56 tones via a shared normalizer before building the frame, so
   every node presents a uniform dimension to the fuser.

This aligns the live path with the canonical-56 pipeline the fuser already
expects (ADR-154) and lets multistatic fusion actually run on a mixed mesh.

### Safety notes

- No CIR estimator is attached to any fuser on the server path (verified: no
  `set_cir_estimator`/`with_cir` calls), so 56-tone frames cannot trip the
  ADR-154 `SubcarrierMismatch` debug-assert path.
- Resampling is a pure function of the input, so engine witnesses stay
  deterministic/reproducible.

## Status

A fix following the above has been implemented and verified locally:

- `wifi-densepose-signal` — `hardware_norm` tests: **14/14 pass**
- `wifi-densepose-sensing-server` (bin) — `multistatic_bridge` / `engine_bridge` /
  `field_bridge` / `tracker_bridge` tests: **28/28 pass**, including a new
  regression test (`heterogeneous_node_counts_canonicalize_and_fuse`) proving a
  mixed 64/192-count cycle now fuses instead of erroring.

One pre-existing test (`engine_bridge::observe_cycle_counts_engine_errors`) was
retargeted: its 56-vs-30 subcarrier setup now canonicalizes and fuses cleanly, so
it was switched to force a `TimestampMismatch` (spread > hard guard) to keep
exercising the engine error-counting / last-good-witness wiring.

Tovli:
## Summary

Fixes #1170 — the `DimensionMismatch` flood on a mixed ESP32 mesh.

`node_frame_from_state()` (`multistatic_bridge.rs`) wrapped each node's **raw**
amplitude (HT20 ≈ 64, HT40 ≈ 128/192 tones) into a `CanonicalCsiFrame` *without
ever resampling*. `MultistaticFuser::fuse()` takes the first node's length as the
expected dimension and rejects any node that differs, so on a heterogeneous mesh:

- the governed engine path logged an escalating `governed trust cycle failed …
  fusion error: Dimension mismatch` and `total_engine_errors` climbed unbounded;
- the bare path silently fell back to per-node sum/dedup — **multistatic fusion
  never actually ran**.

## Fix

Canonicalize every node frame onto the 56-tone grid in the bridge, **before**
fusion, using a **resample-only** transform (not full `normalize()` — z-scoring
drives the amplitude mean to ~0 and would saturate the downstream person-score
`variance / mean²`). This aligns the live path with the canonical-56 pipeline the
fuser already documents (ADR-154).

- **`hardware_norm.rs`** — add `HardwareNormalizer::resample_to_canonical()`:
  cubic resample to the canonical grid without z-score; identity passthrough when
  already canonical.
- **`multistatic_bridge.rs`** — resample raw amplitude to 56 tones via a shared
  `NORMALIZER` in `node_frame_from_state()`, so **both** fusion paths
  (`fuse_or_fallback` and the governed engine) see uniform dimensions.

## Safety

- No CIR estimator is attached on the server path (no `set_cir_estimator`/
  `with_cir`), so 56-tone frames cannot trip the ADR-154 `SubcarrierMismatch`
  debug-assert.
- Resampling is a pure function of the input → engine witnesses stay
  deterministic/reproducible.

## Tests

`cargo test -p wifi-densepose-signal -p wifi-densepose-sensing-server --no-default-features` — **0 failed**.

- New regression test `heterogeneous_node_counts_canonicalize_and_fuse`: a mixed
  64/192-count cycle now **fuses** (`fused.is_some()`, no fallback) instead of
  erroring.
- `hardware_norm` resample/normalize tests pass.
- Retargeted `engine_bridge::observe_cycle_counts_engine_errors`: its old 56-vs-30
  setup now canonicalizes and fuses cleanly, so it forces a `TimestampMismatch`
  (spread > hard guard) to keep exercising the engine error-counting /
  last-good-witness wiring.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/engine_bridge.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/engine_bridge.rs
@@ -247,7 +247,7 @@ impl EngineBridge {
 mod tests {
     use super::*;
     use std::collections::VecDeque;
-    use std::time::Instant;
+    use std::time::{Duration, Instant};
     use wifi_densepose_bfld::PrivacyClass;
 
     fn node_state_with_history(amp: f64, n_sub: usize) -> NodeState {
@@ -402,17 +402,34 @@ mod tests {
         assert!(!bridge.suppress_raw_outputs());
     }
 
-    /// Error wiring (review finding 1a): two live nodes with mismatched
-    /// subcarrier counts make fusion return a `DimensionMismatch` →
-    /// `EngineError` — previously dropped by `if let Some(Ok(..))` at the
-    /// call sites. The counter must increment and the last good trust state
-    /// must survive a later failure.
+    /// Error wiring (review finding 1a): a fusion error must increment the
+    /// engine-error counter (previously dropped by `if let Some(Ok(..))` at the
+    /// call sites) and leave the last good trust state intact.
+    ///
+    /// The error is forced via a timestamp spread beyond the hard guard
+    /// (`MultistaticError::TimestampMismatch`). A subcarrier-count mismatch can
+    /// no longer reach the fuser — `node_frames_from_states` canonicalizes every
+    /// node onto a uniform 56-tone grid upstream — so the previous 56-vs-30
+    /// setup would now fuse cleanly.
     #[test]
     fn observe_cycle_counts_engine_errors() {
         let mut bridge = EngineBridge::new(PrivacyMode::PrivateHome, 1, "r", "R");
+        // Two fresh 56-tone nodes captured 200 ms apart → spread exceeds the
+        // 60 ms hard guard → fusion errors. Both times are derived from one
+        // base instant with node 1 offset *forward*: the bridge's frame
+        // timestamps are `last_frame_time.duration_since(EPOCH)`, and
+        // `Instant::duration_since` saturates to zero for any instant predating
+        // the lazily-initialized `EPOCH`. Offsetting forward keeps both times
+        // ≥ EPOCH, so the 200 ms spread survives regardless of EPOCH timing.
+        // Neither node is stale (< 10 s).
+        let base = Instant::now();
         let mut mismatched = HashMap::new();
-        mismatched.insert(0u8, node_state_with_history(1.0, 56));
-        mismatched.insert(1u8, node_state_with_history(1.05, 30)); // 30 ≠ 56 subcarriers
+        let mut a = node_state_with_history(1.0, 56);
+        a.last_frame_time = Some(base);
+        let mut b = node_state_with_history(1.05, 56);
+        b.last_frame_time = Some(base + Duration::from_millis(200));
+        mismatched.insert(0u8, a);
+        mismatched.insert(1u8, b);
 
         assert!(bridge.observe_cycle(&mismatched, 1_000).is_none());
         assert_eq!(bridge.engine_error_count(), 1);
--- a/v2/crates/wifi-densepose-sensing-server/src/multistatic_bridge.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/multistatic_bridge.rs
@@ -10,7 +10,7 @@ use std::collections::HashMap;
 use std::sync::LazyLock;
 use std::time::{Duration, Instant};
 
-use wifi_densepose_signal::hardware_norm::{CanonicalCsiFrame, HardwareType};
+use wifi_densepose_signal::hardware_norm::{CanonicalCsiFrame, HardwareNormalizer, HardwareType};
 use wifi_densepose_signal::ruvsense::multiband::MultiBandCsiFrame;
 use wifi_densepose_signal::ruvsense::multistatic::{FusedSensingFrame, MultistaticFuser};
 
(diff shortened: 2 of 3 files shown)


Tovli:
hendled via https://github.com/ruvnet/RuView/pull/1193

Issue #432, opened 2026-04-26
RaBitQ similarity sensor: 43-51x faster pose/CSI matching, with anomaly detection, privacy logs, and mesh compression for free


ruvnet:
## What this is

RuView now has a **cheap similarity sensor** baked into the pipeline — a way to ask "have I seen something like this before?" for any embedding (poses, CSI features, room signatures) without paying the full floating-point cost. It uses a technique called **RaBitQ-style binary sketching**: each embedding gets compressed to one bit per dimension (32× smaller in memory) and compared with a single CPU instruction (POPCNT on Intel/AMD, NEON `vcnt` on ARM/Pi/Mac).

The architectural decision is **[ADR-084](docs/adr/ADR-084-rabitq-similarity-sensor.md)** (merged on `main`, status: Proposed). The first implementation pass — the foundation `Sketch` / `SketchBank` API — is on branch `feat/adr-084-pass-1-sketch-module` (commits `6fd5b7d`, `1df9d5f7d`).

## Why this matters in plain language

A lot of what RuView does is the same shape of question over and over:

- "Is this the same person we were tracking a moment ago?" — AETHER re-identification
- "Is this room behaving like it normally does, or is something unusual happening?" — novelty detection
- "Have we recorded a similar CSI signature before?" — recording search
- "Does this mesh node need to send the full sensor data, or can we just say 'same as last time'?" — bandwidth saving

In every case, the system answered by **comparing full floating-point vectors**, which is slow, cache-unfriendly, and means storing the raw vectors forever. RaBitQ collapses that comparison to a single hardware instruction over a 32×-smaller fingerprint. We can keep the *witness* of what was seen without keeping the *signal*.

## New capabilities

| Capability | What it unlocks |
|---|---|
| **Always-on novelty detection** | The heavy CNN / pose model only wakes up when something genuinely new happens. Energy budget per node drops noticeably during quiet rooms. |
| **Faster re-identification** | When 3+ ESP32 nodes are streaming, the tracker can pre-filter candidate matches before running the full Kalman / cosine pass. Targets the ghost-skeleton class of issues we recently fixed in #420. |
| **Mesh-exchange compression** | Inter-cluster broadcasts can carry sketches + witness hashes instead of full embeddings. Less RF traffic, lower bandwidth bills on metered backhauls. |
[...]

PR #1254, opened 2026-07-06
feat(adr-084): Pass 4 mesh-exchange compression + Pass 5 privacy-preserving event log

compiledbyutkarsh:
## Summary

Implements the two remaining passes from #432 (ADR-084 RaBitQ similarity sensor):

- **Pass 4 — Mesh-exchange compression**: the ESP32 swarm bridge (ADR-066) now sketches each happiness vector to a 1-byte sign fingerprint and compares it against the last vector actually sent. Below-threshold novelty suppresses the ingest POST; a `max_suppress_sec` floor guarantees the Seed's view never goes stale indefinitely. Wire format to the Cognitum Seed is unchanged — no Seed-side changes needed.

- **Pass 5 — Privacy-preserving event log**: the cluster-Pi novelty sensor (Pass 3) now feeds every scored frame into a `PrivacyEventLog` audit trail — `(sketch_bytes, sketch_version, novelty, witness_sha256)` instead of raw embeddings. The existing full-precision `EmbeddingHistory` bank is untouched; this adds a durable, non-invertible trail alongside it.

## Changes

- `firmware/esp32-csi-node/main/swarm_bridge.{c,h}` — novelty-gated suppression for the happiness-ingest cycle, heartbeats unaffected
- `v2/crates/wifi-densepose-sensing-server/src/main.rs` — `NodeState` gains a fixed-capacity (256-event) `PrivacyEventLog`, populated in `update_novelty()`
- `v2/crates/wifi-densepose-sensing-server/Cargo.toml` — `wifi-densepose-ruvector` promoted to a direct dependency

## Testing

cargo check -p wifi-densepose-sensing-server --no-default-features   (clean build)
cargo test -p wifi-densepose-sensing-server --no-default-features    (all green)
cargo test -p wifi-densepose-ruvector --no-default-features          (156 passed, 0 failed)

No changes to wifi-densepose-ruvector's existing sketch/event-log primitives — both were already implemented; this PR wires them into their two remaining ADR-084 use sites.

### DIFF
--- a/firmware/esp32-csi-node/main/swarm_bridge.c
+++ b/firmware/esp32-csi-node/main/swarm_bridge.c
@@ -60,11 +60,23 @@ static uint32_t s_cnt_heartbeats;
 static uint32_t s_cnt_ingests;
 static uint32_t s_cnt_errors;
 
+/* ---- ADR-084 Pass 4: mesh-exchange compression state ----
+ * A 1-bit-per-dimension sign sketch of the last vector we actually sent
+ * over the wire. Only the ingest cycle (not the heartbeat) is gated —
+ * heartbeats stay unconditional because they're the Seed's liveness
+ * signal, not a happiness observation. */
+static uint8_t  s_last_sent_sketch;
+static bool     s_have_last_sketch;
+static uint32_t s_last_full_send_s;
+static uint32_t s_cnt_suppressed;
+
 /* ---- Forward declarations ---- */
 static void swarm_task(void *arg);
 static esp_err_t swarm_post_json(esp_http_client_handle_t client,
                                  const char *json, int json_len);
 static void swarm_get_ip_str(char *buf, size_t buf_len);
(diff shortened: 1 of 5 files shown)


PrinceAudre:
The firmware novelty-gating and the Rust `PrivacyEventLog` wiring look correct and self-consistent (the ruvector API calls match, and it's additive). One build-hygiene nit before merge:

The `v2/Cargo.lock` delta drops the `wifi-densepose-worldmodel` package entry (and a `wifi-densepose-core` dependency). That's a regeneration artifact from building without the `crates/worldgraph` submodule checked out — `worldmodel` lives in that submodule. Under `--locked` CI this lockfile will be out of sync and the build fails.

Fix: regenerate `v2/Cargo.lock` with submodules populated (`git submodule update --init`) so the `worldmodel`/`core` entries are preserved, and drop the spurious removal from the diff.

PR #1255, opened 2026-07-06
Wifi tracker

sahufagu30-byte:


PrinceAudre:
Was this opened intentionally? It adds a single stray file at the repo root whose contents are a copy of the project's own quickstart commands (the `docker pull` / `esptool … write_flash` / `pip install ruview` snippets), with no accompanying change. If it was a mistake, it's probably worth closing.

PR #1290, opened 2026-07-09
fix: HeartRateExtractor.extract() PyO3 binding misnames Rust phases param as weights

### LINKED ISSUE #1225: HeartRateExtractor.extract() always returns None with weights=[] — PyO3 binding misnames Rust `phases` param as `weights` (2.0.0a1)
### Summary

`HeartRateExtractor.extract()` (PyPI `wifi-densepose` / `ruview` **2.0.0a1**, PyO3 binding) returns `None` on **every** call when used per its own docstring (`weights=[]` = equal weights). The DSP itself works — the binding's second parameter is misnamed: the Rust core takes per-subcarrier **`phases`**, and an empty list zeroes the gate so nothing ever emits.

### Reproduction (macOS arm64, py3.11, wheel built from sdist)

```python
import math
from wifi_densepose import HeartRateExtractor

hr = HeartRateExtractor.esp32_default()   # 56 subcarriers, 100 Hz, 15 s window
for i in range(6000):                     # 60 s of frames
    t = i / 100.0
    residuals = [0.05 * math.sin(2 * math.pi * 1.2 * t)] * 56   # clean 72 BPM tone
    est = hr.extract(residuals=residuals, weights=[])           # per docstring
    assert est is None                     # never emits — passes for all 6000 frames
```

Tested variants that all return `None` with `weights=[]`: amps 0.02–0.1, noise 0→0.002, combined 0.25 Hz + 1.2 Hz signal, up to 60 s, `esp32_default()` and explicit `HeartRateExtractor(n_subcarriers=56, sample_rate=100.0, window_secs=15.0)`. `BreathingExtractor` works fine with `weights=[]` (recovered an injected 15.0 BPM exactly).

### Root cause

`v2/crates/wifi-densepose-vitals/src/heartrate.rs:90-94` (same in the 2.0.0a1 sdist):

```rust
pub fn extract(&mut self, residuals: &[f64], phases: &[f64]) -> Option<VitalEstimate> {
    let n = residuals.len().min(self.n_subcarriers).min(phases.len());
    if n == 0 {
        return None;
    }
```

The PyO3 binding `python/src/bindings/vitals.rs:270` exposes that `phases` parameter as `weights: Vec<f64>` and passes it straight through — so the documented `weights=[]` makes `phases.len() == 0` → `n == 0` → permanent `None`. `BreathingExtractor` genuinely treats its second arg as optional weights, which makes the shared docstring wording actively misleading for the heart path.

### Proof the DSP is fine

Passing any same-length phase array unblocks it and recovery is accurate:

```python
phases = [i * 0.001 for i in range(56)]
est = hr.extract(residuals=residuals, weights=phases)
# after ~5 s: VitalEstimate(value_bpm=72.29, confidence=0.944, status=Valid)  — injected 72.0
```

### Suggested fix (either works; (1) preserves signature)

1. In the heart binding, treat an empty second arg as neutral phases, matching breathing's empty-weights ergonomics: `let weights = if weights.is_empty() { vec![0.0; residuals.len()] } else { weights };` before the `allow_threads` call.
2. Or rename the binding param to `phases` and correct the docstring so callers know real per-subcarrier phase data is expected (and `[]` is invalid).

Related: ADR-117 / #785 (the PyO3 modernization this binding shipped under).

mvanhorn:
## Summary

`HeartRateExtractor.extract()` now takes a `phases` argument that matches the Rust core, so the documented call path returns real estimates instead of always returning `None`.

## Why this matters

Issue #1225 reports that `hr.extract(residuals=..., weights=[])` always returns `None`. The PyO3 binding named the second parameter `weights`, but the Rust core requires per-subcarrier unwrapped phases:

```rust
// v2/crates/wifi-densepose-vitals/src/heartrate.rs:100
pub fn extract(&mut self, residuals: &[f64], phases: &[f64]) -> Option<VitalEstimate> {
    let n = residuals.len().min(self.n_subcarriers).min(phases.len());
```

Because the sample count is clamped by `phases.len()`, passing the documented `weights=[]` sets `n = 0` and the extractor returns `None` for every frame. `HeartRateExtractor` is not `BreathingExtractor`: it needs phase data, so an empty second argument is never valid.

## Changes

- Rename the binding parameter `weights` -> `phases` and pass it straight through to the core, so the Python API name reflects what the DSP actually consumes.
- Update the docstring with a runnable example and a note that `phases=[]` is invalid for heart-rate extraction.
- Update the two keyword call sites in the test and benchmark suites.

## Testing

Added `test_heart_rate_extract_with_synthetic_signal_and_phases` (drives a synthetic 1.2 Hz / 72 BPM signal with matching phase data and asserts a finite BPM and a confidence in `[0, 1]`) and `test_heart_rate_extract_rejects_old_weights_keyword` (asserts the removed `weights=` keyword now raises `TypeError`). The parameter rename matches the core signature at `heartrate.rs:100`.

Fixes #1225

### DIFF
--- a/python/bench/test_bench_vitals.py
+++ b/python/bench/test_bench_vitals.py
@@ -57,12 +57,12 @@ def test_heart_rate_extract_per_frame_cost(benchmark) -> None:
     hr = HeartRateExtractor.esp32_default()
     rng = Random(43)
     for i in range(1500):
-        residuals, weights = _synth_frame(56, 100.0, i / 100.0, 1.2, rng)
-        hr.extract(residuals=residuals, weights=weights)
+        residuals, phases = _synth_frame(56, 100.0, i / 100.0, 1.2, rng)
+        hr.extract(residuals=residuals, phases=phases)
 
     def _one_frame():
-        residuals, weights = _synth_frame(56, 100.0, 16.0, 1.2, rng)
-        return hr.extract(residuals=residuals, weights=weights)
+        residuals, phases = _synth_frame(56, 100.0, 16.0, 1.2, rng)
+        return hr.extract(residuals=residuals, phases=phases)
 
     benchmark(_one_frame)
 
--- a/python/src/bindings/vitals.rs
+++ b/python/src/bindings/vitals.rs
@@ -242,7 +242,22 @@ impl PyBreathingExtractor {
 // ─── HeartRateExtractor ──────────────────────────────────────────────
 
 /// Extracts heart rate (40–120 BPM) from per-subcarrier amplitude
-/// residuals via 0.8–2.0 Hz bandpass + autocorrelation peak detection.
+/// residuals and per-subcarrier unwrapped phases (radians) via
+/// 0.8–2.0 Hz bandpass + autocorrelation peak detection.
+///
+/// Python:
+/// ```python
+/// from wifi_densepose import HeartRateExtractor
+///
+/// hr = HeartRateExtractor.esp32_default()  # 56 subcarriers, 100 Hz, 15s window
+///
+/// # Feed residuals and matching unwrapped phases from your preprocessor.
+/// # Unlike BreathingExtractor weights, phases=[] is invalid for heart-rate
+/// # extraction because the Rust core requires phase data for each subcarrier.
+/// est = hr.extract(residuals=[0.01, -0.02, …], phases=[0.0, 0.01, …])
+/// if est is not None:
+///     print(est.value_bpm, est.confidence)
+/// ```
 #[pyclass(name = "HeartRateExtractor")]
 pub struct PyHeartRateExtractor {
     inner: HeartRateExtractor,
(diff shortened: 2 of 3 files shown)


PrinceAudre:
The diagnosis and fix here are correct — the PyO3 `HeartRateExtractor.extract()` binding did misname the Rust `phases` parameter as `weights` (core `heartrate.rs` takes `phases`; an empty 2nd arg forced `n=0` → always `None`), and this fixes it consistently (breathing correctly stays `weights`).

However, this change is **already on `main`**: `python/src/bindings/vitals.rs`, `python/tests/test_vitals.py`, and `python/bench/test_bench_vitals.py` are byte-identical between this PR's head and current `main` (which already has the `phases` binding and both new tests). The `weights→phases` diff shown here is against a stale merge-base, not current `main`.

So this looks safe to **close as already-applied** — merging would be a no-op. (Worth double-checking on your side, but that's what current `main` shows.)

mvanhorn:
You're right, and thanks for checking that carefully. I confirmed against current `main`: `python/src/bindings/vitals.rs`, `python/tests/test_vitals.py` and `python/bench/test_bench_vitals.py` are byte-identical to this branch, and `HeartRateExtractor.extract()` on `main` already takes `phases` with the docstring noting that `phases=[]` is invalid unlike `BreathingExtractor`'s weights. The diff here was only against a stale merge-base. Closing as already-resolved.

Issue #1289, opened 2026-07-09
Compact ESP32-S3 clones (Zero/SuperMini) may overheat and fail after a normal CSI session


matiapa:
## Summary

Three ESP32-S3-Zero boards (compact/coin-sized clones, 4MB flash) ran a normal
CSI streaming session (WiFi connected, `edge_tier=2`, ~20-30 min, node_ids
1-3) without any errors. They ran warm/hot to the touch throughout, which
wasn't flagged as unusual at the time. After being unplugged and later
reconnected (to a different, known-good USB power source), none of the three
boards powered on again, and all three heated up immediately when powered —
consistent with permanent regulator/component damage rather than a transient
fault.

## Suspected root cause

This firmware runs the WiFi radio with modem sleep disabled
(`WIFI_PS_NONE`, needed for continuous CSI capture) plus a full edge
DSP pipeline on Core 1 (`edge_tier=2`) — sustained high current draw with
no duty-cycling. Full-size dev boards (DevKitC-1, XIAO) have enough
copper pour / thermal mass around the regulator to handle this
indefinitely. Coin-sized clones (ESP32-S3-Zero, SuperMini) have minimal
PCB copper and often budget voltage regulators, and may not be rated for
sustained full-load operation without a heatsink or airflow. All three
boards in this report were the same model, run side by side in the same
session, which is consistent with a shared thermal/duty-cycle cause
rather than three independent unit failures.

## What we already documented (#1288)

Added an explicit thermal-risk warning to:
- `firmware/esp32-csi-node/README.md` (Hardware Requirements section)
- `CLAUDE.md` (Supported Hardware table)
- `ruview-hardware-setup` skill (so an AI agent flashing on someone's
  behalf asks about board form factor and warns proactively)

## Possible follow-up (not done in #1288)

- A Kconfig / NVS-configurable duty-cycled mode (periodic
  `esp_wifi_stop()`/light-sleep windows) for compact boards, trading
  some CSI throughput for sustainably lower average power.
- A boot-time runtime check or periodic internal temperature read (if
  the SoC's internal temp sensor is exposed) that logs a warning if the
  chip die temperature climbs unusually high, to catch this before
  hardware damage rather than after.
- Explicit non-recommendation of ESP32-S3-Zero/SuperMini for `edge_tier=2`
  continuous deployments, only for short/tier-0-1 test sessions, pending
  real thermal characterization.

## Environment

[...]

PR #1291, opened 2026-07-09
Fix: Compact ESP32-S3 clones (Zero/SuperMini) may overheat and fail after a normal CSI session

anujsingh-cse:
Hey! I've put together a fix for #1289. 

Let me know if this looks good to you or if you need any adjustments!

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -93,6 +93,8 @@ All 5 ruvector crates integrated in workspace:
 
 **Not supported:** ESP32 (original), ESP32-C3 — single-core, can't run CSI DSP pipeline.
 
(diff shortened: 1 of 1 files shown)


PrinceAudre:
This adds a one-line thermal warning to `CLAUDE.md` for #1289, but #1288 already covers the same issue more completely — it fixes the firmware (WPA/WPA2-mixed authmode) **and** adds more detailed thermal-risk docs in three places (README / CLAUDE.md / SKILL.md). Both target #1289.

Recommend **closing in favor of #1288**. (Minor: this diff also drops the trailing newline at EOF of `CLAUDE.md`.)

PR #1296, opened 2026-07-09
fix(sensing-server): parse ADR-081 feature_state (0xC511_0006), the node's default payload

lakinduakash:
## Problem

`udp_receiver_task` in `wifi-densepose-sensing-server` decodes only:
- `0xC511_0001` raw CSI
- `0xC511_0002` edge vitals
- `0xC511_0004` fused vitals
- `0xC511_0007` WASM output

But firmware `rv_feature_state.h` documents `feature_state` (`0xC511_0006`) as:

> "The **default upstream payload** from a node ... **replaces raw ADR-018 CSI as the primary stream**."

So a node emitting its *default* 60-byte compact state (presence / motion / respiration / heartbeat / anomaly / coherence) produces **zero UI updates** — `/health` stuck at `tick:0`, the UI banner reads OFFLINE — even though frames are arriving on UDP `:5005`.

**MEASURED** on an ESP32-S3-DevKitC-1: ~5 pps of `0xC511_0006` received and silently dropped; after this change the same stream advances `tick` and drives the sensing UI.

## Fix

- Add `parse_esp32_feature_state()`: validates the magic + 60-byte length and maps the packet into the existing `Esp32VitalsPacket`.
  - Presence fails **closed**: requires the node's `RV_QFLAG_PRESENCE_VALID` bit **and** `presence_score >= 0.5`.
  - `ts_us → timestamp_ms`; `respiration_bpm` / `heartbeat_bpm` passthrough; `motion_score` drives the motion flag.
- Wire it as an `.or_else()` fallback after `parse_esp32_vitals`, so it flows through the **unchanged** `SensingUpdate` / WebSocket path (no downstream churn).

## Tests

New `feature_state_parser_tests` (4):
- field extraction + presence mapping when valid-flag set
- fail-closed when `PRESENCE_VALID` is clear (high score, still not present)
- wrong-magic rejection
- short-buffer rejection

`wifi-densepose-sensing-server` bin tests **171 → 175, 0 failed**.

## Field note (not in this diff)

On a display-less DevKitC the shipped v0.7.0 binary mis-detects a phantom panel (`disp_hal: TCA9554 not found — assuming display power is always-on`) → `has_display=true` → CSI promiscuous filter stays MGMT-only → 0 pps → `DEGRADED`. A display-disabled build restores ~30 pps yield. `main` already carries the RuView#1000 touch-probe bail, so this looks like a **stale-release / build-variant** issue rather than a code regression — flagging it so the headless release binaries can be rebuilt/re-released. No firmware code changed here.


PrinceAudre:
Nice fix — the wire offsets match the firmware `rv_feature_state.h` layout exactly, and the fail-closed clamp + `presence_valid && score >= 0.5` gate are sound.

One small imprecision: `s.last_feature_state_at = Some(Instant::now())` is set in the branch that handles **both** the `0x02` edge-vitals packet and the `0x06` feature_state packet, but the field is documented as tracking "when the last feature_state (0xC511_0006) presence was received." As a result `prefer_board_presence()` also engages for edge-vitals-only nodes — broader than the stated intent.

Either gate the `last_feature_state_at` update to the `0x06` path only, or rename it to something like `last_board_presence_at` and update the doc comment to match. Not a blocker.

PR #1305, opened 2026-07-10
fix(security): auth + CORS hygiene for nvsim-server, calibrate-serve; repo cleanup

karimalsalah:
## Summary

Five focused fixes for the `lane:ruview` audit:

1. **[P1] nvsim-server zero-auth + wide-open CORS.** The ghcr.io-published Docker image (auto-built on every `main` push) shipped an unauthenticated API bound to `0.0.0.0:7878` with `Access-Control-Allow-Origin: *`. Adds an optional `--token`/`NVSIM_TOKEN` bearer gate on all state-mutating routes (`PUT scene/config/seed`, `POST run/pause/reset/step/witness*/export-proof`) while leaving `GET` routes — including `/api/health` (used by the Docker `HEALTHCHECK`) — and `/ws/stream` open. Narrows the default CORS allowlist from `*` to the dashboard's real origins (`localhost:5173`, `127.0.0.1:5173`, `ruvnet.github.io`); `*` stays available as an explicit opt-in. Logs a loud warning at startup if bound non-loopback with no token set. Mirrors the `homecore-api` (HC-05) CORS-allowlist pattern and `calibrate_api.rs`'s bearer-gate pattern already audited elsewhere in this repo.
2. **[P2] calibrate-serve unconditional `CorsLayer::permissive()`.** Now gated: `permissive()` only on the default local (`127.0.0.1`/`localhost`) bind; a non-default `--http-bind` gets an explicit `GET`/`POST`-only, non-credentialed CORS policy instead.
3. **[P3] Deletes `references/`** — 4.3MB of unrelated "InvisPose" content with zero inbound references (confirmed via repo-wide grep before deletion).
4. **[P3] Removes a duplicate `with: submodules: recursive` key** in the `rust-tests` checkout step of `ci.yml` (the second occurrence silently shadowed the first).
5. **[P2] Documents the already-shipped HEAD commit's Linux passive WiFi-RSSI sensing source** (`--source wifi` via `iw dev <iface> link`, no root required — distinct from the pre-existing root-requiring `--source linux` active-scan mode) and the Observatory RF-constellation UI layer, in `CHANGELOG.md` and `README.md`, per this repo's own pre-merge checklist.

## Test plan

- [x] `cargo check -p nvsim-server` — clean
- [x] `cargo check -p wifi-densepose-cli --no-default-features` — clean
- [x] `cargo test -p nvsim-server -p wifi-densepose-cli --no-default-features` — **37/37 pass** (10 nvsim-server + 27 wifi-densepose-cli)
- [x] Live end-to-end verification of nvsim-server: `GET` routes open without auth; `PUT`/`POST` return `401` with no/wrong bearer and `200` with the correct bearer; CORS preflight from an allowlisted origin returns `Access-Control-Allow-Origin`, a disallowed origin gets none; the non-loopback-no-token startup warning fires as designed
- [x] `references/` confirmed deleted, zero remaining inbound references
- [x] `ci.yml` re-parses as valid YAML with exactly one `with:` mapping in the checkout step

Implemented by Codex (frontier reasoning per repo directive), cross-audited and independently re-verified by Claude. One out-of-scope change Codex made (removing unused `assert_cmd`/`predicates` dev-dependencies from `wifi-densepose-cli`) was reverted to keep this PR focused on the requested fixes — confirmed via grep those deps had zero usages either way, so nothing was lost.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (implementation via Codex)


karimalsalah:
Opened against the wrong repo (should target karimalsalah/RuView, the fork this branch is based on). Closing and re-opening correctly.

Issue #1308, opened 2026-07-11
firmware: display probe false-positives on display-less DevKitC-1, silently disabling the #893 CSI-yield fix (0 pps)


erichkusuki:
## Summary

On a display-less **ESP32-S3-DevKitC-1 (N16R8)**, the ADR-045 display probe **false-positives**, so the #893 MGMT+DATA promiscuous upgrade is skipped and CSI yield collapses to 0 pps — the exact symptom #893 was meant to fix. Since the probe result gates the fix, and both shipped S3 binaries are affected (see below), **no prebuilt S3 image currently works on stock DevKitC-1 boards**.

## Environment

- 2× ESP32-S3-DevKitC-1-N16R8 (no display), flashed with `release_bins/` v0.7.0 (`git-sha 8703ade9b`)
- Sensing server v0.7.x in Docker (`ruvnet/wifi-densepose:latest` and a local source build)
- 2.4 GHz WPA2 network (UniFi), nodes and aggregator on the same subnet

## Symptom

Boot log (full boot captured over UART):

```
I disp_task: Initializing display subsystem...
W disp_hal: TCA9554 not found — assuming display power is always-on (direct wiring)
I disp_hal: QSPI panel IO created (40MHz, quad mode)
I disp_hal: SH8601 panel init OK (368x448)          <-- no panel exists on this board
I disp_task: Display task started (Core 0, priority 1, 30 fps)
...
I adaptive_ctrl: medium tick: state=8 yield=0pps ...  <-- forever
```

The SH8601 init sequence "succeeds" against floating pins, `display_is_active()` returns true, and `main.c` skips `csi_collector_enable_data_capture()`:

- `firmware/esp32-csi-node/main/main.c` — `has_display = display_is_active()` gates the #893 upgrade
- `firmware/esp32-csi-node/main/csi_collector.c` — `csi_collector_enable_data_capture()` (the fix that never runs)

Downstream effect: nodes emit almost no CSI/vitals; the sensing server flaps `esp32` ↔ `esp32:offline` (5 s liveness window), the dashboard alternates ESP32/SIMULATED, and multistatic fusion fails every governed cycle with multi-second timestamp spreads.

## Why the shipped binaries can't work on DevKitC-1

1. `release_bins/esp32-csi-node.bin` (= `s3-adr110/`, v0.7.0): has the #893 code but `CONFIG_DISPLAY_ENABLE=y` + the false-positive probe disables it at runtime.
[...]

PR #1311, opened 2026-07-11
fix(firmware): display-less DevKitC-1 build overlay + stale IDF v5.2 doc refs (#1308)

erichkusuki:
Addresses #1308

## What

- Adds `firmware/esp32-csi-node/sdkconfig.defaults.devkitc` — a build overlay for display-less boards (ESP32-S3-DevKitC-1 and similar) that sets `CONFIG_DISPLAY_ENABLE=n`, making `has_display` constant-false so the RuView#893 MGMT+DATA CSI upgrade always applies.
- Updates the stale `espressif/idf:v5.2` references in the firmware README (Quick Start badge + build commands) to `v5.4`: current source uses `esp_driver_uart` (IDF v5.3+) and fails CMake resolution under v5.2; CI already builds with v5.4. Later sections of the README already said v5.4 — this aligns the rest.
- Adds a Quick Start note pointing display-less-board users to the overlay.

## Why

On a stock DevKitC-1 the ADR-045 display probe false-positives (`SH8601 panel init OK` against floating pins), so `display_is_active()` returns true and `main.c` skips `csi_collector_enable_data_capture()` — CSI yield collapses to 0 pps, the exact symptom #893 fixed. See #1308 for full boot logs and analysis of why neither shipped S3 prebuilt works on these boards.

This PR is the minimal user-side unblock. It deliberately does **not** attempt the deeper fixes suggested in #1308 (hardening the probe with a panel-ID readback, or an NVS `display=0` override) — happy to follow up on either if maintainers have a preference.

## Verification

Hardware-verified on 2× ESP32-S3-DevKitC-1-N16R8: built with the overlay under `espressif/idf:v5.4` (Docker), flashed, and confirmed the boot log now shows `CSI filter upgraded to MGMT+DATA (no display, RuView#893)` with steady 40–45 pps yield (was 0 pps), presence/motion detection active, and the sensing server holding a stable `esp32` source with zero fusion errors.

### DIFF
--- a/firmware/esp32-csi-node/README.md
+++ b/firmware/esp32-csi-node/README.md
@@ -4,7 +4,7 @@
 
 This firmware captures WiFi Channel State Information (CSI) from an ESP32-S3 (production) or ESP32-C6 (research target — Wi-Fi 6 / 802.15.4 / TWT / LP-core hibernation, see [ADR-110](../../docs/adr/ADR-110-esp32-c6-firmware-extension.md)) and transforms it into real-time presence detection, vital sign monitoring, and programmable sensing -- all without cameras or wearables. Part of the [WiFi-DensePose](../../README.md) project.
 
-[![ESP-IDF v5.2](https://img.shields.io/badge/ESP--IDF-v5.2-blue.svg)](https://docs.espressif.com/projects/esp-idf/en/v5.2/)
+[![ESP-IDF v5.4](https://img.shields.io/badge/ESP--IDF-v5.4-blue.svg)](https://docs.espressif.com/projects/esp-idf/en/v5.4/)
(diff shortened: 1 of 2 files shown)


ruvnet:
Merged via #1332 — hardware verification on 2× DevKitC-1-N16R8 appreciated. One fix applied on the integration branch: the overlay header's copy-paste build command still referenced `espressif/idf:v5.2`, the version this PR itself removes from the README (fails CMake under v5.2 due to `esp_driver_uart`); it now says v5.4. The deeper #1308 follow-ups you deferred (panel-ID readback probe hardening / NVS `display=0` override) remain open ideas on #1308's thread — a PR for either is welcome.

ztezxy-hub:
[policies.txt](https://github.com/user-attachments/files/30355609/policies.txt)

Issue #1049, opened 2026-06-13
Make multistatic guard_interval_us configurable (hardcoded 5000 µs causes permanent trust demotion with 2 ESP32-S3 nodes)


Mephisto090:
## Problem

Running 2× ESP32-S3 nodes (firmware v0.7.0) with the Docker image `ruvnet/wifi-densepose:latest`, the fusion engine permanently demotes trust to "Restricted" and suppresses all pose outputs because the timestamp spread between nodes consistently exceeds the hardcoded 5000 µs guard interval.

Typical log output (repeats every 10s indefinitely):

```
WARN sensing_server::engine_bridge: governed trust cycle failed: 
fusion error: Timestamp spread 73600 us exceeds guard interval 5000 us 
total_engine_errors=204712
```

The error counter reaches 200k+ within hours. Once trust is demoted (`raw_outputs_suppressed: true`), the `/api/v1/pose/current` endpoint returns 0 poses and the UI falls back to demo mode. There is no API endpoint or CLI flag to reset trust or adjust the guard interval without restarting the container.

## Environment

- Server: `ruvnet/wifi-densepose:latest` (Docker, Rust sensing-server)
- Nodes: 2× ESP32-S3 (QFN56, rev v0.2), firmware v0.7.0, 16MB flash
- ESP-NOW sync is active (`c6_sync_espnow_init` succeeds on both nodes)
- Measured timestamp spread: 10–150 ms (typical ~70 ms)
- Network: single 2.4 GHz AP, nodes on same WiFi channel, UDP port 5005
- `MultistaticConfig` uses `..Default::default()` → `guard_interval_us: 5000`

## Root Cause

The guard interval is hardcoded in `wifi-densepose-signal/src/ruvsense/multistatic.rs`:

```rust
impl Default for MultistaticConfig {
    fn default() -> Self {
        Self {
            guard_interval_us: 5000,
            soft_guard_us: 1000,
            // ...
        }
    }
}
```

The sensing-server instantiates the fuser with only `min_nodes: 1` overridden:

```rust
MultistaticFuser::with_config(MultistaticConfig {
    min_nodes: 1,
    ..Default::default()
})
```

No CLI argument, environment variable, or runtime API exists to change `guard_interval_us`.

The ESP-NOW sync (100ms beacons, ~540µs WiFi-MAC jitter per ADR-110) cannot keep two independently-clocked ESP32-S3 boards within 5ms — the measured drift is 10–15× larger than the guard allows.

[...]

Issue #1031, opened 2026-06-12
Multistatic fusion failed (Timestamp spread 18194 us exceeds guard interval 5000 us), using per-node sum/dedup fallback


tigermask1978:
Two nodes:
Node 1
python firmware/esp32-csi-node/provision.py `
  --port COM6 `
  --ssid "HC-14-02K" `
  --password "密码" `
  --target-ip 192.168.5.128 `
  --node-id 1 `
  --tdm-slot 0 `
  --tdm-total 2

Node 2
python firmware/esp32-csi-node/provision.py `
  --port COM7 `
  --ssid "HC-14-02K" `
  --password "密码" `
  --target-ip 192.168.5.128 `
  --node-id 2 `
  --tdm-slot 1 `
  --tdm-total 2

$env:RUST_LOG="debug"
$env:CSI_SOURCE="esp32"
cargo run -p wifi-densepose-sensing-server --bin sensing-server --no-default-features -- --source esp32

The sensing-server reports as the title, Why?

PR #1312, opened 2026-07-11
fix(sensing-server): wire WDP_GUARD_INTERVAL_US/WDP_TDM_* into EngineBridge (#1309)

### LINKED ISSUE #1309: sensing-server: EngineBridge ignores WDP_GUARD_INTERVAL_US/WDP_TDM_* — governed trust cycles always use the hardcoded 60 ms guard
## Summary

`WDP_GUARD_INTERVAL_US` / `WDP_TDM_SLOTS` + `WDP_TDM_SLOT_US` (the #1031/#1049 multistatic guard overrides) only reach the `multistatic_fuser` field on `AppState`. The **`EngineBridge` builds its own `StreamingEngine`, which constructs a second `MultistaticFuser` with `MultistaticConfig::default()`** — a hardcoded 60 ms guard. So the governed trust path ignores the deployment's configured guard entirely.

## Symptom

Startup log claims the override took effect:

```
INFO sensing_server: Multistatic fusion guard: 200000 µs hard / 33333 µs soft (override via WDP_GUARD_INTERVAL_US ...)
```

…but every governed cycle still fails against the default:

```
WARN sensing_server::engine_bridge: governed trust cycle failed ...: fusion error:
     Timestamp spread 233055 us exceeds guard interval 60000 us
```

Note the `60000 us` — not the configured `200000`. With WiFi/ESP-NOW-synced ESP32 nodes (which per the #1049 comment drift 10–150 ms), the governed trust path can therefore never succeed on a real multi-node deployment, regardless of configuration.

## Root cause

- `v2/crates/wifi-densepose-sensing-server/src/main.rs`: the `multistatic_fuser` initializer calls `multistatic_guard_config_from_env()` — but the adjacent `engine_bridge: EngineBridge::new(...)` takes no config.
- `v2/crates/wifi-densepose-sensing-server/src/engine_bridge.rs`: `EngineBridge::new()` → `StreamingEngine::new()`.
- `v2/crates/wifi-densepose-engine/src/lib.rs`: `StreamingEngine::new()` → `MultistaticFuser::with_config(MultistaticConfig::default())`, no setter to change it.

## Fix (verified locally, happy to send a PR)

1. `wifi-densepose-engine`: add `StreamingEngine::set_multistatic_config(cfg)` (rebuilds the internal fuser).
2. `engine_bridge.rs`: `EngineBridge::new(..., multistatic_cfg: Option<MultistaticConfig>)`, applied to the engine when `Some`.
3. `main.rs`: thread the same env-derived config used for `multistatic_fuser` into `EngineBridge::new`.

After the patch, the failure message correctly reflects the configured guard, and with healthy nodes (steady CSI at ~42 pps each) governed cycles run with **zero** fusion errors over a 90 s observation window.

erichkusuki:
Fixes #1309

## What

Threads the #1031/#1049 multistatic guard configuration (`WDP_GUARD_INTERVAL_US`, `WDP_SOFT_GUARD_US`, `WDP_TDM_SLOTS` + `WDP_TDM_SLOT_US`) into the governed trust path:

- `wifi-densepose-engine`: new `StreamingEngine::set_multistatic_config()` (rebuilds the internal fuser; previously hardwired to `MultistaticConfig::default()`).
- `engine_bridge.rs`: `EngineBridge::new()` takes an `Option<MultistaticConfig>` and applies it to the engine (`None` keeps today's default — all existing tests updated accordingly).
- `main.rs`: the same env-derived config already used for `AppState.multistatic_fuser` is now also passed to `EngineBridge::new` (with the same `min_nodes: 1` single-node passthrough).

## Why

The startup log claimed the override took effect ("Multistatic fusion guard: 200000 µs hard …") while every governed cycle kept failing against the hardcoded default ("Timestamp spread … exceeds guard interval **60000 us**"). With WiFi/ESP-NOW-synced ESP32 nodes drifting 10–150 ms (per the #1049 comment in `main.rs`), the governed trust path could never succeed on a real multi-node deployment, regardless of configuration.

## Verification

Deployed on a live 2-node ESP32-S3 setup: after the patch the failure message reflects the configured guard, and with both nodes streaming (~42 pps each) governed cycles ran a 90 s observation window with **zero** fusion errors (`engine_error_count: 0` on `/api/v1/status`); previously every cycle failed. The workspace builds clean via `docker/Dockerfile.rust`.

### DIFF
--- a/v2/crates/wifi-densepose-engine/src/lib.rs
+++ b/v2/crates/wifi-densepose-engine/src/lib.rs
@@ -230,6 +230,17 @@ impl StreamingEngine {
     /// cycle on, the adapter id is part of provenance `model_version` — and
     /// therefore of the witness — so the exact weights shaping inference are
     /// pinned in the trust chain. Pass the result of hashing the adapter file.
+    /// Override the multistatic fuser's timestamp guard interval (#1049/#1057).
+    /// Without this, `StreamingEngine::new` always builds
+    /// `MultistaticFuser::with_config(MultistaticConfig::default())` — a
+    /// hardcoded 60 ms hard guard that ignores whatever schedule/override the
+    /// caller derived from `WDP_TDM_SLOTS`/`WDP_GUARD_INTERVAL_US`, so
+    /// WiFi/ESP-NOW-synced multi-node deployments spuriously fail governed
+    /// trust cycles even after widening the guard elsewhere.
+    pub fn set_multistatic_config(&mut self, cfg: MultistaticConfig) {
+        self.fuser = MultistaticFuser::with_config(cfg);
+    }
+
     pub fn set_room_adapter(&mut self, info: AdapterInfo) {
         self.adapter = Some(info);
     }
--- a/v2/crates/wifi-densepose-sensing-server/src/engine_bridge.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/engine_bridge.rs
@@ -38,6 +38,7 @@ use wifi_densepose_bfld::{PrivacyClass, PrivacyMode};
 use wifi_densepose_engine::{AdapterInfo, EngineError, StreamingEngine, TrustedOutput};
 use wifi_densepose_geo::types::GeoRegistration;
 use wifi_densepose_signal::ruvsense::fusion_quality::CalibrationId;
+use wifi_densepose_signal::ruvsense::multistatic::MultistaticConfig;
 use wifi_densepose_worldgraph::WorldId;
 
 use super::multistatic_bridge::node_frames_from_states;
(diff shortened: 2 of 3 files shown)


ruvnet:
Merged via #1332 (all 1,031+ workspace tests green, hardware-verified per your report — thank you, this was the most consequential of the three: the governed trust path could never succeed on real multi-node deployments without it). One small fix applied on the integration branch: the new `set_multistatic_config` doc block had absorbed `set_room_adapter`'s ADR-150 adapter/witness doc comment; they're separated now. `wifi-densepose-engine` 0.3.1 / `wifi-densepose-sensing-server` 0.3.4 are being published to crates.io with the change.

PR #1313, opened 2026-07-11
fix(sensing-server): exempt /api/v1/stream/pose WS from bearer auth + UI token field (#1310)

### LINKED ISSUE #1310: sensing-server: /api/v1/stream/pose WebSocket is unreachable when RUVIEW_API_TOKEN is set (browsers cannot send Authorization on WS upgrade)
## Summary

When bearer auth is enabled (`RUVIEW_API_TOKEN` set, #443), the Live Demo tab can never connect: its pose stream lives at **`/api/v1/stream/pose`**, which falls under the protected `/api/v1/*` prefix — but the browser `WebSocket` constructor **cannot attach an `Authorization` header** to the upgrade request. The UI shows:

```
Failed to start: WebSocket connection failed: Unknown error
```

`/ws/sensing` was already deliberately exempted for exactly this reason (see the module docs in `bearer_auth.rs`: "`/ws/sensing` and `/ui/*` are served to local browsers that can't easily inject headers"). The same reasoning applies to `/api/v1/stream/pose`, which was presumably missed because it sits inside the protected prefix.

## Repro

1. `docker run -e RUVIEW_API_TOKEN=<token> ... ruvnet/wifi-densepose:latest`
2. Open the UI, go to Live Demo → connection fails with the error above; server returns `401` on the WS upgrade.

## Fix (verified locally, happy to send a PR)

`v2/crates/wifi-densepose-sensing-server/src/bearer_auth.rs`: add a narrow exempt-path list checked alongside the prefix:

```rust
const EXEMPT_PATHS: &[&str] = &["/api/v1/stream/pose"];
...
if !path.starts_with(PROTECTED_PREFIX) || EXEMPT_PATHS.contains(&path) {
    return next.run(request).await;
}
```

plus a regression test asserting the pose stream stays reachable without a token while other `/api/v1/*` paths remain 401. (Deliberately **not** a query-string token — the existing CWE-598 regression test correctly forbids that channel.)

## Related UI gap (mention or split as preferred)

Even for plain REST, the bundled UI has no way to supply the token: `ui/services/api.service.js` has `setAuthToken()`, but nothing ever calls it and there is no settings field for it — so enabling `RUVIEW_API_TOKEN` currently breaks the entire dashboard (constant `HTTP 401` on `/api/v1/*`). We added a "API Access" token field to the QuickSettings panel (stores in `localStorage`, applies before first request) and can include that in the PR too.

erichkusuki:
Fixes #1310

## What

- `bearer_auth.rs`: adds a narrow `EXEMPT_PATHS` list (`/api/v1/stream/pose`) checked alongside `PROTECTED_PREFIX`, plus a regression test asserting the pose-stream WS stays reachable without a token **and** that the exemption does not leak to other `/api/v1/*` paths. The existing CWE-598 test (query-string tokens rejected) is untouched.
- `ui/utils/quick-settings.js` + `ui/style.css`: adds an "API Access" section to the QuickSettings panel — a password-type field that stores the bearer token in `localStorage` and applies it via `apiService.setAuthToken()` before the first request (with save/clear + status line).

## Why

Browsers cannot attach an `Authorization` header to a WebSocket upgrade, so with `RUVIEW_API_TOKEN` set the Live Demo pose stream always failed (`Failed to start: WebSocket connection failed`). `/ws/sensing` is already exempted for exactly this reason (see the bearer_auth module docs); `/api/v1/stream/pose` was presumably missed because it sits inside the protected prefix.

Separately, the bundled dashboard had no way to supply the token at all: `api.service.js` ships `setAuthToken()` but nothing ever called it, so enabling bearer auth broke every `/api/v1/*` call from the UI (constant 401s). The two changes together make `RUVIEW_API_TOKEN` actually usable with the bundled dashboard. Can split the UI part into its own PR if preferred.

## Verification

Deployed against a live 2-node ESP32-S3 setup with `RUVIEW_API_TOKEN` set: token entered via the new Settings field → all REST 401s gone; Live Demo tab connects (`LIVE — ESP32 HARDWARE CONNECTED`, frames processed, 0 errors). New unit test passes alongside the existing bearer_auth suite; image builds clean via `docker/Dockerfile.rust`.

### DIFF
--- a/ui/style.css
+++ b/ui/style.css
@@ -4088,6 +4088,20 @@ a:focus-visible,
   color: #fff;
 }
 
+.qs-text-input {
+  padding: var(--space-6) var(--space-8);
+  border-radius: var(--radius-sm);
+  border: 1px solid var(--color-border);
+  background: var(--color-secondary);
+  color: var(--color-text);
+  font-size: var(--font-size-sm);
+}
+
+.qs-text-input:focus {
+  outline: 2px solid var(--color-primary);
+  outline-offset: 1px;
+}
+
 /* --- Screenshot Flash --- */
 
 .screenshot-flash {
--- a/ui/utils/quick-settings.js
+++ b/ui/utils/quick-settings.js
@@ -1,6 +1,10 @@
 // Quick Settings Panel - Centralized configuration for all UI features
 // Accessible via gear icon in header
 
+import { apiService } from '../services/api.service.js';
+
+const API_TOKEN_STORAGE_KEY = 'ruview-api-token';
+
 export class QuickSettings {
   constructor(app) {
     this.app = app;
@@ -10,8 +14,19 @@ export class QuickSettings {
   }
 
   init() {
+    this.applyStoredApiToken();
     this.createButton();
     this.createPanel();
+  }
+
+  // Apply a previously-saved bearer token to apiService as early as
+  // possible, before any tab's REST calls fire. The server only ever
+  // checks the `Authorization: Bearer` header (see bearer_auth.rs) — this
+  // intentionally never puts the token in a URL query string.
+  applyStoredApiToken() {
+    let token = null;
+    try { token = localStorage.getItem(API_TOKEN_STORAGE_KEY); } catch { /* noop */ }
+    if (token) apiService.setAuthToken(token);
   }
 
   createButton() {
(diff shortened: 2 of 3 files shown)


ruvnet:
Merged via #1332 — thanks, this makes `RUVIEW_API_TOKEN` actually usable with the bundled dashboard. One behavioral fix applied on the integration branch: the stored token is now applied at `api.service.js` module load rather than in `QuickSettings.init()` — `initializeServices()` and the dashboard tab fire their first `/api/v1/*` requests before `initializeEnhancements()` runs, so the first requests on every page load still went out without the Authorization header (transient 401s). The QuickSettings panel keeps the save/clear UI. The pose-WS exemption was kept as-is, matching the `/ws/sensing` posture; a ticket-based WS auth flow is noted as a candidate follow-up.

PR #1343, opened 2026-07-17
fix(sensing-server): unstick empty-room field-model calibration deadlock

rjperry36:
## Problem

`POST /api/v1/calibration/*` is a dead endpoint on a live streaming node — the field-model calibration can never start. Observed on a real ESP32-S3 deployment as `{"status":"Uncalibrated","frame_count":0}` that never advances, no matter how long the room is left empty.

### Root cause — a two-gate deadlock

- `calibration_start` creates the `FieldModel` in `Uncalibrated`.
- The per-frame server feed `field_bridge::maybe_feed_calibration` only fed observations **while the model was already `Collecting`**.
- But the **only** thing that sets `Collecting` is `feed_calibration` itself, on its first fed frame.

So nothing ever fed the first frame → status never left `Uncalibrated` → `calibration_frame_count` stayed pinned at 0 → the SVD room eigenstructure (eigenvalue-based person counting / localization) could never calibrate.

Presence/motion/vitals are unaffected — they use the separate automatic rolling baseline, not the field model.

## Fix

- **`field_bridge::maybe_feed_calibration`** — feed while `Uncalibrated | Collecting`, so the first frame flips the model to `Collecting` and the count advances.
- **`calibration_stop`** — return a structured `{success:false, frame_count, frames_needed}` instead of an opaque 500 when finalized before enough empty-room frames accumulate.
- **`FieldModel::min_calibration_frames()`** — new accessor used by the guard above.

## Tests

- New regression test `field_bridge::tests::maybe_feed_calibration_advances_uncalibrated_to_collecting` — asserts `Uncalibrated → Collecting` and frame count `0 → 1 → 2` (fails on old code).
- Verified locally (`--no-default-features`):
  - `wifi-densepose-sensing-server` bin: **172 passed, 0 failed**
  - `wifi-densepose-signal` lib: **477 passed, 0 failed**


PrinceAudre:
The core deadlock fix is correct — feeding while `Uncalibrated | Collecting` so the first frame flips the model to `Collecting`, plus the HT40→canonical-56 resample (since `LinkStats::update` rejects non-56 widths) and the structured `calibration_stop`. And it's safely gated: `field_model` is `None` unless `--calibrate` / an explicit `calibration_start`, so the broadened feed gate can't auto-collect without operator intent.

Two things bundled in here make it hard to merge as one PR, though:

1. **An unrelated, under-tested classifier change.** `BASELINE_SUBTRACT_FACTOR` 0.7 → 1.0 (`types.rs`) changes motion detection everywhere (`(raw_motion - baseline*factor).max(0.0)`). Full baseline subtraction can over-suppress *real* motion → false negatives. The only added test proves empty-room → `absent`; it doesn't prove a present person still registers (the PR itself defers that to live hardware). This deserves its own PR with a person-present regression test asserting motion-above-baseline still classifies `present_moving`.

2. **Fork-specific CI/deploy artifacts.** A GitHub Actions workflow that builds and pushes to a personal container-registry namespace (triggered on the fork's own branch), plus fork deployment templates and a deploy script. These aren't upstream changes and will run/behave oddly on `main`.

Suggestion: split into (a) the deadlock + resample + `calibration_stop` core (mergeable as-is), (b) the `BASELINE_SUBTRACT_FACTOR` change with a sensitivity test, and (c) drop the fork CI/deploy files from the upstream PR entirely.

PR #1362, opened 2026-07-19
fix(ci): restore workflow and LAN smoke tests

ruvnet:
## Root cause

Two repository-level CI defects remained after the ADR-270 release:

1. `.github/workflows/ci.yml` was rejected before job creation because the Rust checkout step declared `with:` twice. Its Codecov v6 step also used the obsolete singular `file` input.
2. The sensing-server image built and pushed, but its LAN smoke test launched the hardened entrypoint on `0.0.0.0` without a token or the explicit trusted-LAN opt-in, so it exited 64 before HTTP checks.

These failures predate ADR-270 and reproduce on the prior Qualcomm/MediaTek commits.

## Fix

- Keep one recursive-submodule checkout mapping.
- Use Codecov v6 `files`.
- Set `RUVIEW_ALLOW_UNAUTHENTICATED=1` only in the explicit LAN-mode smoke container. Production defaults remain fail-closed; the bearer-token smoke test remains separate.

## Validation

- `actionlint` passes for both modified workflows (ignoring only unrelated shellcheck style diagnostics).
- Replayed the previously failing published image locally with the exact new LAN environment: `/health` and `/api/v1/info` both returned HTTP 200.
- `git diff --check` passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,8 +88,6 @@ jobs:
       # ADR-262 P1: `wifi-densepose-rufield` path-deps the `vendor/rufield`
       # submodule. Without a recursive checkout the workspace build fails to
       # resolve those path deps in CI even though it passes locally.
-      with:
-        submodules: recursive
 
     # `wifi-densepose-desktop` is a Tauri v2 app — `glib-sys`, `gtk-sys`,
     # `webkit2gtk-sys`, etc. need the Linux dev libraries via pkg-config or the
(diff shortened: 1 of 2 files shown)


ruvnet:
Follow-up from the restored CI run: the ADR-147 step targeted a package that was intentionally extracted into the nested `crates/worldgraph` workspace, so `cargo test -p` from the parent workspace failed with “cannot specify features for packages outside of workspace.” The workflow now uses the nested package manifest explicitly. Local validation passed: 12 unit tests + 1 doctest.

PR #1382, opened 2026-07-21
feat(sensing-server): opt-in OpenTelemetry log export with a weaver-checked ruview.* registry

jensholdgaard:
## What

Opt-in OpenTelemetry **logs** export for the sensing server, with a curated set of sensing events under a `ruview.*` namespace whose names and attribute keys are defined in a semantic-conventions registry and enforced in CI:

- **`semconv/registry/`** — a weaver-validated (0.23, `registry check --future`) OTel semantic-conventions registry: 13 attributes and 8 log event names, all `stability: development`, with requirement levels on the event attribute refs and enums mirroring the code's actual value sets (the adaptive classifier's class list, the sensing source labels).
- **`src/semconv.rs`** — Rust constants **generated** from the registry (`weaver registry generate`, template under `templates/registry/rust/`). The new `semconv` workflow pins weaver v0.23.0 by sha256, validates the registry, regenerates the module, and fails on drift (codegen no-diff).
- **`src/telemetry.rs`** — tracing bootstrap with optional OTLP export, **doubly gated**:
  1. build: a new `otel` cargo feature (same gating principle as `mqtt`) keeps the opentelemetry 0.32 stack out of the default binary;
  2. runtime: even with the feature, export only activates when `OTEL_EXPORTER_OTLP_ENDPOINT` is set.

  Unset ⇒ the OTLP pipeline is never constructed and behavior is byte-identical to today's stderr `fmt` subscriber. When enabled, every tracing event exports as an OTel log record (`service.name = "ruview"`) with a telemetry-induced-telemetry loop guard muting the exporter's own crates.
- **`docker/otel-compose.yml` + `docs/observability.md`** — a one-command demo pipeline (sensing server on synthetic CSI → OTel Collector → an OTLP log backend) and the guide.

## Why

The sensing server already narrates the interesting moments (presence flips, vitals, node churn, falls, MQTT failures) as free-text log lines. Giving the highest-value ones **stable, registry-backed event names and attributes** makes them consumable by any OTLP collector/backend — dashboards, alerting on `ruview.fall.detected`, per-node filtering — without scraping stderr, and the weaver CI gate keeps the schema from drifting silently.

## The events

| Event | Emitted when |
| --- | --- |
| `ruview.node.online` | first frame from a sensing node (CSI or edge-vitals path) |
| `ruview.node.offline` | node evicted after 60 s without frames (now logged per node id) |
| `ruview.presence.changed` | smoothed presence classification flips — transition-only, from all five `SensingUpdate` publish paths via one helper |
| `ruview.vitals.estimate` | breathing / heart-rate estimate with confidences, every 100 ticks |
| `ruview.csi.stats` | capture snapshot (frames processed, active nodes), every 100 ticks |
| `ruview.fall.detected` | edge-vitals fall flag, edge-triggered per node |
| `ruview.mqtt.error` | HA publisher publish/connect failures |
| `ruview.model.loaded` | model loaded via the model-management API |

Rate discipline: nothing new logs at frame rate — transitions and the 100-tick cadence only.

Deliberate omissions: `ruview.room` and `ruview.mesh.channel` were considered but **not** added — the registry only defines attributes for data this crate actually has (rooms live in the calibration crate; channel isn't in the frame structs).

## Verification

- `cargo check -p wifi-densepose-sensing-server` — default features, `--features otel,mqtt`, and `--all-targets`: clean (only pre-existing warnings, verified against the base commit). Each commit in this PR builds independently (default features checked at every commit boundary).
- `weaver registry check --future`: pass; the generate + rustfmt cycle is hash-stable, so the CI no-diff gate is sound.
- Runtime smoke: with the endpoint unset the binary behaves identically; with the endpoint set and no backend listening it runs without panic (export retries in the background).
- End-to-end: ran the instrumented server (`--source simulated`) against a real OTLP logs backend — we dogfood with [Ourios](https://github.com/jensholdgaard/ourios), but the export is backend-neutral and works with any OTLP collector/backend. Logs arrived under `service.name = ruview`, the curated events came through with their registry names/attributes, and template-level queries (event frequency by mined log template, `severity >= warn` filtering, log-template drift between time windows) answered correctly.

## Caveats (honest limits)

- The sensing server has no graceful-shutdown path, so the final OTLP batch flush on kill is best-effort (the guard's `Drop` never runs on SIGKILL/SIGTERM); the periodic batch export bounds the loss to the last export interval.
- `docker/otel-compose.yml` and `docker/otel-collector.yaml` are parser-validated only (no docker CLI on the dev machine); the same env/feature path was verified end-to-end outside compose as described above.

https://claude.ai/code/session_01WQY9wfrfRggqSpMLH8Xj3Y


ruvnet (changes requested):
The opt-in exporter and registry direction looks useful, but I found three correctness/contract gaps that should be addressed before merge:

1. **The default behavior is not actually unchanged.** The new `info!`/`warn!` call sites are compiled and sent through the existing fmt subscriber even without the `otel` feature or endpoint. That means the issue's “nothing is emitted unless the exporter is configured” promise is false and default stderr volume/behavior changes. Either gate the curated emissions with the opt-in feature/runtime state, or revise the contract and explicitly test/document the default logging change.
2. **`ruview.node.online` can be skipped for normal CSI nodes.** The SyncPacket handler creates `node_states[node_id]` at `main.rs:6076`; the first CSI frame later emits online only when `!node_states.contains_key(&node_id)` at `main.rs:6225`. A node that sends sync before CSI therefore never emits the promised first-frame online event. Track first data-frame observation separately (or test `last_frame_time`/an explicit flag) and add a regression test for sync → first CSI.
3. **The claimed semconv enforcement is incomplete.** Event names use generated constants, but attribute keys remain hard-coded string literals at the call sites, and generated `SCHEMA_URL` is never attached to the OTel resource (`telemetry.rs:88` only sets `service.name`). Weaver/codegen can therefore stay green while emitted attributes drift from the registry, and consumers receive no schema URL. Please wire or otherwise validate the emitted keys and attach the registry schema URL (or narrow the documented contract if the SDK cannot represent it).

I have not treated the contributor-hosted demo backend as a blocker, but pinning demo container images by digest would make the one-command pipeline more reproducible and reduce tag-mutation risk.

ruvnet:
Thank you @jensholdgaard — this work has been imported, security-hardened, and merged via #1465. Your authorship is preserved in the merged commit. The replacement keeps OTLP export explicitly opt-in, adds TLS roots and a resolvable schema contract, fixes first-CSI lifecycle reporting, pins demo images, and adds registry/codegen drift checks. Closing this PR as superseded by #1465.

jensholdgaard:
Thanks for the thorough review and for carrying the work over the line in #1465 — the authorship preservation is appreciated. The three findings were sharp; the sync-before-CSI ordering one in particular was a real bug, and I'm glad the merged version fixes it (and attaches the schema URL to the resource).

One technical note on finding 3 (attribute keys as string literals), because it applies to the merged code too: the events are emitted through `tracing`, which is the OTel-sanctioned shape for logs — the Logs Bridge API is explicitly *not* for application code; applications keep their logging library and a bridge maps it to OTLP ([Logs concepts, "Log Appender / Bridge"](https://opentelemetry.io/docs/concepts/signals/logs/#opentelemetry-logging-components)). `tracing` registers callsites statically, so field names are compile-time tokens — `info!({ semconv::RUVIEW_NODE_ID = … })` cannot compile. That's why the generated constants can't reach the call sites in either version: `main` after #1465 carries the same literals (e.g. `"ruview.node.id"` at the online/offline/fall sites).

The gap you identified is still real, though — codegen can stay green while emissions drift. The OTel-idiomatic closure is on the *emission* side: **weaver live-check**, validating actually-emitted telemetry against the registry. That's what OTel's guidance positions weaver for ("measure and validate adherence" — [platform blueprint, guideline 2](https://opentelemetry.io/docs/guidance/blueprints/managed-telemetry-platforms-for-k8s-workloads/#guideline-2); [weaver intro post](https://opentelemetry.io/blog/2025/otel-weaver/)), and it's how opentelemetry-python gates its own semconv emissions in CI (their test-utils run `weaver live check` against captured telemetry). Concretely for RuView: a CI job boots the sensing server with the opt-in exporter, replays a fixture capture, and runs `weaver registry live-check` over the emitted OTLP — at that point both codegen drift *and* call-site drift fail the gate, which no amount of constant-wiring can achieve through a `tracing` bridge.

Happy to send a follow-up PR adding that job if it'd be useful — I run the same gate elsewhere and it ports mechanically. And +1 on digest-pinning the demo images.

PR #1387, opened 2026-07-22
Python components to SOTA: ADR-184..187 (pip publish, P6 bindings, training API, archive/v1 deprecation)

ruvnet:
## Summary

Four ADRs bringing the Python-facing components to SOTA, grounded in verified evidence (open issues, live PyPI state, real CI runs) rather than assumption:

- **ADR-184** — ADR-117 pip publish completion. Root cause of the failing `pip-release` CI (stale/403 PyPI token) fixed, migrated to OIDC Trusted Publishing. wifi-densepose promoted 2.0.0a1 → 2.0.0 stable + ruview 2.0.0. **P3 (real publish) is blocked on a manual step only the PyPI project owner can do**: register a Trusted Publisher on pypi.org for `owner=ruvnet, repo=RuView, workflow=pip-release.yml` for both `wifi-densepose` and `ruview`.
- **ADR-185** — P6 SOTA Python bindings (AETHER re-ID, MERIDIAN cross-domain, MAT triage), each behind an opt-in extra with a bit-for-bit parity harness against native Rust. All three wheel-size concerns resolved (one real fix, one false alarm caught via measurement, one minor cleanup — see ADR for the honest breakdown). Weight-loading capability added to AETHER. Combined `[sota]` extra verified: 218 tests, 2.2MB. Remaining gap, honestly documented: real SOTA accuracy validation needs labeled fixtures + an eval harness — out of this session's scope.
- **ADR-186** — **Accepted.** The dashboard's "Start Training" button was a silent no-op; the real trainer existed but was never wired in (dead, uncompiled module). Reconnected, added `/ws/train/progress`, dashboard honesty for disabled builds, full test coverage. A CI-only test failure surfaced and fixed a real production bug (same-second training-run filename collisions silently overwriting models).
- **ADR-187** — **Accepted.** `archive/v1` formally deprecated; honest three-tier model-weights labeling added to README/user-guide, answering the "does this actually work" / "where are the trained weights" issues directly.

## Test plan
- [x] Rust workspace tests (`cargo test --workspace --no-default-features`) — verified locally and in CI
- [x] Python test suites (aether/meridian/mat/base) — 218 passed combined
- [x] Parity harnesses (binding == native Rust, SHA-256 gated) — all green
- [x] CI run on this branch: Code Quality & Security, Tests (3.10/3.11/3.12), Rust Workspace Tests all green
- [ ] PyPI Trusted Publisher registration (manual, repo owner) — blocks ADR-184 P3/P4


proffesor-for-testing:
## Adversarial review + fixes (updated)

Ran a cross-vendor adversarial pass and built the wheel + ran the real suite on Apple Silicon rather than trusting green x86 CI. All findings are now **fixed and tested** on this branch; a packaging attempt was reverted in favor of the deliberate small-base design.

### Landed

**1. AETHER parity tests are arch-portable** (`test_aether.py`, `aether_parity.rs`, `aether_weights_parity.rs`, `wifi-densepose-aether/tests/golden_parity.rs`). They hashed raw f32 embedding bytes (SHA-256) against a committed golden — f32 + transcendentals (`ln`/`sqrt`/`cos`) aren't bit-reproducible across CPUs, so it only matched the arch that generated it and would fail on the aarch64/macOS-arm wheels this project ships. Now a golden **vector** compared within tolerance (atol=rtol=1e-4) + a whole-vector cosine bound + NaN rejection. A **native** parity test was added inside the std-only aether crate so `cargo test --workspace` runs the native≈golden half in CI (previously only pytest ran; the `.rs` reference tests link against PyO3 and no workflow ran them). Golden proven native-faithful bit-for-bit.

**2. PyO3 bindings validate input instead of panicking** (`python/src/bindings/aether.rs`). `EmbeddingExtractor(n_heads=0)` reached `d_model % n_heads` → panic; `AetherConfig(d_model=100_000)` allocated multi-GB and aborted. Both constructors now return `PyResult` and raise `ValueError` (positive dims, `d_model % n_heads == 0`, a generous cap). +10 pytest cases.

**3. The single-training-job guard is now atomic** (`training_api.rs`). It checked `is_active()` under a shared `state` READ lock, released it, then set `active` later — so two concurrent `POST /train/start` could both pass and spawn jobs, corrupting status/cancel and orphaning a handle. Extracted `claim_training_slot` doing check-and-set in one status-mutex scope. Test: 32 threads race a barrier, exactly one wins — mutation-proven (the split shape fails `left: 3, right: 1`).

**4. Import shims no longer mislead** — they told users to `pip install wifi-densepose[<x>]` (an empty extra); now they point to a source build with `--features <x>` and #1412.

### Deferred to #1412 (not a defect fix)
`pip install wifi-densepose[aether]` raises `ImportError`: release wheels are feature-less (correctly, to keep the base small), and a pip extra cannot enable a cargo feature on a prebuilt wheel — so the extras are inert. Delivering P6 in binary wheels while keeping the base small requires **separate native packages** → **#1412**. Until then P6 is source-build-only, and the shims say so.

> An earlier version of this comment described building SOTA into every published wheel; that was reverted (`2febbb81`) because it abandoned the guarded small-base design and left the published wheel un-budget-checked.

**Verified on aarch64/macOS:** `training_api` 28, full `python/tests` 237, the aether native parity test, and the full workspace all pass; mutation checks confirm the race and NaN tests have teeth.

PR #1396, opened 2026-07-22
fix(python): authenticate SensingClient WebSocket connections

### LINKED ISSUE #1395: python: client cannot authenticate — no token handling at all, so it fails against any auth-enabled server
## Summary

The Python client (`python/wifi_densepose/client/`) sends no `Authorization` header anywhere. Against a sensing server with `RUVIEW_API_TOKEN` set, it simply cannot connect — there is no way to supply a credential.

The TypeScript MCP client (`tools/ruview-mcp/src/http.ts`) has honoured `RUVIEW_API_TOKEN` for some time, so this is a gap between the two clients rather than a missing feature overall.

## Why now

Independent of the OAuth work (ADR-271): this is broken today, for the auth mechanism that already ships.

## Done looks like

- `SensingClient` sends `Authorization: Bearer <token>` when a token is configured, read from `RUVIEW_API_TOKEN` by default and overridable in the constructor.
- Connecting to an auth-enabled server from Python works.
- A test covers the header being sent, and the version-compat concern below.

## Work items

- [ ] Accept a token in `SensingClient.__init__`, defaulting to `RUVIEW_API_TOKEN`
- [ ] Pass it on the WebSocket upgrade
- [ ] Resolve the `websockets` keyword-argument compatibility problem below — deliberately, not by discovery
- [ ] Test with auth on and off

---

## Technical detail

`git grep -n "Authorization\|RUVIEW_API_TOKEN" python/` returns nothing. `python/wifi_densepose/client/{ws,mqtt,ha,primitives}.py` have no credential path.

### The version trap this fix must not walk into

`python/pyproject.toml` pins `websockets>=12.0`, unbounded, and the header keyword argument was **renamed** inside that range:

- `websockets` ≤ 13 → `extra_headers=`
- `websockets` ≥ 14 → `additional_headers=`

`client/ws.py` calls `websockets.connect(...)` with neither today. A naive fix works on the author's machine and raises `TypeError` on a user's, in whichever direction they differ. Either detect the installed version and pass the right keyword, or raise the floor to `>=14` — but pick deliberately.

### Python is not browser-constrained

The browser UI needs a workaround for WS authentication because browsers cannot set `Authorization` on an upgrade request. Python has no such limitation — `websockets` can set the header directly. The Python client should send the bearer on the upgrade and should **not** be routed through any browser-oriented ticket mechanism; that would add a round-trip and a second credential path for no benefit.

The same applies to the TS MCP client, which sends the bearer on REST but should carry it onto WS upgrades too.

### Relationship to ADR-271

ADR-271 adds Cognitum OAuth as an additional, opt-in credential for `/api/v1/*`. `RUVIEW_API_TOKEN` remains supported, so this fix is worth landing on its own and is a prerequisite for the Python client participating in OAuth later — the transport plumbing is the same either way.

### Note

There is no Python OAuth client anywhere in the Cognitum ecosystem today (the existing implementations are Rust and TypeScript), so the eventual Python OAuth login flow will be written from scratch. That is a separate, larger piece of work; this issue is only about sending a bearer token that the user already has.

pengqima:
## Summary

- Add bearer-token authentication to the Python `SensingClient`.
- Read the token from `RUVIEW_API_TOKEN` by default.
- Allow callers to override the token through the constructor.
- Support both WebSocket header APIs used across the declared `websockets>=12` dependency range.

## Details

`SensingClient` now sends:

```text
Authorization: Bearer <token>
```

during the WebSocket upgrade.

websockets renamed extra_headers to additional_headers in version 14. The client inspects the installed connect() API and deliberately selects the supported argument, preserving compatibility with versions 12, 13, 14, and later.

Passing token="" explicitly disables authentication even when RUVIEW_API_TOKEN is set.

## Validation

Client test suite with websockets==12.0: 11 passed
Client test suite with websockets==15.0.1: 11 passed
Python compile check passed
Targeted Ruff checks passed

Closes #1395

### DIFF
--- a/python/tests/test_client_ws.py
+++ b/python/tests/test_client_ws.py
@@ -10,7 +10,7 @@ binary, but the wire protocol is the contract under test here.
 
 from __future__ import annotations
 
-import asyncio
+import inspect
 import json
 from typing import Any
 
@@ -76,11 +76,32 @@ async def _handler(websocket: Any) -> None:
     await websocket.send(json.dumps({"type": "edge_vitals", "node_id": "post-bad-frame"}))
 
 
+async def _auth_handler(websocket: Any) -> None:
+    request = getattr(websocket, "request", None)
+    headers = request.headers if request is not None else websocket.request_headers
+    if headers.get("Authorization") != "Bearer test-token":
+        await websocket.close(code=1008, reason="missing or invalid bearer token")
+        return
+    await websocket.send(json.dumps(_FIXTURE_MESSAGES[0]))
+
+
 @pytest.fixture
 async def ws_server() -> Any:
     """Start a websocket server on a random port; yield the bound URL."""
     server = await websockets.serve(_handler, "127.0.0.1", 0)
     # Get the bound port (host="127.0.0.1" returns one socket).
+    port = server.sockets[0].getsockname()[1]  # type: ignore[union-attr]
+    try:
+        yield f"ws://127.0.0.1:{port}/ws/sensing"
+    finally:
+        server.close()
+        await server.wait_closed()
+
+
+@pytest.fixture
+async def authenticated_ws_server() -> Any:
+    """Start a WS server that rejects clients without the expected bearer."""
+    server = await websockets.serve(_auth_handler, "127.0.0.1", 0)
     port = server.sockets[0].getsockname()[1]  # type: ignore[union-attr]
     try:
         yield f"ws://127.0.0.1:{port}/ws/sensing"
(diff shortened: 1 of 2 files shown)


PrinceAudre:
Heads-up: #1396 and #1400 implement the same feature — bearer-token auth on the `SensingClient` WebSocket upgrade — and edit the same lines of `python/wifi_densepose/client/ws.py` and `python/tests/test_client_ws.py` (both close #1395). So only one can merge; the other will conflict.

Both look sound. A small point in this one's (#1396) favor: it adds a dedicated unit test that monkeypatches `websockets.connect`'s signature to verify the `additional_headers` (websockets ≥14) vs `extra_headers` (<14) branch — the one non-trivial piece here. Worth picking one and closing the other.

PR #1397, opened 2026-07-22
Cognitum OAuth for RuView: resource server, sign-in, and WebSocket authentication (ADR-271/272)

proffesor-for-testing:
## The question to answer first

**Is this the integration you meant?**

This branch makes RuView an OAuth **resource server** — a user signs in to *their own* RuView instance with their Cognitum account, and RuView verifies the token offline against `auth.cognitum.one`'s published JWKS.

That reading was inferred, not confirmed. The evidence: RuView makes **no authenticated calls to any Cognitum API**. Every outbound `https://` in the sensing server is GCS, Netgear/Mist, github or prometheus — the only Cognitum dependency is the ADR-102 registry fetch, an anonymous GET against a public bucket.

Worth stating because it inverts the usual direction. meta-proxy, musica, metaharness and the dashboard CLI are all OAuth **clients** — they obtain a token so the app can *call* Cognitum. RuView doesn't call Cognitum. The only directly relevant precedent is `meta-llm/src/auth/oauthBearer.ts`, and **RuView is the org's first Rust resource server** for these tokens.

If you wanted RuView to *call* Cognitum services, that is a plane that doesn't exist yet and a materially larger piece of work.

---

## What this does

Replaces "everyone shares one `RUVIEW_API_TOKEN` string" with per-user Cognitum identity, and closes three authentication holes found along the way.

| | |
|---|---|
| **Verify** | `ruview-auth` — offline ES256/JWKS verification of Cognitum access tokens |
| **Sign in (CLI)** | `wifi-densepose login` / `logout` / `whoami` — PKCE, loopback, OOB paste for headless Pis |
| **Sign in (browser)** | `/oauth/start`, `/oauth/callback`, `/oauth/logout`, `/oauth/status` + a QuickSettings panel. The server runs the flow and issues a signed session cookie; **the browser never holds an access token** |
| **Gate** | Deny-by-default, scope-separated: `sensing:read` vs `sensing:admin` |
| **WebSockets** | Previously unauthenticated even with auth ON — now gated by bearer or single-use ticket (ADR-272) |

Platform side already merged and **deployed to production**: cognitum-one/dashboard#116 registers the `ruview` client with `sensing:read`/`sensing:admin` (ADR-060) — the first non-`inference` scopes the AS has issued.

Backwards compatible: with OAuth unconfigured, behaviour is byte-identical to today. `RUVIEW_API_TOKEN` deployments keep working.

---

## Please read this part

**More than a dozen defects in this PR were found by adversarial review, not by the test suite — and the suite was green for every one of them.**

Every single one lived at a **seam**: fixtures vs production, library vs caller, one router vs another, one port vs another, server vs service worker, REST route vs its WebSocket twin. Four were specifically *"implemented, unit-tested, and never actually called."*

That pattern is the most useful thing in this branch, and it is why the diff is larger than the feature warrants.

The twelve worth your attention:

| Defect | Why the suite couldn't see it |
|---|---|
| Verifier required an `iss` claim Cognitum **never issues** — rejected every real token | Fixtures invented the claim. 41 tests green. |
| The dedicated `--ws-port` listener had **no auth layer at all** | Every auth test built its own `Router`. That is the port the shipped UI connects to. |
| `/ws/field` was `.merge()`d **after** the auth layer | axum layers wrap only what is already registered. |
| `/api/field` served the live presence/pose/vitals stream **unauthenticated on both ports** | The gate protected `/api/v1/*` by prefix. `/ws/field` was gated in this PR; its REST twin one segment over was not. |
| Scope gate was **fail-open** — `/api/v1/adaptive/train` reachable with `sensing:read` | A denylist missed it. Now inverted. |
| Blocking JWKS fetch under a `std::sync::Mutex` held across `.await` | Filed independently by two reviewers. |
| Two `Set-Cookie` headers **collapsed into one** — sign-in was a silent no-op | axum's array form replaces same-name headers. Caught with `curl -D-`. |
| Service worker cached `/oauth/status`, freezing sign-in until a hard reload | The Cache API is **not** the HTTP cache and ignores `Cache-Control` entirely. |
| Service worker cached authenticated `/api/*` responses | Sibling of the above: sign in as A → sign out → sign in as B → lose network → B is served A's sensing data. |
| A malformed `RUVIEW_OAUTH_CLIENT_IDS` (`","`) **silently disabled the audience check** | Empty is the opt-out sentinel downstream. |
| Credentials world-readable for a window on **every** refresh | `fs::write` then `chmod`. The existing permissions test asserted on the *final* file and passed throughout. |
| The one test that boots the real binary **passed vacuously when the server failed to boot** | It returned an `Option` every test turned into `return`, reporting "5 passed" with zero assertions. |

---

## Verification

Not "the tests pass" — each claim below has evidence behind it.

**Against production, with a real login:**

```
no credential            GET  /api/v1/models         -> 401
garbage bearer           GET  /api/v1/models         -> 401
real sensing:read token  GET  /api/v1/models         -> 200
real sensing:read token  GET  /api/v1/recording/list -> 200
real sensing:read token  POST /api/v1/train/start    -> 401
real sensing:read token  POST /api/v1/train/stop     -> 401
```

**Browser sign-in** was driven end-to-end in a real browser, not asserted from unit tests — that is how the two service-worker defects were found and confirmed fixed.

**Mutation-verified, not assumed.** Every security-critical guard had its mutation applied and the failure observed:

| Mutation | Result |
|---|---|
| delete the OAuth `state` comparison | `a_callback_whose_state_does_not_match_is_refused` **FAILED** |
| `is_live().then_some(s)` → `Some(s)` | 2 tests **FAILED** (expired sessions would authenticate forever) |
| `session.has_scope(required)` → `true` | `a_read_scoped_browser_session_cannot_delete_or_train` **FAILED** |
| revert the gate to prefix-matching | `http port served /api/field to an anonymous caller` **FAILED** |
| revert the JWKS stale-path guard | 26 outbound fetches vs 1 |
| revert to first-match cookie reads | `a_shadowing_cookie_cannot_silently_take_over_the_session` **FAILED** |

**Counts:** `ruview-auth` 89 tests (`--all-features`) / 48 (`--no-default-features`) · `bearer_auth` 60 · `browser_session` 27 · `auth_wiring` 7 (boots the real binary, probes both listeners) · UI 22 · workspace 176 suites clean.

**CI gap closed:** the workspace runs `--no-default-features`, which switches off `ruview-auth`'s `login`/`pkce` features — **41 of its 89 tests never executed in CI**, including the entire interactive sign-in path. Added a dedicated `--all-features` step. Nothing ran the UI JavaScript at all, so there is now a `ui-tests` job too.

---

## Three design decisions worth disagreeing with

1. **`__Host-` cookie prefix deliberately omitted.** It requires `Secure`, and RuView is routinely reached over plain HTTP on a LAN, where such a cookie is never sent and sign-in silently fails. The cost is that a same-host writer can *shadow* a session cookie — so both credential paths now accept only when **exactly one** candidate verifies. That blocks takeover; an attacker can still cause a refusal.

2. **Session TTL 1 hour, down from 12, and the browser is read-only.** The session outlives the ~15-minute access token and Cognitum publishes no introspection endpoint, so a stale session is authority nobody can revoke — an hour is the size of that hole, chosen deliberately over building a server-side refresh-token store. Browser sign-in requests `sensing:read` **only**, permanently: destructive operations live in the CLI where `--admin` is explicit. So the UI's delete controls return 401 from a Cognitum session (the token-paste field is unchanged), and the `ADMIN_REVERIFY_SECS` freshness gate is a fail-closed **backstop** for a case that cannot currently arise, not an active control.

3. **`RUVIEW_WS_LEGACY_UNAUTHENTICATED`** restores the old unauthenticated WebSocket behaviour for deployments that cannot update server and UI in lockstep. It is scoped to WebSockets, warns loudly on boot, and should be deleted once the shipped UI fetches tickets.

---

## Known incomplete — deliberately

- **`redirect_uri` defaults to `http://127.0.0.1:8080/oauth/callback`**, overridable only via `RUVIEW_PUBLIC_BASE_URL`. Browser sign-in therefore works only on a host reached at exactly that origin; `http://localhost:8080` or a LAN IP cannot complete the flow. PKCE keeps the code unexchangeable, so this is a broken flow, not a token leak. Deriving it from the request is the fix, and deriving a redirect URI from attacker-controllable headers deserves its own decision.
- **The CLI credential is not consumed by any shipped client.** `wifi-densepose login` obtains and stores a token; no CLI subcommand, MCP server or Python client reads `~/.ruview/credentials.json` yet.
- **The Python client cannot authenticate at all** — `ws.py` passes no headers, so every published `wifi-densepose[client]` consumer 401s once auth is enabled. ADR-272 previously claimed the opposite; corrected in this branch. Tracked as #1395.
- **The browser session cookie branch in `require_bearer` is unreachable** when an `Authorization` header is present. It fails closed, so it is not a hole, but "try the next credential" is what the code reads like. Pinned by a test rather than silently left ambiguous.

---

## On merge

`sensing-server-docker.yml` is path-filtered to `v2/crates/wifi-densepose-sensing-server/**`, so merging **automatically builds and pushes to Docker Hub and ghcr.io tagged `latest`** — no approval gate. Worth knowing before merging rather than after.

`cd.yml` deploys nothing: it fails at `Verify image exists` before reaching any deploy job, and has done so on every recent run. Pre-existing, unrelated to this branch, probably worth its own issue.

No npm/pip/desktop release is triggered — those are path-filtered elsewhere or manual.

---

## ADRs

- **ADR-271** — RuView as a Cognitum OAuth resource server (+ browser sign-in, + P1/P2/P3 resolutions)
- **ADR-272** — WebSocket authentication tickets
- **cognitum-one/dashboard ADR-060** — OAuth scopes beyond `inference` (merged, deployed)

Both RuView ADRs contained claims the code contradicted; an adversarial pass caught them and they are corrected here rather than left to mislead.


proffesor-for-testing:
## Decision: browser sessions are read-only, permanently

Recorded here because it changes what two earlier commits mean, and because the PR description said the opposite until now.

**Browser-side admin is not wanted.** `BROWSER_SIGNIN_SCOPE` stays `sensing:read`. The escalate-on-demand design sketched while this was still open is not being built. Destructive operations — training, model delete, recording delete — keep their home in the CLI, where `--admin` is explicit and typed by a person. Routing them through a browser would mean either asking every user to consent to delete capability in order to watch a stream, or building a second consent flow to avoid that.

### What changed as a result

- **Removed the client-side step-up redirect** (`ui/services/api.service.js`). It caught an RFC 6750 challenge that can never be issued to a browser, and it ended in `return new Promise(() => {})`. Had any other 401 ever grown that header, every caller awaiting it would have hung forever — no error, no timeout. Dead code with a trap in it is worse than no code.
- **Kept `ADMIN_REVERIFY_SECS` as a server-side backstop.** Fail-closed and free, so if the requested scope is ever widened the freshness requirement is already in place. It is documented at its definition as a backstop specifically so nobody reads its passing tests as evidence the control is exercised — those tests reach it through a crate-internal seam that mints an admin cookie the real flow does not produce.
- **The UI's admin controls return 401 from a Cognitum browser session**, and that is now intended rather than a gap. The manual token-paste field is unchanged and still carries whatever authority the pasted token has.

### A correction worth flagging

An earlier revision of ADR-271 claimed that capping the session at `sensing:read` had been "considered and rejected, because the dashboard genuinely performs admin operations." That was **false**. `/oauth/start` already requested `sensing:read` and nothing else, deliberately, with a comment saying so — the browser session was already read-only, and the breakage the ADR claimed capping would cause was simply the existing behaviour.

A cross-vendor pre-merge sweep caught it. Corrected in the ADR rather than quietly dropped, since the wrong version had been up for several commits.

### Session TTL

ADR-271 now stops hedging: **one hour**, chosen over building a server-side refresh-token store. The residual is named rather than implied closed — within that hour, a revoked Cognitum grant still reads sensing data through an existing browser session. There is no introspection endpoint, so nothing short of the refresh-token store closes it.

### Also added

`the_cookie_max_age_matches_the_session_expiry` — a test the previous ADR revision asked for and nobody had written. `Max-Age` and the payload's `exp` are two independent expressions of one lifetime; drift means either the browser presents a session we reject, or we hold authority the browser already discarded.

**Verified:** workspace 176 suites clean, UI 22/22.

Issue #1004, opened 2026-06-11
[BUG] sensing-server --source auto (default) latches to 'simulate' at boot and never binds UDP :5005 → live CSI silently ignored


markt-heximal:
## Summary
`wifi-densepose-sensing-server` with `--source auto` — which is the **default** (`cli.rs` / `main.rs`: `#[arg(long, default_value = "auto")]`) — does a **one-shot** hardware probe at startup. If no ESP32 CSI happens to be flowing **at that instant** (totally normal at boot, before the sensors/forwarder are up), it latches to `simulate` for the **entire process lifetime** and **never binds the UDP :5005 receiver**, so CSI that starts flowing seconds later is silently ignored and the server serves simulated poses forever.

## Evidence (current source, `wifi-densepose-sensing-server/src/main.rs`)
```rust
// ~6424  Auto-detect data source
let source = match args.source.as_str() {
    "auto" => {
        if probe_esp32(args.udp_port).await { "esp32" }       // one-shot, at startup only
        else if probe_windows_wifi().await { "wifi" }
        else { "simulate" }                                    // no CSI right now => simulate, permanently
    }
    other => other,
};
```
Then the data-source tasks are spawned **by that resolved string**: `udp_receiver_task(state, args.udp_port)` is spawned only for `"esp32"`, while `"simulate"` spawns `simulated_data_task`. There is no re-probe and no UDP bind in the simulate path, so a node that boots before CSI is flowing never recovers without a restart.

## Reproduce
1. Start `sensing-server` (defaults: `--source auto`) with no ESP32 frames arriving yet.
2. Log shows `No hardware detected, using simulation` → `Data source: simulate`; nothing binds UDP :5005.
3. Start the ESP32/CSI forwarder afterward → frames hit a closed port; dashboard stays on simulated data indefinitely.

Workaround in the field: force `--source esp32` in the systemd unit (which unconditionally binds :5005). Several of our appliance units were silently on `simulate` because the stock unit relied on `auto` and only worked when CSI happened to be live at boot.

## Impact
Default configuration is boot-order-fragile: real deployments silently fall back to simulated output if the server starts before CSI flows. Easy to miss (the UI looks "live").

## Suggested fix (any)
- Always bind the UDP :5005 receiver regardless of resolved source, and **promote** `simulate→esp32` when the first real frame arrives (the inverse transition — `esp32→offline` after a 5s gap — already exists in `effective_source()`).
[...]

PR #1399, opened 2026-07-22
feat(sensing-server): make UDP CSI receiver bind address configurable

### LINKED ISSUE #1394: sensing-server: UDP CSI receiver binds 0.0.0.0 with no bind-address option, while HTTP defaults to 127.0.0.1
## Summary

`wifi-densepose-sensing-server` lets an operator bind the HTTP API to a chosen address (`--bind-addr` / `SENSING_BIND_ADDR`, default `127.0.0.1`), but the UDP CSI receiver binds `0.0.0.0` unconditionally, with no equivalent flag. An operator who has deliberately kept the API on loopback still has a CSI ingest port open on every interface.

## Why now

Not a newly-introduced bug and not an alarm — unauthenticated CSI datagrams are a documented design position for the data plane (ESP32 nodes are data-plane sensors, not attested mesh members). This is narrower: the two listeners disagree about what "localhost only" means, and only one of them is configurable.

## Done looks like

- The UDP receiver accepts a bind address the same way the HTTP listener does, defaulting to today's behaviour so nothing changes for existing deployments.
- Optionally, a peer allow-list so a hub can accept frames only from its paired nodes.
- `--help` makes the exposure explicit either way.

## Work items

- [ ] Add a UDP bind-address flag + env var mirroring `SENSING_BIND_ADDR`
- [ ] Keep `0.0.0.0` as the default (changing it would break real deployments)
- [ ] Consider an optional source-address allow-list
- [ ] Note the exposure in the startup log, as the edge-registry fetch already does for its outbound dependency

---

## Technical detail

`v2/crates/wifi-densepose-sensing-server/src/main.rs`:

- HTTP: `#[arg(long, default_value = "127.0.0.1", env = "SENSING_BIND_ADDR")]`
- UDP (`udp_receiver_task`): `let addr = format!("0.0.0.0:{udp_port}");` — hardcoded, port configurable, address not.

`recv_from` passes the datagram straight to `parse_esp32_frame`; the sender address is discarded (`_`). There is no PSK, HMAC, allow-list or replay protection on the CSI path — grep for `psk|shared_secret|hmac` across the crate's `src/` returns nothing.

### Interaction with the `#1004` source state machine

In `auto` mode the receiver is always bound, and the first real frame promotes `source` → `esp32`. That behaviour is correct and was fixed deliberately (a one-shot boot probe used to latch `simulate` forever). The side effect is that anything able to reach the port can also flip a server out of simulate mode into presenting injected frames as live sensor data.

### Scope of impact

- **Bundled desktop install (ADR-055)** — server is a child process, traffic is loopback. The open UDP port is the only part of that topology not actually restricted to loopback.
- **LAN / hub deployment** — the intended path is a direct link to a paired Seed or V0 device. The pairing credential exists (`seed_token` bearer, ADR-066, plus a device-id-bound PSK for the V0 `/edge` swarm endpoints), but it is not applied to the UDP data plane.

### Deliberately out of scope

Authenticating the CSI datagrams themselves (per-node HMAC, DTLS) is a device-plane design change spanning firmware and consumers. This issue is only about making the listener as configurable as its HTTP sibling.

Related, unrelated to this fix: the mapping between the V0 `/edge` PSK and `provision.py --seed-token` is documented as an open gap elsewhere.

sridhar-3009:
Fixes #1394.

## Problem

`wifi-densepose-sensing-server` lets an operator restrict the HTTP API to loopback via `--bind-addr` (default `127.0.0.1`), but the UDP CSI receiver bound `0.0.0.0` unconditionally with no equivalent flag. An operator who deliberately kept the API on loopback still had an unauthenticated CSI ingest port open on every interface.

## Change

- Added `--udp-bind-addr` / `SENSING_UDP_BIND_ADDR`, defaulting to `0.0.0.0` — **no behaviour change for existing deployments**.
- Kept it as its own flag rather than reusing `--bind-addr`, since the common real-hardware topology is HTTP on loopback with UDP still open to the LAN so paired ESP32/MediaTek/Qualcomm/RTL8720F nodes can reach it.
- Threaded the new address through both places that bind the UDP port:
  - `probe_esp32` — the one-shot 2s auto-detect probe run at boot in `auto` mode
  - `udp_receiver_task` — the persistent listener
- Startup log now prints the actual UDP bind address (was hardcoded to `0.0.0.0` in the log even before this change would have made it wrong), and adds an explicit exposure note — `LAN-reachable, unauthenticated` vs `restricted` — mirroring the existing `bind_ip.is_unspecified()` pattern used for the HTTP bearer-auth warning.

## Out of scope (per the issue)

Authenticating the CSI datagrams themselves (per-node HMAC/PSK, DTLS, source allow-list) is a device-plane design change spanning firmware and consumers — this PR is only about making the listener's bind address as configurable as its HTTP sibling.

## Testing

I don't have ESP32/CSI hardware to exercise this against a live radio, so I traced both UDP bind call sites and the `auto`-mode startup sequence by reading the surrounding code (source-state-machine from #1004, the boot probe timing) rather than running it end-to-end. Happy to adjust if a maintainer can test against real hardware.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -90,6 +90,16 @@ struct Args {
     #[arg(long, default_value = "5005")]
     udp_port: u16,
 
+    /// Bind address for the UDP CSI receiver (default 0.0.0.0, unchanged from
+    /// prior behaviour — ESP32/MediaTek/Qualcomm/RTL8720F nodes are typically
+    /// on the LAN). Set to 127.0.0.1 to restrict CSI ingest to loopback, e.g.
+    /// for a bundled desktop install where the HTTP API is already loopback-only
+    /// (see `bind_addr`) and no LAN hardware is paired. Mirrors `--bind-addr`'s
+    /// shape but is independently configurable, since the common real-hardware
+    /// case is HTTP on loopback with UDP still open to the LAN. (#1394)
+    #[arg(long, default_value = "0.0.0.0", env = "SENSING_UDP_BIND_ADDR")]
+    udp_bind_addr: String,
+
     /// Path to UI static files (repo `ui/`; from `v2/` use `../ui` or rely on auto-detect)
     #[arg(long, default_value = "../ui")]
     ui_path: PathBuf,
@@ -2870,8 +2880,8 @@ async fn probe_windows_wifi() -> bool {
 }
 
 /// Probe if ESP32 is streaming on UDP port
-async fn probe_esp32(port: u16) -> bool {
-    let addr = format!("0.0.0.0:{port}");
+async fn probe_esp32(port: u16, bind_addr: &str) -> bool {
+    let addr = format!("{bind_addr}:{port}");
     match UdpSocket::bind(&addr).await {
         Ok(sock) => {
             // 2048 covers the largest ADR-018 frame: an ESP32-C6 HE-SU
@@ -5587,8 +5597,8 @@ async fn info_page() -> Html<String> {
 
 // ── UDP receiver task ────────────────────────────────────────────────────────
 
-async fn udp_receiver_task(state: SharedState, udp_port: u16) {
-    let addr = format!("0.0.0.0:{udp_port}");
+async fn udp_receiver_task(state: SharedState, udp_port: u16, udp_bind_addr: String) {
+    let addr = format!("{udp_bind_addr}:{udp_port}");
     let socket = match UdpSocket::bind(&addr).await {
         Ok(s) => {
             info!("UDP listening on {addr} for ESP32, MediaTek, Qualcomm CSI, and RTL8720F radar frames");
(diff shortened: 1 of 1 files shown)


PrinceAudre:
Clean, behavior-preserving change (default stays `0.0.0.0`, both bind sites updated). One small nit:

`let udp_ip: std::net::IpAddr = args.udp_bind_addr.parse().expect("Invalid --udp-bind-addr ...")` — since clap takes the arg as a `String`, a malformed value panics at startup (with a backtrace) instead of producing a clean CLI usage error. Letting clap parse it directly is nicer:

```rust
#[arg(long, env = "SENSING_UDP_BIND_ADDR", default_value = "0.0.0.0",
      value_parser = clap::value_parser!(std::net::IpAddr))]
udp_bind_addr: std::net::IpAddr,
```

Then drop the `.parse().expect(...)` at the use site. Fail-fast is fine; this just makes bad input a usage error rather than a panic.

ruvnet:
Closing as superseded by the more complete ADR-296 implementation already on `main` (`34c980400`). It includes configurable binding plus a loopback-safe default, source IP/CIDR allowlisting, fail-closed routable binding, explicit insecure override, startup reporting, and regression tests. Thank you for identifying and scoping the original exposure.

PR #1400, opened 2026-07-22
feat(python-client): send bearer token on WS upgrade

### LINKED ISSUE #1395: python: client cannot authenticate — no token handling at all, so it fails against any auth-enabled server

sridhar-3009:
Fixes #1395.

## Problem

`SensingClient` (`python/wifi_densepose/client/ws.py`) had no credential path at all — `git grep -n "Authorization\|RUVIEW_API_TOKEN" python/` returned nothing. Against a sensing-server with `RUVIEW_API_TOKEN` set, the Python client simply cannot connect. The TS MCP client (`tools/ruview-mcp/src/http.ts`) has honoured this token for a while, so this was a gap between the two clients rather than a missing feature overall.

## Fix

- `SensingClient.__init__` accepts a `token` keyword, defaulting to `RUVIEW_API_TOKEN` when not passed explicitly.
- When a token is present, `__aenter__` sends `Authorization: Bearer <token>` on the WS upgrade. Python isn't browser-constrained the way the web UI is, so this goes straight on the upgrade request — no ticket/round-trip indirection, per the issue's explicit guidance.
- Handled the `websockets>=12.0` keyword rename **deliberately**, per the issue's callout: the header kwarg to `websockets.connect()` is `extra_headers` below version 14 and `additional_headers` from 14 onward. Detected via `inspect.signature(websockets.connect).parameters` at import time, not by parsing `__version__` — so it can't be thrown off by a version-string format change on either side of the cutover.

## Testing

I didn't just trust the issue's description of the version split — I installed `websockets` 12.0, 13.1, and 16.1.1 in three isolated venvs and inspected the actual `connect()` signature in each to confirm exactly where `extra_headers` becomes `additional_headers` before writing the branch logic.

Added three tests to `test_client_ws.py` using the same in-process `websockets.serve()` pattern the file already uses:
- token passed to the constructor arrives as the correct `Authorization: Bearer <token>` header
- no token passed and `RUVIEW_API_TOKEN` unset → no `Authorization` header sent at all (not an empty one)
- `RUVIEW_API_TOKEN` env var is honoured when no token is passed explicitly

Ran the full `test_client_ws.py` suite (9 tests, including the 3 new ones) against both `websockets` 13.1 and 16.1.1 in separate venvs — all pass on both sides of the keyword rename. Also confirmed the change doesn't newly break `ruff check` beyond one pre-existing style debt already present in this file (`Optional[X]` vs `X | None`) that I matched rather than mixed styles.

### DIFF
--- a/python/tests/test_client_ws.py
+++ b/python/tests/test_client_ws.py
@@ -65,6 +65,15 @@ _FIXTURE_MESSAGES = [
 ]
 
 
+def _request_headers(websocket: Any) -> Any:
+    """Return the upgrade-request headers regardless of websockets version:
+    the legacy server protocol (<14, still reachable via >=12.0's floor)
+    exposes `request_headers`; the newer asyncio-based server exposes
+    `request.headers` instead."""
+    legacy = getattr(websocket, "request_headers", None)
+    return legacy if legacy is not None else websocket.request.headers
+
+
 async def _handler(websocket: Any) -> None:
     for msg in _FIXTURE_MESSAGES:
         await websocket.send(json.dumps(msg))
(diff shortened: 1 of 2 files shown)


PrinceAudre:
Heads-up: #1400 and #1396 implement the same feature — bearer-token auth on the `SensingClient` WebSocket upgrade — editing the same lines of `python/wifi_densepose/client/ws.py` and `python/tests/test_client_ws.py` (both close #1395). So only one can merge; the other will conflict.

Both look sound. This one (#1400) fits the file's existing `Optional[...]` style and resolves the `additional_headers`/`extra_headers` kwarg once at import; #1396 adds a dedicated unit test for that version-split branch. Either is a fine choice — just pick one and close the other.

PR #1406, opened 2026-07-23
Fix: short pytest param ids to avoid Windows env-length error

ekariminmhc-ui:
Pytest parametrized tests generated extremely long node ids on Windows which triggered a ValueError due to environment variable length during test teardown. This change adds short explicit ids for ADVERSARIAL_PAYLOADS in tests/test_invariant_Cargo.py to avoid massive node ids while preserving all payloads. Verified: full test suite passes (34 passed, 1 warning) on Windows.\n\nNo functional change to test logic; only test metadata (ids).

### DIFF
--- a/tests/test_invariant_Cargo.py
+++ b/tests/test_invariant_Cargo.py
@@ -93,7 +93,7 @@ def simulate_version_comparison(version_str: str) -> bool:
         return False
 
 
-@pytest.mark.parametrize("payload", ADVERSARIAL_PAYLOADS)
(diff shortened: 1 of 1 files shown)


ekariminmhc-ui:
Automated change: short explicit pytest param ids added to tests/test_invariant_Cargo.py to avoid extremely long pytest node ids on Windows which caused a ValueError during test teardown. Verified locally on Windows: full test suite passes (34 passed, 1 warning). I don't have permission to merge into ruvnet/RuView; please review and merge if acceptable. — Copilot

PR #1418, opened 2026-07-24
fix(firmware): add DEGRADED state recovery path and N16R8 sdkconfig

### LINKED ISSUE #1401: Interesting Concept but Doesn't Work
# Experience Report: ESP32-S3 N16R8 + RuView Real-World Deployment

## Hardware
- **5x AITRIP ESP32-S3 N16R8** (ESP32-S3-WROOM-1 module, 16MB flash, 8MB Octal PSRAM, no display)
- Connected via native USB-JTAG (`/dev/ttyACM0`) and FT232 UART (`/dev/ttyUSB0`)
- WiFi: Standard 2.4 GHz AP (SSID: Blose)

### 1. Pre-built firmware binaries do NOT work on N16R8 boards
The release binaries at `firmware/esp32-csi-node/release_bins/` produce boards that:
- Connect to WiFi successfully
- Boot into the adaptive controller
- **Immediately enter state 8 (ADAPT_STATE_DEGRADED)** and never leave
- Only send 60-byte feature-state heartbeat packets (type `0x06`) — **no CSI frames ever**
- CSI callback count stays at 0 permanently — `pkt_yield_per_sec` is 0, so the degraded gate at `adaptive_controller_decide.c:27` (`obs->pkt_yield_per_sec < cfg->min_pkt_yield`) immediately locks the board in DEGRADED state with no recovery path

### 2. Building firmware from source fixes it
Building the firmware using the IDF Docker container:
```bash
cd firmware/esp32-csi-node
docker run --rm -v "$(pwd):/project" -w /project \
  espressif/idf:v5.4 bash -c \
  'rm -rf build sdkconfig && idf.py set-target esp32s3 && idf.py build'
```
Self-built firmware boots to **state 6 (ADAPT_STATE_SENSE_ACTIVE)** with real CSI frames streaming at 30-40 fps. Boards successfully enter sensing states and respond to motion/presence.

### 3. Flashing considerations
- `--flash_mode dio --flash_size 8MB` worked fine for 16MB flash boards — the partition table only uses ~7MB so the extra space is unused
- The eFuse on these boards reports `Flash type: quad (4 data lines)` but DIO mode still functions
- Must flash: `bootloader.bin`, `partition-table.bin`, `ota_data_initial.bin`, `esp32-csi-node.bin` at `0x20000`
- Provision with `--edge-tier 2 --force-partial`

### 4. Docker sensing server works well
- `ruvnet/wifi-densepose:latest` receives and processes real ESP32 CSI data
- Use `--disable-host-validation` for LAN access
- Multi-node auto-detection works — nodes appear as they come online
- WebSocket at `:3001/ws/sensing` and REST API at `:3000/api/v1/sensing/latest` both reflect real data

### 5. Model loading
- The `wifi-densepose-v1.rvf` in the `docker/` directory loads but produces ~20K+ engine errors and gets trust-demoted
- HuggingFace model at `ruvnet/wifi-densepose-pretrained` can be converted to RVF via `--convert-model` (after fixing safetensors header null-padding issue), but still gets trust-demoted with hundreds of engine errors
- Person detection (`n_persons`) appears to use a heuristic, not the model — documented as "slot-capacity heuristic, not a learned classifier" in the firmware readme

### 6. Current observations
- 3 nodes streaming real CSI simultaneously
- Person detection state flickers between present/moving and still even with no one in the room (sleeping family on another floor)
- Adaptive controller shows `motion=1.00 presence=3.79` in state 6 even in empty rooms
- The `yield=0pps` problem in DEGRADED state has no recovery mechanism — once a board enters state 8 it stays there permanently

## Questions / Suggestions

1. Could the pre-built release binaries be rebuilt with the current sdkconfig? The ones in `release_bins/` appear to have an incompatible config for boards with Octal PSRAM
2. The DEGRADED state should have a timeout or recovery path — if yield recovers above threshold, the board should leave state 8
3. Training documentation would be valuable — what does "empty room" mean for calibration? Do pets trigger false positives? How many minutes of calibration data are needed?
4. The trust/engine-error demotion system needs documentation — what causes demotion, how to diagnose, how to recover

## Current Setup (working)
- 3/5 boards flashed with self-built firmware, provisioned via `provision.py`
- Docker server receiving UDP on port 5005, dashboard at port 3000
- HuggingFace model converted and loaded (but trust-demoted)
- All nodes send to `192.168.1.4:5005` with unique `--node-id`

alvinjaison:
## Summary

Fixes #1401 — ESP32-S3 N16R8 boards entering DEGRADED state permanently with no recovery path.

## Problem

The adaptive controller DEGRADED state (state 8) had no exit mechanism. Once `pkt_yield_per_sec` dropped below `min_pkt_yield`, the board stayed in DEGRADED permanently — even if yield later recovered.

On N16R8 boards the pre-built binaries lacked PSRAM config, so CSI callbacks never fired, yield stayed at 0, and DEGRADED was immediate and permanent.

### DEGRADED recovery with hysteresis (`adaptive_controller_decide.c`)

- If a board is in DEGRADED and yield recovers above `min_pkt_yield` for `degraded_recovery_ticks` consecutive fast-loop ticks (default 15 = ~3s at 200ms), it transitions back to `SENSE_IDLE`
- Counter resets on any yield drop mid-recovery, preventing flapping
- New config field `degraded_recovery_ticks` (Kconfig: `CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS`)

### N16R8 sdkconfig overlay (`sdkconfig.defaults.n16r8`)

- New overlay for ESP32-S3 WROOM-1 N16R8 modules (16MB flash, 8MB Octal PSRAM)
- Adds `CONFIG_SPIRAM=y`, `CONFIG_SPIRAM_MODE_OCT=y`, `CONFIG_SPIRAM_SPEED_80M=y`
- Disables display (no QSPI contention on these boards)

### Unit tests (`tests/host/test_adaptive_controller.c`)

- 5 new test cases covering recovery scenarios
- Updated existing tests for new function signature
- All 82 host assertions pass, 0 failures

## Testing

- **Host tests:** `make check` in `tests/host/` — 82 pass, 0 fail
- **Docker build:** `espressif/idf:v5.4` builds successfully with N16R8 overlay (binary: 920KB, 56% partition free)
- **Hardware:** Not tested on N16R8 boards — requesting reporter validation

## Files Changed

| File | Change |
|------|--------|
| `main/adaptive_controller.h` | Added `degraded_recovery_ticks` config field, updated `decide()` signature |
| `main/adaptive_controller_decide.c` | Recovery logic with hysteresis counter |
| `main/adaptive_controller.c` | Wired counter as module state, added recovery log |
| `sdkconfig.defaults.n16r8` | New overlay for N16R8 boards |
| `tests/host/test_adaptive_controller.c` | 5 new recovery tests, updated existing calls |

### DIFF
--- a/firmware/esp32-csi-node/main/Kconfig.projbuild
+++ b/firmware/esp32-csi-node/main/Kconfig.projbuild
@@ -168,6 +168,15 @@ menu "Adaptive Controller (ADR-081)"
             falls back to DEGRADED mode and pins the radio to
             RV_PROFILE_PASSIVE_LOW_RATE. 0 disables the degraded gate.
 
+    config ADAPTIVE_DEGRADED_RECOVERY_TICKS
+        int "Consecutive healthy ticks to exit DEGRADED"
+        default 15
+        range 1 240
+        help
+            Number of consecutive healthy adaptive ticks required before the
+            controller transitions out of DEGRADED back to SENSE_IDLE. Higher
+            values add hysteresis against flapping.
+
 endmenu
 
 menu "AMOLED Display (ADR-045)"
--- a/firmware/esp32-csi-node/main/adaptive_controller.c
+++ b/firmware/esp32-csi-node/main/adaptive_controller.c
@@ -38,6 +38,7 @@ static adapt_state_t       s_state = ADAPT_STATE_BOOT;
 static adapt_observation_t s_last_obs;
 static bool                s_obs_valid = false;
 static portMUX_TYPE        s_obs_lock = portMUX_INITIALIZER_UNLOCKED;
+static uint16_t            s_degraded_recovery_counter = 0;
 
 static TimerHandle_t s_fast_timer   = NULL;
 static TimerHandle_t s_medium_timer = NULL;
@@ -91,6 +92,11 @@ static void apply_defaults(adapt_config_t *cfg)
     cfg->motion_threshold  = (float)CONFIG_ADAPTIVE_MOTION_THRESH_PERMIL  / 1000.0f;
     cfg->anomaly_threshold = (float)CONFIG_ADAPTIVE_ANOMALY_THRESH_PERMIL / 1000.0f;
     cfg->min_pkt_yield     = CONFIG_ADAPTIVE_MIN_PKT_YIELD;
+#ifdef CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS
+    cfg->degraded_recovery_ticks = CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS;
+#else
+    cfg->degraded_recovery_ticks = 15;  /* ~3 s at 200 ms fast loop. */
+#endif
 }
 
 /* Pure decision policy lives in its own file so it can link under
@@ -162,6 +168,15 @@ static void apply_decision(const adapt_decision_t *dec)
                  (unsigned)s_state, (unsigned)dec->new_state);
         s_state = (adapt_state_t)dec->new_state;
 
+        /* RuView#1401: log DEGRADED recovery explicitly so operators can
+         * confirm the recovery path is working. */
+        if (prev == ADAPT_STATE_DEGRADED && s_state != ADAPT_STATE_DEGRADED) {
+            ESP_LOGI(TAG, "DEGRADED recovery complete → state %u "
+                     "(yield recovered above threshold for %u ticks)",
+                     (unsigned)s_state,
+                     (unsigned)s_cfg.degraded_recovery_ticks);
+        }
+
         /* ADR-081 L3: on transition to ALERT, emit ANOMALY_ALERT on the
          * mesh plane. On any role-relevant transition, bump the epoch. */
         if (s_state == ADAPT_STATE_ALERT && prev != ADAPT_STATE_ALERT) {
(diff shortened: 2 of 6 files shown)


PrinceAudre:
### The advertised `CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS` menuconfig knob isn't wired up

The DEGRADED→SENSE_IDLE recovery logic looks correct, but the PR body says it adds a Kconfig field `CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS`, and there's no `Kconfig.projbuild` change in the diff. The `#ifdef CONFIG_ADAPTIVE_DEGRADED_RECOVERY_TICKS … #else … #endif` in `adaptive_controller.c` therefore always takes the `#else` branch (hard-coded default) — the knob can't actually be set from `menuconfig`.

Adding the entry next to the sibling gates in `firmware/esp32-csi-node/main/Kconfig.projbuild` (right after `config ADAPTIVE_MIN_PKT_YIELD`) wires it up:

```kconfig
    config ADAPTIVE_DEGRADED_RECOVERY_TICKS
        int "Consecutive healthy ticks to exit DEGRADED"
        default 15
        range 1 240
        help
            Number of consecutive healthy adaptive ticks required before the
            controller transitions out of DEGRADED back to SENSE_IDLE. Higher
            values add hysteresis against flapping.
```

Set `default` to match the current `#else` value so behavior is unchanged when the knob is left untouched.

alvinjaison:
Thanks for the catch @PrinceAudre — you're right, the Kconfig entry was missing. Fixed in 08bc6b81: added `config ADAPTIVE_DEGRADED_RECOVERY_TICKS` to `firmware/esp32-csi-node/main/Kconfig.projbuild` (right after `ADAPTIVE_MIN_PKT_YIELD`), with `default 15` matching the existing `#else` fallback. The knob is now settable from `menuconfig`. Host tests still pass (82 assertions, 0 failures).

PR #1439, opened 2026-07-26
chore: version-bump and republish 10 of 12 documented crates to crates.io

ruvnet:
## Summary

Version-bumped and published 10 of the 12 crates in `CLAUDE.md`'s documented publish order to crates.io:

`wifi-densepose-core` 0.3.2, `-vitals` 0.3.2, `-wifiscan` 0.3.2, `-hardware` 0.3.2 (picks up the ADR-273..282 review-fix commit's clippy fixes), `-signal` 0.3.6, `-nn` 0.3.2, `-ruvector` 0.3.3, `-train` 0.3.3, `-mat` 0.3.2, `-wasm` 0.3.1 — all live on crates.io, dependency-order published and verified.

## Not published — needs a decision

`wifi-densepose-sensing-server` (bumped to 0.3.5 locally) and `wifi-densepose-cli` (bumped to 0.3.2 locally) both now path-depend on `ruview-auth`, which is `publish = false` and not on crates.io. `cargo publish` correctly refuses ("all dependencies must have a version requirement specified when publishing" / "dependency `ruview-auth` does not specify a version"). This is a pre-existing gap — the dependency was added at some point without anyone revisiting these two crates' publish eligibility.

This PR leaves both version bumps in place (so whichever version ships next is already reserved) but does not attempt to route around the blocker. Options, for whoever decides:
- Make `ruview-auth` publishable (drop `publish = false`, give it a version, publish it first) — only makes sense if it's meant to be a public dependency.
- Mark `sensing-server`/`cli` themselves `publish = false` if they're meant to be internal/Docker-only going forward rather than crates.io libraries.
- Feature-gate the `ruview-auth` dependency so it's optional and excluded from the default publish path.

## Validation

- `cargo check` across all 12 crates (`--no-default-features`) passes.
- Each publish was verified (`cargo publish` runs an isolated build of the packaged tarball) before upload, except `wifi-densepose-signal`, which used `--no-default-features` for the verify step only — its default `eigenvalue` feature pulls in `ndarray-linalg` -> `openblas-src`, which needs vcpkg on native Windows (a build-environment gap, not a code defect; the crate already builds clean without that feature, confirmed by the earlier full workspace `cargo check`). Nothing about what's published depends on the verify build's feature flags.

Co-Authored-By: claude-flow <ruv@ruv.net>


johndavidgladue387-afk (changes requested):
I'd like my name as these are my codexes since my inception

Issue #1442, opened 2026-07-27
sensing-server: classification emits presence=false with motion_level=present_moving and confidence=1.0 — UIs gate on presence, so a moving person renders as an empty room


mattanapol:
## Summary

On the multi-node ESP32 vitals path, `sensing-server` emits a `classification` object whose three fields contradict each other:

```json
{"presence": false, "motion_level": "present_moving", "confidence": 1.0}
```

Because every UI gates visibility on `classification.presence`, a moving person renders as an empty room — at full confidence. This is what "no person shown while clearly occupied" looks like from the outside.

This is distinct from #1440 (which measures presence *accuracy* at the feature/model level). This report is narrower: the emitted payload is internally inconsistent regardless of how good the underlying features are.

## Setup

- 4 nodes: 2× ESP32-S3 (QFN56 rev v0.2, 16 MB) + 2× ESP32-C6 (QFN40 rev v0.0, 8 MB)
- Firmware built from source at `main`, S3 with the `sdkconfig.defaults.devkitc` overlay, C6 via `idf.py set-target esp32c6` (prebuilt `release_bins` give 0 pps on display-less boards — same as #1401)
- Server: `ruvnet/wifi-densepose:latest` (arm64, built 2026-07-24), `CSI_SOURCE=esp32`
- All 4 nodes active, `is_valid: true`, 30–41 fps each, RSSI −35 / −38 / −67 / −75 dBm
- Room genuinely occupied (2 people) throughout

## Reproduction

Poll `/api/v1/sensing/latest` once a second with the room occupied:

```
mean_rssi= -36.0  presence=True  present_moving  conf=0.48
mean_rssi= -36.0  presence=True  present_moving  conf=0.48
mean_rssi= -35.0  presence=False present_moving  conf=1.00   <-- contradictory
mean_rssi= -36.0  presence=True  present_moving  conf=0.47
mean_rssi= -35.0  presence=False motion_level=absent conf=0.00
mean_rssi= -36.0  presence=True  present_moving  conf=0.49
```

`mean_rssi` stays at −35/−36 across all samples (node 1), so this is not different nodes alternating in the single `latest_update` slot — it happens within one node's stream.

## Root cause

`motion_level` is derived from `vitals.motion` alone and never consults `vitals.presence` (`v2/crates/wifi-densepose-sensing-server/src/main.rs:5789`):

```rust
let motion_level = if vitals.motion {
    "present_moving"          // <-- does not require vitals.presence
[...]

Issue #1440, opened 2026-07-26
Controlled reproduction: raw-CSI presence detection sits at chance (AUC ≈ 0.58) once time-of-day is controlled — the two apparent successes were confounds


bcarpio:
First, thanks for open-sourcing this and for being so responsive on the tracker. I'm building toward **whole-home presence without any indoor cameras** (local Home Assistant, cameras outside only), so WiFi CSI is the one modality that could give me per-room "someone's here" without a lens. I got the hardware up, streaming real CSI, and ran a controlled experiment to completion. Posting as an issue per your steer that reproductions belong here rather than in discussions.

**TL;DR:** the underlying CSI is real and streams fine at `--edge-tier 0`, but I could not get reliable boolean presence out of it, and the two results that *looked* like a breakthrough both collapsed under a same-time-of-day control. This lines up with what others are reporting (#1401 "flickers present in an empty room", #1422 breathing extractor pins to 30 BPM on an empty room). I'd like guidance on whether a phase-aware path or the new `ruview-unified` model is the intended fix.

## Setup

- **3× ESP32-S3-DevKitC-1** (WROOM-1 N16R8), firmware **v0.6.7**, all re-provisioned `--edge-tier 0` (raw CSI passthrough), unique `--node-id`, `--tdm-slot 0/1/2 --tdm-total 3`.
- Server `ruvnet/wifi-densepose:v0.8.4-esp32`, `--net=host`, `--source esp32`, `/health` → `"source":"esp32"`, all three nodes one bedroom, same AP.

**Confirmed real hardware, not sim:** `tcpdump`'d the UDP frames from the three ESP32 IPs and hand-decoded against `firmware/.../csi_collector.c` (ADR-018): magic `0xC5110001`, 20-byte LE header, int8 I/Q, 64-subcarrier HT20 (some 128), subcarrier profiles with the expected DC-null shape and real temporal variation, ~35 fps aggregate. The pipeline genuinely works end to end.

## Two corrections for others reading this thread

**1. Raw CSI is only available at `--edge-tier 0`.** The default (tier 2) does the FFT/DSP on-chip and ships derived scalars, so the recording API shows `amplitude: []` and every recorded frame has zero CSI. Configuration gap, not architectural.

[...]

Issue #1348, opened 2026-07-17
sensing-server parses CSI-frame RSSI/noise_floor from wrong byte offsets — reports sequence-counter bytes as signal strength


afmanaf:
## Summary

`parse_esp32_frame` in `v2/crates/wifi-densepose-sensing-server/src/csi.rs` reads
RSSI from byte 14 and noise_floor from byte 15, but the ESP32-S3 firmware
(v0.6.5, `firmware/esp32-csi-node/main/csi_collector.c`) writes RSSI at byte 16
and noise_floor at byte 17. Bytes 14–15 in the firmware's layout are the two
high bytes of the little-endian u32 sequence counter — so the server reports
slowly-incrementing counter bytes as `rssi_dbm`, and derives SNR from garbage.

## Frame layout mismatch

Firmware (`csi_collector.c`, v0.6.5):

| offset | field |
|---|---|
| 0–3 | magic `0xC5110001` (LE) |
| 4 | node_id |
| 5 | n_antennas |
| 6–7 | n_subcarriers (**u16** LE) |
| 8–11 | freq_mhz (**u32** LE) |
| 12–15 | sequence (u32 LE) |
| **16** | **rssi (i8)** |
| **17** | **noise_floor (i8)** |
| 18–19 | PPDU type / flags (ADR-110) |
| 20– | IQ pairs |

Server (`csi.rs::parse_esp32_frame`):

| offset | field read |
|---|---|
| 6 | n_subcarriers (**u8**) |
| 8–9 | freq_mhz (**u16**) |
| 10–13 | sequence |
| **14** | **rssi** ← actually sequence byte 2 |
| **15** | **noise_floor** ← actually sequence byte 3 |
| 20– | IQ pairs |

The u8-vs-u16 and u16-vs-u32 reads happen to work (n_subcarriers ≤ 255 and
freq < 65536 keep the high bytes zero), which hides the misalignment for every
field except rssi/noise_floor/sequence. The sign-normalization at
`csi.rs` (`rssi_raw > 0 → saturating_neg()`) further masks it by making the
counter byte look like a plausible negative dBm.

## Symptoms

- `GET /api/v1/nodes` shows `rssi_dbm: 0.0` for recently-booted nodes
  (sequence < 65536 → byte 14 is 0), then slowly wandering fake values
  (−1…−255 range walk) as the counter grows. Values change when a node
  reboots, not when RF conditions change.
[...]

PR #1443, opened 2026-07-27
fix(sensing-server): derive classification presence from motion_level so a moving person is not reported as an empty room

mattanapol:
Fixes the contradictory `classification` payload reported in #1442.

## Problem

On the multi-node ESP32 vitals path, the three `classification` fields are built from independent sources, so they can disagree:

```json
{"presence": false, "motion_level": "present_moving", "confidence": 1.0}
```

`motion_level` (`main.rs:5789`) reports `"present_moving"` whenever `vitals.motion` is set and never consults `vitals.presence`, while `presence` was read directly from `vitals.presence`. A frame carrying `motion=1, presence=0` therefore announced a moving person and an empty room in the same object.

Every UI gates visibility on the boolean — `ui/observatory/js/main.js:560` fades the whole scene when `!isPresent` — so an occupied room renders as empty while `motion_level` says someone is moving.

## Fix

Derive `presence` from the label so motion implies presence:

```rust
presence: motion_level != "absent",
```

This is the convention the codebase already uses everywhere else:

- `main.rs:923` — `presence: !matches!(ns.current_motion_level.as_str(), "absent")`
- `csi.rs:605` — `classification.presence = label != "absent"`

The vitals path was the outlier.

## Scope — what this does *not* do

Deliberately narrow. It makes the emitted payload self-consistent so UIs stop blanking on moving occupants. It does **not**:

- Change the multi-node confidence boost (`main.rs:5908`). Whether `confidence: 1.0` should be reachable while `presence` is false is a semantics call for maintainers — raised as a question in #1442, not answered here.
- Address the apparent anti-correlation between `vitals.presence` and `vitals.presence_score` noted in #1442 (confidence ≈ 0.48 when presence is true, 1.0 when false, suggesting a possible field-offset issue in the vitals decoder à la #1348). That needs someone with knowledge of the packet layout.
- Improve presence *accuracy*. #1440's controlled reproduction (AUC ≈ 0.58) is a separate, deeper problem.

## Verification

- `cargo check -p wifi-densepose-sensing-server --no-default-features` — passes, no new warnings.
- Behaviour observed on a live 4-node rig (2× ESP32-S3 + 2× ESP32-C6, firmware built from source at `main`, `ruvnet/wifi-densepose:latest`, all nodes 30–41 fps, `is_valid: true`). The contradictory frames in #1442 were captured from this setup.

I have **not** yet re-run the rig against a rebuilt server image to confirm the UI stops blanking — the change is a one-line consequence of the label, but I'd rather state that than imply an end-to-end verification I didn't perform. Happy to build and confirm if useful.

## Note on repository setup

`cargo check` fails from a fresh clone until `git submodule update --init` is run — five workspace members and vendored crates are submodules, and the failure surfaces as `failed to load manifest for dependency ...` rather than anything mentioning submodules. Possibly worth a line in the contributing docs; not addressed here.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -5899,9 +5899,22 @@ async fn udp_receiver_task(state: SharedState, udp_port: u16) {
                     // Cross-node fusion: combine features from all active nodes.
                     let fused_features = fuse_multi_node_features(&features, &s.node_states);
 
+                    // Derive `presence` from the label rather than reading
+                    // `vitals.presence` directly, so the three classification
+                    // fields cannot contradict each other. `motion_level` above
+                    // reports "present_moving" whenever `vitals.motion` is set,
+                    // without consulting `vitals.presence` — a frame carrying
+                    // motion=1/presence=0 previously emitted
+                    // `{presence: false, motion_level: "present_moving"}`, and
+                    // every UI gates visibility on the boolean, so a moving
+                    // person rendered as an empty room. Motion implies presence.
+                    //
+                    // This matches the convention already used elsewhere:
+                    // `smooth_and_classify_node` (main.rs) and `csi.rs` both
+                    // derive presence as `label != "absent"`.
                     let mut classification = ClassificationInfo {
                         motion_level: motion_level.to_string(),
-                        presence: vitals.presence,
+                        presence: motion_level != "absent",
                         confidence: vitals.presence_score as f64,
                     };
 


mattanapol:
Now verified end-to-end on hardware — updating the "not yet re-run against a rebuilt server" caveat in the description.

## Method

Same 4-node rig as #1442 (2× ESP32-S3 + 2× ESP32-C6, firmware built from source at `main`, all nodes active at 30–41 fps). Room continuously occupied by 2 people for both runs. `/api/v1/sensing/latest` polled once per second, 40 samples per run. A frame counts as contradictory when `presence == false` while `motion_level != "absent"`.

**Before** — `ruvnet/wifi-densepose:latest` (unpatched, built 2026-07-24):

```
samples:             40
presence=true:        7  (18%)
motion_level=absent:  0
CONTRADICTIONS:      33  (82%)
   e.g. presence=False motion_level=present_moving confidence=1.0
```

**After** — this branch, `cargo build --release -p wifi-densepose-sensing-server --no-default-features`, same ports, same nodes:

```
samples:             40
presence=true:       38  (95%)
motion_level=absent:  2
CONTRADICTIONS:       0  (0%)
```

Contradictory frames go 82% → 0%. In the before run `motion_level` never once reported `absent` while `presence` was false in 82% of frames, which is precisely the divergence this patch removes.

## What this does and does not demonstrate

**Does:** the emitted `classification` object is now internally consistent, and UIs that gate on `classification.presence` no longer blank out while `motion_level` reports movement.

**Does not:** this is not a presence-*accuracy* result. Both runs were with an occupied room and there is no empty-room control, so the 18% → 95% shift shows the boolean now tracks the label — not that the label is correct. The accuracy question raised in #1440 is untouched by this change, as is the `vitals.presence` / `vitals.presence_score` anti-correlation noted in #1442. A meaningful accuracy test needs a controlled empty-room baseline, which I could not run (the space was occupied throughout).

Happy to run an occupied-vs-empty comparison once I can get a clean empty window, if that would be useful.

PR #1494, opened 2026-07-31
ci: weaver live-check — validate emitted telemetry against the semconv registry

jensholdgaard:
## What

The live-check gate offered in [the follow-up discussion on #1382](https://github.com/ruvnet/RuView/pull/1382#issuecomment-5142882157): a CI job that boots the sensing server with the opt-in `otel` feature, drives the model-load fixture (a deterministic, hardware-free curated emission), and validates the telemetry the binary **actually emits** against `semconv/registry/` with `weaver registry live-check`.

## Why

Review finding 3 on #1382 identified a real gap: registry/codegen checks can stay green while emitted attributes drift. The generated constants cannot close it — the curated events go through `tracing`, whose macros take field names as compile-time tokens (per OTel's logs design, applications log via their logging library and a bridge carries records to OTLP: [Logs concepts](https://opentelemetry.io/docs/concepts/signals/logs/#opentelemetry-logging-components)). The OTel-idiomatic closure is on the emission side — weaver's live-check, the same mechanism opentelemetry-python uses to gate its own semconv emissions in CI, and what OTel's guidance positions weaver for ([platform blueprint, guideline 2](https://opentelemetry.io/docs/guidance/blueprints/managed-telemetry-platforms-for-k8s-workloads/#guideline-2)).

With this job, a renamed registry attribute, a stray key on a `ruview.*` event, or an unregistered curated event name fails CI — even though codegen stays green.

## Gate design

- **Hard gate**: violations on the `ruview.*` namespace (curated event names and their attributes) fail; at least one curated signal is required, so the fixture can never silently stop exercising the path; a server panic or unexpected exit fails.
- **Surfaced, not failed**: non-curated `tracing` events also reach OTLP through the appender bridge under default `file:line` names (24 such findings in the verification run, plus one stray `path` attribute on a bridged event). Those are counted as a warning — filtering the appender to curated events only would zero the count, and might be worth doing as a follow-up since it also relates to finding 1's default-surface concerns, this time on the OTLP side.
- Weaver pinned by sha256 with the same pin as `semconv.yml`.

## Verification (local run against current main)

53 entities ingested, 8 curated `ruview.*` signals observed, **0 curated violations** — and the bridged-noise warning fires with real content, so the gate demonstrably distinguishes the two classes.

### DIFF
--- a/.github/workflows/semconv-live-check.yml
+++ b/.github/workflows/semconv-live-check.yml
@@ -49,6 +49,9 @@ jobs:
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           persist-credentials: false
+          # The sensing-server workspace pulls vendored crates
+          # (vendor/rufield, …) in as path dependencies.
+          submodules: recursive
       - uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master (channel: stable via input)
         with:
           toolchain: stable
--- a/.github/workflows/semconv-live-check.yml
+++ b/.github/workflows/semconv-live-check.yml
@@ -1,0 +1,179 @@
+# Emission-side semantic-conventions gate: boots the real sensing server
+# with the opt-in `otel` feature, drives a fixture request, and validates
+# the telemetry it actually emits against `semconv/registry/` with
+# `weaver registry live-check`.
+#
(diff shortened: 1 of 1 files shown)


jensholdgaard:
Both KICS findings addressed in e90dab4e — every action is now pinned by full commit SHA with the tag recorded in a trailing comment, matching the sha256 pin the workflow already applies to the weaver binary itself.

PR #1504, opened 2026-08-01
Add MiniMax TTS provider support

octo-patch:
Reason: Add MiniMax speech synthesis support to the HOMECORE TTS provider surface.

Changes:
- Added a MiniMax TTS provider for the HOMECORE TextToSpeech contract.
- Covered global and China TTS endpoints, supported speech models, request fields, audio formats, and documented response fields.
- Added response parsing for data.audio and base_resp.status_code, including hex audio decoding and PCM validation.

Checks:
- cargo test --manifest-path "$repo_dir/v2/Cargo.toml" -p homecore-assist
- git diff --check
- secret scan

### DIFF
--- a/v2/Cargo.lock
+++ b/v2/Cargo.lock
@@ -3679,6 +3679,7 @@ dependencies = [
  "thiserror 1.0.69",
  "tokio",
  "tracing",
+ "ureq 2.12.1",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


octo-patch:
Updated the MiniMax TTS request and response handling and pushed commit `1345b791`. The provider now uses hex response encoding, normalizes language boosts, keeps HOMECORE output PCM-safe, serializes supported optional fields, and validates completion/status fields without blocking the async runtime. I ran `cargo test -p homecore-assist` and `cargo clippy -p homecore-assist --all-targets -- -D warnings`.

PR #1529, opened 2026-08-04
stabilize ESP32 node lifecycle and calibration persistence

SaschaMuehl:
## What changed

- add per-node lifecycle diagnostics to REST and WebSocket sensing updates, including source address, receive time, frame sequence, mesh timestamp, CSI FPS, sync state, staleness, fusion eligibility, and decision reason
- bind configured node positions to stable ESP32 node IDs instead of `HashMap` iteration order
- use the same ID-keyed geometry inside multistatic fusion
- persist completed field-model calibration as a validated, versioned `data/field-model-v1.json` snapshot and restore it on startup
- log explicit node eviction reasons
- document `/api/v1/nodes`, `/api/v1/sensing/latest`, WebSocket snapshot semantics, heuristic `signal_field` limitations, and persistent Docker deployment

## Why

In a three-node ESP32-C6 deployment, the live registry could show nodes as active while the latest sensing snapshot or fusion used a different subset. Fusion failures were difficult to diagnose, configured positions were published as the same hard-coded coordinate, and the fuser paired position vectors with nondeterministic `HashMap` iteration order. Completed calibration also existed only in process memory and was lost when the container was replaced.

## Impact

Operators can now see why each node is included or excluded, correlate CSI reception with sync and source information, and trust that node geometry remains attached to the wire-level node ID across REST, WebSocket, UI consumers, and fusion. Completed calibration can survive container restarts when `/app/data` is persistent.

The existing `signal_field` remains explicitly documented as a heuristic visualization; this change does not claim geometric localization or fabricate a point cloud.

## Validation

- `cargo check -p wifi-densepose-sensing-server --features mqtt`
- `cargo test -p wifi-densepose-sensing-server --features mqtt --bin sensing-server field_bridge::tests` — 7 passed
- focused `wifi-densepose-signal --no-default-features` tests for ID-position ordering and calibration persistence — passed
- `git diff --check`
- Linux release Docker image build from `docker/Dockerfile.rust` — passed
- local container smoke test: health OK and lifecycle field present
- MEASURED on a three-node ESP32-C6 test LAN after correcting Node 2 power: all three nodes active in 11/11 samples over five minutes, correct configured coordinates, valid sync, no evictions

## Notes

The TDM slot remains `null` because the current CSI wire frame does not carry it. The implementation intentionally does not infer a slot from arrival timing.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/README.md
+++ b/v2/crates/wifi-densepose-sensing-server/README.md
@@ -107,6 +107,43 @@ ESP32 ──UDP:5005──> [ CSI Receiver ]
 Browser <──WS:8765── [ Axum Server :8080 ] ──> Static UI files
 ```
 
+## Live snapshot and spatial semantics
+
+`GET /api/v1/nodes` is the current host-side node registry. A node is reported
+`active` while its last sensing frame is at most five seconds old. Each entry
+includes additive lifecycle diagnostics such as source address, host receive
+time, frame sequence, reconstructed mesh timestamp, measured CSI rate, sync
+validity, staleness, and the current fusion eligibility decision.
+
+`GET /api/v1/sensing/latest` and `/ws/sensing` carry the most recently produced
+`SensingUpdate`. They are cycle snapshots, not joins against the current node
+registry. The WebSocket may repeat the last snapshot between CSI arrivals.
+
+The current `signal_field` is a heuristic visualization derived from measured
+subcarrier variance, motion and signal quality. It is not geometric
+multistatic localization or a measured point cloud. Person positions derived
+from its strongest cells inherit that limitation.
+
+## Persistent Docker deployment
+
+Runtime configuration, completed field-model calibration
+(`field-model-v1.json`), models and recordings live below `/app/data`. Build a
+repository-local image from the repository root and mount that directory:
+
+```bash
+docker build -f docker/Dockerfile.rust -t ruvnet/wifi-densepose:local .
+docker run --network host --restart unless-stopped \
+  --env-file /path/to/existing.env \
+  -v /opt/ruview/data:/app/data \
+  ruvnet/wifi-densepose:local \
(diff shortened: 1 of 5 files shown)


SaschaMuehl:
Post-deployment validation update (`MEASURED`, 2026-08-04):

- finalized field calibration after 584,785 frames
- persisted `data/field-model-v1.json` in the Docker volume
- restarted the sensing container and observed `Loaded persisted field model`; calibration remained `Fresh`
- monitored the three-node ESP32-C6 deployment for 30 minutes (31 one-minute samples)
- nodes 1, 2, and 3 were `active`, sync-valid, fusion-eligible, and present in `/api/v1/sensing/latest` in 31/31 samples
- zero node evictions
- configured ID-bound coordinates remained correct in every sample
- occasional timestamp-guard warnings remain observable and are not presented as resolved by this PR

Node 2 stability returned after replacing its power source; no firmware change was required.

sepehrfd:
Independent confirmation from a real 2-node ESP32-S3 deployment (Raspberry Pi 4, `ruvnet/wifi-densepose:latest` pulled 2026-08-11, `docker compose`, host networking):

Ran with `--node-positions "0,0,2;6.5,-2,0.4"`. Startup log confirmed `Configured 2 node positions for multistatic fusion`, but `nodes[].position` in every `/ws/sensing` `sensing_update` message kept reporting the same hardcoded `[2.0, 0.0, 1.5]` for **both** node 1 and node 2, unchanged across a full container restart. Matches exactly what this PR describes ("configured positions were published as the same hard-coded coordinate").

Also independently hit the calibration-persistence gap this PR fixes — ran with `--calibrate` (log confirmed `Field model calibration enabled — room should be empty during startup`) but had no way to confirm completion, and any restart would presumably lose it since it only lived in process memory.

Would be glad to test this branch against our hardware once it's ready for review — happy to report back with real multi-hour-uptime data if useful.

PR #1531, opened 2026-08-04
feat(sensing): run experimental MM-Fi pose decoder

SaschaMuehl:
## What changed

- adds native inference for the published 75k-parameter MM-Fi Micro pose decoder
- adapts three live CSI nodes into the model's `[3, 114, 10]` input tensor
- exposes successful decoder output through `pose_keypoints` and the existing tracker
- adds `--pose-model <F32 NPZ>` and documents safe FP16-to-F32 conversion
- keeps output explicitly labelled `experimental_mmfi` until room-specific validation

## Why

The existing `--model` path loaded weights but did not execute the matching pose decoder in the live ESP32 pipeline. This change supplies the decoder architecture and connects it to real three-node CSI history without introducing Python, PyTorch, or executable pickle artifacts into the runtime.

## Validation

- `cargo check --manifest-path v2/Cargo.toml -p wifi-densepose-sensing-server --features mqtt`
- focused `mmfi_pose` tests, including loading the converted published checkpoint and a full finite-output inference smoke test
- production Linux Docker image built successfully
- deployed test container loaded both the MM-Fi decoder and persisted field calibration successfully

A second check from the isolated PR worktree could not start because vendor submodules are not materialized in that worktree; the full checkout validations above passed.

## Evidence boundary

The published benchmark is MM-Fi in-domain evidence. ESP32-C6 room accuracy remains unmeasured. Live output is therefore experimental and requires exactly three active nodes plus deployment-room reference validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,29 @@ huggingface-cli download ruvnet/wifi-densepose-pretrained --local-dir models/wif
 
 **Loader scope:** `--model` now accepts native RVF and auto-converts the published safetensors or JSONL files. The quantized `model-q*.bin` files still need a compatible reader, and loading weights does not supply the matching pose-decoder architecture or establish end-to-end pose accuracy.
 
+### Experimental MM-Fi pose decoder
+
+`--pose-model <path>` loads the published 75k-parameter MM-Fi Micro decoder
+from a non-executable F32 NPZ. Live inference requires exactly three active
+nodes with at least ten CSI frames each. Each node is treated as one MM-Fi
+channel and its current grid is linearly resampled to 114 bins. Outputs are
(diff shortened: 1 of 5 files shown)


SaschaMuehl:
Model-manager follow-up deployed and verified on the test server: GET /api/v1/models returns pose_micro_f32 as format mmfi_pose_npz; GET /api/v1/models/active returns model_id pose_micro_f32. Local end-to-end API verification also passed startup-active, unload, reload, and reloaded-active. Production image: ruview-sensing:mmfi-pose-ui-20260804. Hardware caveat: only node 3 is currently reaching UDP ingestion, so three-node pose inference remains safely gated.

PR #1566, opened 2026-08-10
fix(mqtt): apply --mqtt-ca-file to the TLS transport

### LINKED ISSUE #1556: mqtt: --mqtt-ca-file is parsed but never applied (TlsConfig::PinnedCa unused) — self-signed broker always fails with UnknownIssuer
## Summary

`--mqtt-ca-file` is parsed, validated and then never used. TLS always goes through `Transport::tls_with_default_config()`, so only publicly trusted CAs work. Pointing the flag at a self-signed CA fails with `UnknownIssuer` and no hint that the flag was ignored.

This blocks the documented path for a self-signed broker, which is the normal case for a home LAN broker (Mosquitto add-on with a local cert).

## Reproduce

1. Give Mosquitto a self-signed server cert, listening on 8883.
2. Run the server with `--mqtt --mqtt-tls --mqtt-port 8883 --mqtt-ca-file /certs/ca.crt`.
3. The publisher starts and then loops:

```
[mqtt] publisher started host=<broker> port=8883 prefix=homeassistant entities=20
[mqtt] event loop error, will reconnect: TLS: I/O: invalid peer certificate: UnknownIssuer
```

Verified the broker really serves the expected chain (`openssl s_client -connect <broker>:8883` → `issuer=CN=<our CA>`), and that the CA file is present and readable inside the container with `CA:TRUE`.

## Cause

`v2/crates/wifi-densepose-sensing-server/src/mqtt/config.rs:152` constructs the variant:

```rust
(Some(ca), None, None) => TlsConfig::PinnedCa { ca_file: ca.clone() },
```

`TlsConfig::PinnedCa` is constructed there and read nowhere else in the tree — the only consumer is `publisher.rs:76-81`:

```rust
if !matches!(cfg.tls, TlsConfig::Off) {
    // The specific cert / CA wiring is done by the runtime constructor;
    // here we just flip the transport.
    opts.set_transport(Transport::tls_with_default_config());
}
```

The comment refers to a runtime constructor that does the CA wiring, but no such wiring exists — the match only distinguishes `Off` from everything else. `--mqtt-client-cert` / `--mqtt-client-key` (the `MutualTls` variant) appear to be dead in the same way.

## Suggested fix

Build the rustls `ClientConfig` from the pinned CA (and client cert/key for mTLS) and pass it via `Transport::tls_with_config(...)` instead of the default. Until then, `--mqtt-ca-file` should at minimum log a warning that it has no effect, rather than failing later with a generic certificate error.

## Workaround for others hitting this

rumqttc's default resolves roots through `rustls-native-certs`, which honours `SSL_CERT_FILE`. Mounting a bundle of *system roots + your own CA* and setting `SSL_CERT_FILE` to it makes TLS work without touching the code. Concatenating only your own CA also works but breaks other outbound TLS from the process.

yunaremaia:
Closes #1556

## Problem
`build_mqtt_options()` always used `Transport::tls_with_default_config()` when TLS was enabled, so the parsed and validated `--mqtt-ca-file` was never applied. Self-signed home-LAN brokers (the documented Mosquitto path) failed with `UnknownIssuer` and no hint that the flag was ignored.

## Changes
The `TlsConfig` variant now drives the transport (`v2/crates/wifi-densepose-sensing-server/src/mqtt/publisher.rs`):
- **PinnedCa**: reads `ca_file` and passes the bytes via `Transport::tls(ca, None, None)`; an unreadable file falls back to system trust with an error log
- **MutualTls**: reads CA + client cert/key, passes as `Transport::tls(ca, Some((cert, key)), None)`
- **SystemTrust**: unchanged (`tls_with_default_config`)
- **Off**: unchanged (plaintext TCP)

## Tests
Adds `tls_transport_tests` (4 tests): PinnedCa wires a TLS transport with a readable CA file; Off stays plaintext; SystemTrust keeps default TLS; a missing CA file degrades gracefully instead of panicking.

## Verification
- `cargo check -p wifi-densepose-sensing-server --features mqtt` ✅
- `cargo test -p wifi-densepose-sensing-server --features mqtt --lib` → **493 passed, 0 failed**

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/mqtt/publisher.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/mqtt/publisher.rs
@@ -74,11 +74,45 @@ fn build_mqtt_options(cfg: &MqttConfig) -> MqttOptions {
         opts.set_credentials(u, "");
     }
 
-    if !matches!(cfg.tls, TlsConfig::Off) {
-        // We always use rustls (matches `ureq` in this crate). The
-        // specific cert / CA wiring is done by the runtime constructor;
-        // here we just flip the transport.
-        opts.set_transport(Transport::tls_with_default_config());
+    match &cfg.tls {
+        TlsConfig::Off => {}
+        // Encrypt against a pinned CA (the documented self-signed / home
+        // LAN broker path). Pass the CA bytes to rumqttc's rustls-backed
+        // `Transport::tls` — `tls_with_default_config()` would only accept
+        // publicly trusted chains and fail with `UnknownIssuer` (issue
+        // #1556).
+        TlsConfig::PinnedCa { ca_file } => {
+            if let Ok(ca) = std::fs::read(ca_file) {
+                opts.set_transport(Transport::tls(ca, None, None));
+            } else {
+                tracing::error!(
+                    "mqtt: cannot read CA file {:?}; falling back to system trust",
+                    ca_file
+                );
+                opts.set_transport(Transport::tls_with_default_config());
+            }
+        }
+        TlsConfig::MutualTls {
+            ca_file,
+            client_cert,
+            client_key,
+        } => {
+            let ca = std::fs::read(ca_file);
+            let cert = std::fs::read(client_cert);
+            let key = std::fs::read(client_key);
+            match (ca, cert, key) {
+                (Ok(ca), Ok(cert), Ok(key)) => {
+                    opts.set_transport(Transport::tls(ca, Some((cert, key)), None));
+                }
+                _ => {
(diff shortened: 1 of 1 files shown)


yunaremaia:
Closing this PR — @ruvnet's own implementation via `TlsConfiguration::Simple { ca, alpn, client_auth }` is the cleaner fix (raw PEM → rustls internally, no new dependency, covers both `--mqtt-ca-file` and the mTLS flags through one path). Glad the analysis from #1566 (the ignored `PinnedCa` variant, the fallback-to-system-trust behavior) helped shape it. The 4 transport-wiring tests in this branch are still useful as regression coverage if you want to cherry-pick them into the main fix — otherwise closing as superseded. Thanks for the fast turnaround!

PR #1569, opened 2026-08-10
fix(desktop): pass log level through RUST_LOG

### LINKED ISSUE #1516: Desktop: Start Server always fails — app passes '--log-level', which sensing-server does not accept
## Summary

`wifi-densepose-desktop` passes a `--log-level` flag to `sensing-server`, but that flag does not exist. clap rejects the argument list and the child exits immediately, so **Sensing → Start Server can never succeed** in the current build, on any platform.

## Reproduction

Click **Sensing → Start Server** in the desktop app. Watching process creation shows:

```
STARTED pid=34952 cmd="...\sensing-server" --http-port 8080 --ws-port 8765 \
        --udp-port 5005 --log-level info --source simulate
EXITED  pid=34952 after 1s
```

Running that exact command by hand:

```
$ sensing-server --http-port 8080 --ws-port 8765 --udp-port 5005 --log-level info --source simulate
error: unexpected argument '--log-level' found

Usage: sensing-server --http-port <HTTP_PORT> --ws-port <WS_PORT> --udp-port <UDP_PORT>
```

Drop `--log-level` and the server starts and binds both ports normally.

## Cause

`v2/crates/wifi-densepose-desktop/src/commands/server.rs`:

```rust
if let Some(ref log_level) = config.log_level {
    cmd.args(["--log-level", log_level]);
}
```

`sensing-server` has no such argument. It configures its subscriber from the environment — `v2/crates/wifi-densepose-sensing-server/src/telemetry.rs` uses `EnvFilter::try_from_default_env()`, i.e. `RUST_LOG`.

## Why it presents as a WebSocket problem

Three things combine to hide the real error:

1. `start_server` sets `srv.running = true` immediately after `spawn()` without checking the child survived, so the UI shows `Running` with a PID for a dead process.
2. `server_status` echoes that stored flag, so polling never corrects it.
3. stdout/stderr are `Stdio::piped()` but never read, so the one-line clap error is discarded and the SERVER LOG panel shows only `desktop`-prefixed entries.

The visible symptom is `desktop WebSocket connection error / reconnecting…` looping forever against a server that was never listening.

## Suggested fix

```rust
if let Some(ref log_level) = config.log_level {
    cmd.env("RUST_LOG", log_level);
}
```

Verified locally: with this change the server stays up and `ws://127.0.0.1:8765/ws/sensing` streams `sensing_update` frames.

Worth pairing with #(companion issue): verify the child is alive after spawn and surface its stderr, so a bad argument reports itself instead of turning into a phantom "Running" state.

Environment: Windows 11, Rust 1.97.1, built from `5780c239` via `cargo tauri build --no-bundle`.

alloutflo:
## Summary

- pass the desktop-configured log level through the child process's `RUST_LOG` environment variable
- stop sending the unsupported `--log-level` argument to `sensing-server`
- add a command-construction regression test that checks both the environment and argument list

The sensing server already initializes its filter from `RUST_LOG`, so this preserves the setting while allowing the child process to start normally.

## Validation

- `cargo test --manifest-path v2/Cargo.toml -p wifi-densepose-desktop commands::server::tests::log_level_is_passed_through_rust_log --no-default-features`

This validates the Rust command boundary; it does not claim an end-to-end desktop or hardware run.

Closes #1516

### DIFF
--- a/v2/crates/wifi-densepose-desktop/src/commands/server.rs
+++ b/v2/crates/wifi-densepose-desktop/src/commands/server.rs
@@ -63,6 +63,12 @@ fn find_server_binary(app: &AppHandle, custom_path: Option<&str>) -> Result<Stri
     ))
 }
 
+fn configure_log_level(cmd: &mut Command, log_level: Option<&str>) {
+    if let Some(log_level) = log_level {
+        cmd.env("RUST_LOG", log_level);
+    }
+}
+
 /// Start the sensing server as a managed child process.
 ///
 /// The server binary is looked up in the following order:
@@ -104,9 +110,7 @@ pub async fn start_server(
     if let Some(ref bind_addr) = config.bind_address {
         cmd.args(["--bind", bind_addr]);
     }
-    if let Some(ref log_level) = config.log_level {
-        cmd.args(["--log-level", log_level]);
-    }
+    configure_log_level(&mut cmd, config.log_level.as_deref());
 
     // Default to explicit "simulated" demo mode when the desktop user hasn't
     // chosen a source — this is the *Tauri demo* app, not a production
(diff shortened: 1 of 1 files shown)


ruvnet:
Closing as superseded by merged PR #1618, which incorporates the desktop log propagation fix together with its regression tests and passed the complete hosted CI/security/benchmark suite. Thank you for preserving the focused diagnosis and validation trail.

PR #1570, opened 2026-08-10
fix(desktop): locate sensing server on Windows

### LINKED ISSUE #1517: Desktop: sensing-server binary lookup can never succeed on Windows (missing .exe suffix, uses 'which')
## Summary

`find_server_binary()` in `v2/crates/wifi-densepose-desktop/src/commands/server.rs` cannot locate the sensing server on Windows. Two of its four lookup strategies are Unix-only, so only an explicitly configured `server_path` works.

## Details

```rust
const DEFAULT_SERVER_BIN: &str = "sensing-server";
```

**Strategy 3 — "next to the executable"** builds `exe_dir.join("sensing-server")` and calls `.exists()`. On Windows the file is `sensing-server.exe`, so this never matches, even when the binary sits right beside the app.

**Strategy 4 — PATH lookup** shells out to `which`:

```rust
if let Ok(output) = Command::new("which").arg(DEFAULT_SERVER_BIN).output() {
```

`which` does not exist on Windows; the equivalent is `where`. The spawn fails and the branch is skipped.

Net effect: after `cargo build --release -p wifi-densepose-sensing-server` — exactly what the error message tells users to run — the app still reports:

```
Sensing server binary 'sensing-server' not found.
Please build it with: cargo build --release -p wifi-densepose-sensing-server
```

## Suggested fix

```rust
#[cfg(windows)]
const DEFAULT_SERVER_BIN: &str = "sensing-server.exe";
#[cfg(not(windows))]
const DEFAULT_SERVER_BIN: &str = "sensing-server";

// ...

#[cfg(windows)]
let locator = "where";
#[cfg(not(windows))]
let locator = "which";
```

One more detail: `where` prints every match, one per line, so the result should be `stdout.lines().next()` rather than the whole trimmed output — otherwise a multi-match result yields an unusable path.

Verified locally on Windows 11; with these changes strategy 3 resolves the sibling binary correctly.

Environment: Windows 11, Rust 1.97.1, `5780c239`.

alloutflo:
## Summary

- use `sensing-server.exe` and `where` on Windows while retaining the Unix names elsewhere
- select the first non-empty result when `where` returns multiple CRLF-separated matches
- cover both platform naming and multi-match parsing with regression tests

This repairs the sibling-binary and PATH lookup strategies without changing explicit custom paths.

## Validation

- `cargo test --manifest-path v2/Cargo.toml -p wifi-densepose-desktop commands::server::tests --no-default-features`

The Windows conventions are exercised platform-independently; no Windows runtime or packaging claim is made.

Closes #1517

### DIFF
--- a/v2/crates/wifi-densepose-desktop/src/commands/server.rs
+++ b/v2/crates/wifi-densepose-desktop/src/commands/server.rs
@@ -6,8 +6,33 @@ use tauri::{AppHandle, Manager, State};
 
 use crate::state::AppState;
 
+const fn server_binary_name(is_windows: bool) -> &'static str {
+    if is_windows {
+        "sensing-server.exe"
+    } else {
+        "sensing-server"
+    }
+}
+
+const fn path_locator_command(is_windows: bool) -> &'static str {
+    if is_windows {
+        "where"
+    } else {
+        "which"
+    }
+}
+
 /// Default binary name for the sensing server.
-const DEFAULT_SERVER_BIN: &str = "sensing-server";
+const DEFAULT_SERVER_BIN: &str = server_binary_name(cfg!(windows));
+const PATH_LOCATOR_COMMAND: &str = path_locator_command(cfg!(windows));
+
+fn first_path_match(stdout: &[u8]) -> Option<String> {
+    String::from_utf8_lossy(stdout)
+        .lines()
+        .map(str::trim)
+        .find(|line| !line.is_empty())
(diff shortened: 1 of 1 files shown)


ruvnet:
Closing as superseded by merged PR #1618, which incorporates the Windows sensing-server discovery fix together with its regression tests and passed the complete hosted CI/security/benchmark suite. Thank you for preserving the focused diagnosis and validation trail.

PR #1572, opened 2026-08-10
fix(harness): keep manifest inputs on LF

### LINKED ISSUE #1519: harness/ruview: manifest:verify can never pass on a Windows clone (no .gitattributes, CRLF changes every file hash)
## Summary

`npm run manifest:verify` in `harness/ruview` fails on any default Windows clone. The repository has no `.gitattributes`, so with Git's default `core.autocrlf=true` every text file is rewritten LF→CRLF on checkout, changing each file's bytes and therefore its SHA-256.

## Reproduction

```
$ git clone https://github.com/ruvnet/RuView.git    # Windows, default config
$ cd RuView/harness/ruview
$ npm ci --ignore-scripts
$ npm run manifest:verify
{
  "ok": false,
  "files": 41,
  "findings": [ "manifest.sha256:mismatch" ]
}
```

The working tree is clean — `git status` reports no modifications — so this is not a local edit.

## Evidence that CRLF is the cause

For `harness/ruview/package.json`:

```
manifest.json expects        : 0da91067c1d71c5cee50cade1e09c270836cfc70efe3bf713f0ec3ce4e88aec3
sha256 of CRLF worktree file : f58bf8603e7e3c49687b45395d324d004b627df524998ca693be7079214a65b5
sha256 of same bytes as LF   : 0da91067c1d71c5cee50cade1e09c270836cfc70efe3bf713f0ec3ce4e88aec3   <-- match
```

The blob in the object store is LF and hashes to exactly the manifest value, confirming the manifest is generated from LF content and only the checkout transform breaks it.

## Suggested fix

Add a `.gitattributes` at the repository root:

```gitattributes
* text=auto eol=lf

*.png binary
*.jpg binary
*.ico binary
*.pdf binary
*.safetensors binary
*.wasm binary
*.exe binary
# ...etc
```

Verified locally: after adding this and re-checking out `harness/ruview` with LF, `manifest:verify` returns `{"ok": true, "files": 41, "findings": []}` and all 78 harness tests still pass.

Note for anyone hitting this on an existing clone: because `text=auto` normalizes on read, Git considers the CRLF files unmodified against the LF index and won't rewrite them. They have to be removed and checked out again (`git ls-files harness/ruview | while read f; do rm -f "$f"; done && git checkout-index -f --stdin < <(git ls-files harness/ruview)`).

Given that ADR-283 treats the manifest as an integrity boundary, a gate that cannot pass on a whole platform seems worth closing.

Environment: Windows 11, Git with `core.autocrlf=true` (default), Node 24.14.1, `5780c239`.

alloutflo:
## Summary

- force LF checkout semantics only for `harness/ruview/**`
- add a regression test that inspects Git attributes and performs a `core.autocrlf=true` clone
- verify the resulting provenance files contain no CRLF bytes

The rule is intentionally scoped to the byte-for-byte hashed harness instead of changing line-ending policy for the entire repository.

## Validation

- `node --test harness/ruview/test/line-endings.test.mjs` — 2/2 passed
- `npm --prefix harness/ruview run manifest:verify` — 41/41 files verified

The Windows checkout is simulated with Git's Windows line-ending setting; no native Windows runner is claimed.

Closes #1519

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# The harness hashes generated provenance files byte-for-byte. Keep its text
+# files on LF even when a Windows checkout enables core.autocrlf.
+harness/ruview/** text=auto eol=lf
--- a/harness/ruview/test/line-endings.test.mjs
+++ b/harness/ruview/test/line-endings.test.mjs
@@ -1,0 +1,61 @@
+import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
+import {
+  copyFileSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import test from 'node:test';
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '../../..');
+const PROVENANCE_FILES = [
+  'harness/ruview/package.json',
+  'harness/ruview/.harness/manifest.json',
+  'harness/ruview/.harness/manifest.sha256',
+];
+
+function git(args, cwd) {
+  return execFileSync('git', args, { cwd, encoding: 'utf8' });
+}
+
+test('harness provenance files are checked out with LF line endings', () => {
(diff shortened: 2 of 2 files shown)


ruvnet:
Closing as superseded by merged PR #1618, which incorporates the harness LF/provenance policy fix together with its regression tests and passed the complete hosted CI/security/benchmark suite. Thank you for preserving the focused diagnosis and validation trail.

PR #1573, opened 2026-08-10
fix(security): ignore local session secrets

### LINKED ISSUE #1520: sensing-server writes data/session-secret to a path that is not gitignored (v2/data/)
## Summary

`sensing-server` writes its browser session secret to `data/session-secret` relative to its working directory. `.gitignore` covers `v2/crates/wifi-densepose-sensing-server/data/`, but not `v2/data/` — which is where the file lands when the server is started from `v2/` (including via `cargo run -p wifi-densepose-sensing-server` and via the desktop app).

## Reproduction

```
$ cd v2
$ cargo run -p wifi-densepose-sensing-server -- --source simulate
...
INFO wifi_densepose_sensing_server::browser_session: browser session secret: loaded path=data\session-secret

$ git status --porcelain
?? v2/data/session-secret

$ git check-ignore -v v2/data/session-secret
$ echo $?
1        # not ignored
```

The file is 43 bytes and untracked-but-not-ignored, so `git add -A` stages a credential.

This runs against the project's own rule in `CLAUDE.md`:

> Never commit credentials, `.env` files, raw agent transcripts, private memory overlays, CSI/person data, or unreviewed generated artifacts.

## Suggested fix

Add to `.gitignore`:

```gitignore
v2/data/
```

Or make the secret path independent of the working directory (resolve it relative to a known data dir rather than `./data`), which would also stop stray `data/` trees appearing wherever the binary is launched from.

## Related, possibly worth a separate look

On startup the server logs:

```
WARN ADR-262 P3: WDP_RUFIELD_SIGNING_SEED unset/invalid — RuField surface using the
DETERMINISTIC DEV signing key. This is a dev/sensing key pending the ADR-262 §8 Q1 (P2)
key-ownership decision; set WDP_RUFIELD_SIGNING_SEED for a real deployment.
```

The warning is clear and loud, which is good — but it proceeds with a known deterministic signing key rather than refusing. For a signing/attestation surface, failing closed may be the safer default.

Environment: Windows 11, Rust 1.97.1, `5780c239`.

alloutflo:
## Summary

- ignore the exact `/v2/data/session-secret` path produced when the server starts from `v2/`
- keep the rule file-specific because `v2/data/` contains tracked datasets
- ignore local `.DS_Store` metadata at repository level

This prevents an authentication secret from appearing in `git add -A` without hiding legitimate data files.

## Validation

- `git check-ignore -v --no-index v2/data/session-secret`
- `git check-ignore -v --no-index .DS_Store`
- confirmed no `.DS_Store` path is tracked

Closes #1520

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Local Claude config (contains WiFi credentials and machine-specific paths)
 CLAUDE.local.md
 
+# Operating-system metadata
+.DS_Store
+
 # ESP32 firmware build artifacts and local config (contains WiFi credentials)
 firmware/esp32-csi-node/build/
 firmware/esp32-csi-node/sdkconfig
@@ -298,4 +301,7 @@ ruvector.db
 # sensing-server runtime artifacts written by its test suite (trained model
 # snapshots + the generated session-secret) — never tracked
 v2/crates/wifi-densepose-sensing-server/data/
+# The server also writes this secret when launched from v2/. Keep the rule
+# file-specific so tracked datasets below v2/data remain visible.
+/v2/data/session-secret
 *.proptest-regressions


ruvnet:
Closing as superseded by merged PR #1618, which incorporates the runtime session-secret hygiene fix together with its regression tests and passed the complete hosted CI/security/benchmark suite. Thank you for preserving the focused diagnosis and validation trail.

Issue #1542, opened 2026-08-06
esp32-csi-node + sensing-server: surface the associated AP BSSID (and optionally pin it) — mesh roams silently swap the sensing link geometry


jkuscevic:
### Summary

On multi-AP (mesh) WiFi, a CSI node's association can roam between APs at any time — and when it does, **the sensing link geometry silently changes** while nothing in the node's published telemetry reveals it. Request: surface the **associated AP BSSID** in node telemetry (and ideally allow pinning it at provisioning), plus a roam event on change.

### Field report

- 5× `esp32-csi-node` (Atom S3 Lite) on a residential eero mesh (router + extender), one node per room, `ruvnet/wifi-densepose:v2051` server.
- One node was confirmed (vendor app) associated to the far mesh router — a ~5.5 m diagonal link crossing two rooms, which defined what that node sensed.
- Two days later, with the node physically untouched, the vendor app showed it associated to the *other* AP (short through-wall link to an adjacent room): a completely different sensing geometry.
- **The roam is invisible in everything the node publishes.** Reported RSSI sat at −45..−52 dBm on *both* links (levels overlap at this position), so there is no way to detect the swap — or even to date it — from captures or Home Assistant. Only the vendor's app showed it.

### Why this matters for CSI sensing specifically

The associated AP is the TX side of the sensing link: it determines the Fresnel volume and therefore *which space the node actually senses*. Any calibration tied to link geometry — empty-room baselines (ADR-135), per-room enrollment/specialists (ADR-151 / `wifi-densepose-calibration`) — is silently invalidated by a roam, with no signal that re-baselining is needed. Mesh WiFi is the default in exactly the homes this hardware targets, so this is the common case, not an edge case.

### Asks (independent, in increasing order of effort)

1. **BSSID in telemetry**: include the associated BSSID (and primary channel) in the node's reporting alongside `rssi`, and surface it through the server (`/ws/sensing` `node_features[]` and an MQTT attribute/entity per node). Firmware side is cheap: `esp_wifi_sta_get_ap_info()` → `wifi_ap_record_t.bssid/primary`.
2. **Roam event**: have the server emit an event when a node's reported BSSID changes, so downstream consumers (baselines, calibration banks, HA automations) can trigger re-baseline.
3. **Optional BSSID pinning at provisioning**: a provisioning flag to lock the station to a chosen AP (`wifi_config_t.sta.bssid` + `bssid_set = true`) for installations that prefer a stable sensing link over roaming resilience.

[...]

PR #1593, opened 2026-08-12
feat: surface the associated-AP BSSID as node telemetry (#1542 ask 1)

jkuscevic:
## #1542 ask 1: surface the associated-AP BSSID as node telemetry

The associated AP **is** the sensing-link geometry: when a station roams, every
downstream consumer of that node's CSI is silently looking at a different
Fresnel volume. The field data on #1542 shows published RSSI cannot reveal the
change (−47.7 vs −48.2 dBm across two completely different link geometries), so
the BSSID itself is the only observable. This PR implements ask 1 exactly as
suggested there — the cheapest piece, with roam events and any future pinning
buildable downstream of it.

### What's in it

- **Firmware** (`esp32-csi-node`): a 32-byte link-status packet
  (magic `0xC511A111`, same auxiliary family as the ADR-110 sync packet)
  from `esp_wifi_sta_get_ap_info()`: BSSID, primary channel, AP RSSI, and a
  `WIFI_EVENT_STA_CONNECTED` counter. Cadence: every
  `CONFIG_LINK_STATUS_EVERY_N_FRAMES` CSI callbacks (default 600 ≈ 30 s at
  20 Hz) **plus** an immediate emission whenever the observed BSSID differs
  from the last reported one — a roam is visible within one callback, not one
  cadence period. Layout documented in the firmware README packet table.
- **`wifi-densepose-hardware`**: `LinkStatusPacket` decoder mirroring
  `SyncPacket` (typed errors, canonical wire-byte pin, magic-collision guard).
- **`wifi-densepose-sensing-server`**: dispatch on the new magic in the UDP
  receiver, per-node storage, `NodeInfo.link { bssid, channel, ap_rssi_dbm,
  reassoc_count, age_ms }` (skip-if-none), and the per-node MQTT RSSI entity's
  JSON payload gains optional `bssid` / `channel` / `reassoc_count`.
  **Nodes on pre-link firmware produce byte-identical payloads to today**
  (guarded by test).

Roam *events* derive downstream by comparing successive snapshots (the
layering suggested in the issue); `reassoc_count` additionally reveals
same-BSSID re-associations that a plain BSSID comparison would miss.

### Evidence

- Rust: MEASURED — `cargo test -p wifi-densepose-hardware -p
  wifi-densepose-sensing-server --no-default-features`: 0 failed
  (198 + 527 lib tests incl. 8 new).
- Firmware: MEASURED on real silicon — built for ESP32-S3 (Atom S3 Lite)
  and run on a live node in a 5-node CSI mesh (~40 pps, edge_tier=2):
  - Link-status packets decode correctly on the host; the reported BSSID
    matches the router's documented 2.4 GHz radio MAC exactly, with
    plausible channel/AP-RSSI and `reassoc_count = 1` per boot.
  - Cadence as designed: one immediate emission within seconds of boot
    (the on-change path — `s_link_ever_sent` starts false), then steady
    ~14 s intervals (600 CSI callbacks at the node's ~40 pps).
  - Verified across both a USB-flashed boot and an OTA-booted image
    (the OTA path needed its own fix — see the companion PR for the
    httpd stack overflow this validation uncovered).

Fixes the telemetry half of #1542; asks 2–3 intentionally not included per the
maintainer's suggested sequencing.

### DIFF
--- a/firmware/esp32-csi-node/README.md
+++ b/firmware/esp32-csi-node/README.md
@@ -203,6 +203,29 @@ All packets are sent over UDP to the configured aggregator. The magic number in 
 | `0xC5110001` | CSI Frame (ADR-018) | ~20 Hz | Variable | Raw I/Q per subcarrier per antenna |
 | `0xC5110002` | Vitals Packet | 1 Hz | 32 bytes | Presence, breathing BPM, heart rate, fall flag, occupancy |
 | `0xC5110004` | WASM Output | Event-driven | Variable | Custom events from WASM modules (u8 type + f32 value) |
+| `0xC511A111` | Link Status (#1542) | ~30 s + on BSSID change | 32 bytes | Associated-AP BSSID, channel, AP RSSI, reassoc count |
+
+### Link Status Packet Format (#1542)
+
+The associated AP is the sensing-link geometry; a roam silently changes what a
+node senses, and frame RSSI cannot reveal it. This packet surfaces the BSSID.
+Sent every `CONFIG_LINK_STATUS_EVERY_N_FRAMES` CSI callbacks (default 600, or
+about 30 s at 20 Hz) plus immediately whenever the BSSID observed via
+`esp_wifi_sta_get_ap_info()` differs from the last reported one.
+
+```
+Offset  Size  Field
+0       4     Magic: 0xC511A111 (LE u32)
+4       1     Node ID
+5       1     Protocol version (0x01)
+6       1     Flags: bit 0 = ap_info_valid
+7       1     Primary channel (0 when not associated)
+8       6     BSSID (all-zero when not associated)
+14      1     AP RSSI as seen by the station, dBm (i8)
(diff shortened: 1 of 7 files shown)


jkuscevic:
Rebased onto current `main` (`a3b6e1d5`) — this was CONFLICTING because main moved under it (including ADR-297 `node_inference` on the sensing-server snapshot path).

Conflict resolution in `v2/crates/wifi-densepose-sensing-server/src/main.rs`: kept main's `node_inference` (ADR-297) and still attach `vs.link_bssid` / `channel` / `reassoc_count` onto the per-node MQTT RSSI payload. The unique diff vs main remains the original 7 files / 1 commit.

Local: `cargo test -p wifi-densepose-hardware -p wifi-densepose-sensing-server --no-default-features` — hardware 198 + sensing-server 530 (the +3 vs the original 527 are main's, not this change).

PR #1683, opened 2026-08-22
fix(security): close remaining sensing and firmware boundaries

iaduartec:
## Summary
- validate recording IDs before filesystem writes
- require API auth on routable binds
- protect all WASM handlers with OTA auth
- keep provisioning passwords out of argv, CSV fallbacks, and local state

## Verification
- `cargo test -p wifi-densepose-sensing-server --no-default-features --lib`: 531 passed, 1 ignored
- firmware unittest suite: 18 passed


iaduartec:
Closing because this repository is not owned by iaduartec; the fork will be removed.

PR #1684, opened 2026-08-22
feat: add native iPhone LiDAR sensor and web viewer

ruvnet:
Implements the first RuView iPhone LiDAR path.

What is included

1. Native SwiftUI and ARKit capture using sceneDepth with device capability checks.
2. Depth confidence intrinsics camera pose sequence and provenance schema.
3. Geometry only transport with no RGB upload.
4. Compact UInt16 millimeter wire encoding with spatial downsampling and 15 FPS cap.
5. Local WebSocket relay and browser point cloud viewer.
6. Node codec tests.
7. ADR 340 defining the production RuView HAL and trust boundary.

Validation

The browser codec tests pass 3 of 3 locally. Apple documents sceneDepth and smoothedSceneDepth as the supported rear LiDAR depth APIs on capable devices. Physical iPhone and Xcode compilation remain required before merge because this environment does not include the Apple SDK or LiDAR hardware.

Key constraint

Mobile Safari does not expose ARKit scene depth to an ordinary web page, so native acquisition is required. The web implementation is intentionally a receiver and visualization surface.

Next production step

Terminate the stream inside the RuView sensing server, authenticate the iPhone via the existing sensor identity path, convert frames into ruv​iew HAL Observation objects, then attach witness and fusion stages.


ruvnet:
Review remediation is now on `080930370`: fixed the relay broadcast-state and text-frame defects; added per-run token authentication, a 2 MB WebSocket limit, static-file allowlisting, CSP headers, canonical packet validation, and a reproducible lockfile; removed the non-Sendable `ARFrame` actor crossing; and corrected the nested Swift wire type compile failure. Validation: local Node tests 5/5, `npm audit` 0 vulnerabilities, claim-check PASS, plus [Node and iOS 17 strict-concurrency compile](https://github.com/ruvnet/RuView/actions/runs/32600678683) PASS. Physical iPhone behavior and the 150 ms p95 target remain explicitly unverified and are not claimed by this merge.

Issue #1755, opened 2026-08-31
fix(provision): per-device state keyed by serial port path, so a different board inherits the previous board's node_id


proffesor-for-testing:
## Summary

`provision.py` keys its per-device state file by **serial port path only**, with no board
identity, so provisioning a different board on a port previously used by another board
silently offers the earlier board's settings — including `node_id`, which is the fusion cohort
key.

`provision.py:126-129`:

```python
def _state_path_for(port: str, state_dir: str) -> str:
    safe = port.replace("/", "_").replace(":", "_").replace("\\", "_")
    return os.path.join(state_dir, f"{safe}.json")
```

Port names are reused aggressively — on macOS a second board plugged into the same hub slot
commonly enumerates as the same `/dev/cu.usbmodemXXXX`.

To be fair to the current design: the merge behaviour is documented (`provision.py:22`),
`--reset` is offered for recycled boards (`:31`, `:362`), and the merged values are printed
before flashing (`:443`). This is a footgun rather than a hidden failure. But the printed
values look identical whether they were chosen by the operator or inherited from a **different
physical board**, so there is nothing to notice.

## Why now

Hit three times while provisioning three ESP32-S3 boards in one session. Each new board was
offered the previous board's `node_id`. Without `--reset` every board would have been
provisioned as the same node id, which on a multistatic deployment silently corrupts the
cohort rather than failing loudly.

## Reproduction

1. `provision.py --port <PORT> --node-id 2 ...` with board A.
2. Unplug A, plug board B into the same port (same device path).
3. `python3 provision.py --port <PORT> --state` → prints board A's `node_id`, `target_port`
   and `channel`.
4. Provisioning B without `--reset` writes those values to B.

## Done when

State is bound to board identity, so a different board on the same port does not inherit.

## Work items

- [ ] Key the state file by chip identity — `esptool read-mac` / chip id — instead of, or in
      addition to, the port path. The script already talks to the chip.
- [ ] When a port's stored identity does not match the connected board, do not merge silently:
      either ignore the stale state or require `--reset`.
- [ ] Mark inherited values distinctly in the pre-flash summary (`:443`), so "inherited from a
      previous board" is visually different from "you passed this".

## Note

`--force-partial` is marked deprecated "since #391/#574 … the script now merges with prior
state by default", which is the change that introduced this. Worth checking those for prior
discussion.

Issue #1754, opened 2026-08-31
fix(provision): WiFi password persisted world-readable (0644) in per-port state file


proffesor-for-testing:
## Summary

`provision.py` persists the WiFi password in cleartext to a per-port state file created with
default permissions, so it lands world-readable (`0644`) under the default umask on macOS and
mainstream Linux.

- The password is deliberately kept in the merged state dict (`provision.py:103`) and
  round-tripping it is covered by tests (`tests/test_provision_state.py:37`), so persistence
  is intentional — the issue is only how it is protected on disk.
- `save_state()` (`provision.py:147-157`) calls `os.makedirs(state_dir, exist_ok=True)` and
  `open(tmp, "w")` with no mode argument. Both take whatever the umask allows.
- Observed on a real host: state directory `0755`, files `0644`.

## Why now

Found while provisioning several nodes for hardware testing. Any local process or user on a
shared machine can read the WiFi credentials of every network the operator has provisioned.

## Reproduction

```
$ python3 provision.py --port <PORT> --ssid <SSID> --password <PW> --target-ip <IP>
$ ls -l ~/.config/wifi-densepose/esp32-provision-state/
-rw-r--r--  ...  _dev_cu.usbmodemXXXX.json
$ python3 -c "import json;print('password' in json.load(open('<that file>')))"
True
```

## Done when

The state directory is `0700` and state files are `0600`, on both a fresh install and an
existing directory created by an earlier version.

## Work items

- [ ] `os.makedirs(state_dir, mode=0o700, exist_ok=True)` and `os.chmod` the directory on the
      existing-directory path, since `makedirs` does not alter an existing one.
- [ ] Create the temp file restrictively (e.g. `os.open(..., 0o600)` or `os.chmod` before
      `os.replace`) so the secret is never briefly world-readable.
- [ ] Tighten permissions on directories/files left behind by earlier versions.
- [ ] Consider whether the password needs to be persisted at all, or could be prompted per
      invocation — that would remove the exposure rather than narrow it.

PR #1760, opened 2026-08-31
fix(provision): owner-only state files, chip-identity binding, and --ota-psk

### LINKED ISSUE #1753: fix(firmware): OTA PSK cannot be provisioned — documented `provision.py --ota-psk` does not exist
## Summary

OTA firmware upload cannot be enabled on any node built from the shipped source, because the
only documented way to provision the device-side PSK does not exist.

`main/ota_update.c:29-31` reads the pre-shared key from NVS namespace `security`, key
`ota_psk`. When it is empty, `ota_check_auth()` rejects every upload (fail-closed, RuView#596).
The comment at `main/ota_update.c:39` tells the operator how to set it:

> The OTA server still starts so the operator can `provision.py --ota-psk <hex>` over USB-CDC
> without a reflash, but the upload endpoint will reject every request until the PSK is set.

`firmware/esp32-csi-node/provision.py` has no `--ota-psk` argument, and contains no reference
to `ota_psk` or the `security` namespace at all — it generates only the `csi_cfg` namespace
(from `provision.py:177`). The only other `ota_psk` in the tree is in
`v2/crates/wifi-densepose-desktop` (`src/domain/config.rs:26`,
`ui/src/pages/Settings.tsx:138`), which is the **client-side** token the desktop app sends,
not something written to device NVS.

Net effect: `OTA=ready` in the boot log means only that the HTTP server started
(`main/main.c:371`, `:504`); authentication can never be satisfied through any shipped
workflow.

## Why now

Observed while testing v0.8.4 on hardware. Nodes are otherwise healthy and streaming, but the
only non-USB update path is unusable, so every firmware change requires physical access.

## Reproduction

On a node flashed from the published v0.8.4 assets and joined to WiFi:

```
$ curl -X POST --data-binary @esp32-csi-node-v0.8.4-8mb.bin \
       -H 'Content-Type: application/octet-stream' \
       http://<node-ip>:8032/ota
HTTP 403 after ~135 KB uploaded
Authentication required. Use: Authorization: Bearer <psk>
```

```
$ python3 provision.py --help | grep -c ota-psk
0
$ grep -c 'ota_psk\|security' provision.py
0
```

`GET /ota/status` on the same node responds normally, so the server is up:

```
{"version":"0.8.4","running_partition":"ota_0","next_partition":"ota_1","max_size":921600}
```

## Done when

- `provision.py` can write `security`/`ota_psk` to device NVS over USB-CDC without a reflash,
  matching the behaviour `ota_update.c:39` already documents; or
- the firmware comment is corrected to name whatever the supported path actually is.

## Work items

- [ ] Add `--ota-psk <hex>` to `provision.py`, emitting the `security` namespace alongside
      `csi_cfg` in the generated NVS CSV.
- [ ] Validate the value (the firmware caps it at `OTA_PSK_MAX_LEN` 65, hex-encoded SHA-256).
- [ ] Decide whether the PSK belongs in the per-port state file — if so it needs the same
      protection as the WiFi password (see the state-file permissions issue).
- [ ] Consider making the boot log distinguish "OTA server started" from "OTA can accept
      uploads", since `OTA=ready` currently implies the latter.

## Secondary, untested

`/ota/status` advertises `max_size: 921600` while the shipped v0.8.4 image is 1,126,624 bytes
and `ota_1` is 2,097,152 bytes. Whether that gate would reject a valid image could not be
tested, because the auth gate rejects first. Flagging only; it may be a stale constant.

proffesor-for-testing:
Fixes #1753, #1754, #1755.

Three operator-safety defects found while provisioning three ESP32-S3 boards against v0.8.4 on real hardware.

## What changed

**#1754 — credential exposure.** The per-port state file holds the WiFi password in cleartext and was created with the process umask, landing `0644` on macOS and mainstream Linux (observed `0755`/`0644` on a real host). Directory is now `0700`, file `0600`, opened via `os.open` with the mode so the secret is never briefly world-readable. An explicit `chmod` covers directories left by earlier versions, since `makedirs` cannot tighten an existing one. `chmod` failures warn rather than abort, so a read-only or exotic filesystem still provisions.

**#1755 — cross-board state inheritance.** State is keyed by sanitised port path only, and port names are reused when boards are swapped. The board MAC is now recorded in the state file and checked before merging; a mismatch exits non-zero and points at `--reset`.

Deliberately permissive where it must be: unknown identity on **either** side still merges, so existing state files keep working after upgrade and a board that will not answer over serial is still provisionable exactly as before. `--state` remains read-only and skips the chip round-trip.

**#1753 — OTA unprovisionable.** `main/ota_update.c:39` documents `provision.py --ota-psk <hex>` as the way to set the PSK without a reflash, but the flag did not exist and nothing wrote the `security` namespace — so every node rejected all OTA uploads (verified: HTTP 403 on a live v0.8.4 node). The flag now emits `security`/`ota_psk` alongside `csi_cfg`, validated against the firmware's own constraints (hex, ≤64 chars per `OTA_PSK_MAX_LEN`).

## Testing

23 tests pass (9 added). Each new test was confirmed to **fail with its fix reverted**, so they bind to the behaviour rather than passing incidentally:

| control | result |
|---|---|
| revert permissions to default `open()` | 3 failures |
| revert identity check to always-match | 1 failure |
| revert `security` namespace emission | 2 failures |
| all fixes restored | 23 pass |

## Notes for review

- The identity key is `_chip_mac`, underscore-prefixed so it cannot collide with an NVS key.
- `--ota-psk` is added to `MERGEABLE_ATTRS`, so it persists like other settings. If you would rather the PSK were **not** written to a state file at all, say so and I will make it invocation-only — that is a policy call, and #1754 makes the file owner-only either way.
- I have not changed the `OTA=ready` boot log line, which reports only that the HTTP server started. Mentioned in #1753 as a separate call.

Found during a hardware testing session against v0.8.4 with three ESP32-S3 nodes.

### DIFF
--- a/firmware/esp32-csi-node/provision.py
+++ b/firmware/esp32-csi-node/provision.py
@@ -108,6 +108,7 @@ MERGEABLE_ATTRS = [
     "channel", "filter_mac",
     "hop_channels", "hop_dwell",
     "seed_url", "seed_token", "zone", "swarm_hb", "swarm_ingest",
+    "ota_psk",
 ]
 
 
@@ -129,6 +130,47 @@ def _state_path_for(port: str, state_dir: str) -> str:
     return os.path.join(state_dir, f"{safe}.json")
 
 
+# Key under which the board's own identity is recorded inside the per-port
+# state file. Underscore-prefixed so it can never collide with an NVS key and is
+# trivially filtered out of the merge.
+STATE_IDENTITY_KEY = "_chip_mac"
+
+
+def read_chip_mac(port: str, baud: int, chip: str):
+    """Return the connected board's MAC, or None if it cannot be read.
+
+    Best-effort and never fatal: identity binding is a safety net, so a board
+    that will not answer must still be provisionable exactly as before.
+    """
+    try:
+        out = subprocess.run(
+            [sys.executable, "-m", "esptool", "--chip", chip, "--port", port,
+             "--baud", str(baud), "read-mac"],
+            capture_output=True, text=True, timeout=60,
+        ).stdout
+    except (OSError, subprocess.SubprocessError):
+        return None
+    for line in out.splitlines():
+        # esptool prints e.g. "MAC: 80:b5:4e:c1:b5:68"
+        if line.strip().upper().startswith("MAC:"):
+            mac = line.split(":", 1)[1].strip().lower()
+            if len(mac.split(":")) == 6:
+                return mac
+    return None
+
+
+def identity_matches(prior: dict, chip_mac) -> bool:
+    """True when prior state may be merged onto the connected board.
+
+    Unknown on either side means "cannot prove a mismatch", which stays
+    permissive so existing state files keep working after upgrade.
+    """
+    recorded = prior.get(STATE_IDENTITY_KEY)
+    if not recorded or not chip_mac:
+        return True
+    return recorded == chip_mac
+
+
 def load_state(port: str, state_dir: str) -> dict:
     """Return the merged-state dict for `port`, or `{}` if absent / unreadable."""
     path = _state_path_for(port, state_dir)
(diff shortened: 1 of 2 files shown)


proffesor-for-testing:
Converting to draft — an adversarial review of my own patch found it does not do what the description claims. Notably:

1. `--ota-psk` alone is rejected by `has_config_value()` (`ota_psk` is missing from `CONFIG_VALUE_CHECKS`), so the headline feature cannot provision a fresh device standalone. Verified: exits 2 with "At least one config value must be specified".
2. The MAC is only read when prior state already exists, so a **first** provisioning records no identity — which is exactly the swap case #1755 describes. The guard misses the common path.
3. The error text suggests "pass every value explicitly" as an alternative, but the check runs before the merge, so only `--reset` actually works. The diagnostic is misleading.
4. `--dry-run` now opens the serial port and can stall ~60s.
5. The temp file is opened `O_TRUNC` without `O_EXCL`/`O_NOFOLLOW`, so a stale or planted path defeats the 0600 mode — in a patch whose whole point is not leaking the secret.
6. `--ota-psk` accepts a single hex character for a firmware-upload bearer credential.
7. The new tests sit below `unittest.main()`; direct execution silently runs 11 of 23.

Reworking. Apologies for the noise.

proffesor-for-testing:
Reworked and back in review. Every defect from the self-review above is fixed, each with a regression test, and each test verified to fail with its fix reverted.

| defect | fix | control |
|---|---|---|
| `--ota-psk` alone rejected by `has_config_value()` | added to `CONFIG_VALUE_CHECKS` | 1 failure when reverted |
| first provisioning recorded no MAC | identity read on any run that talks to the device | covered |
| "pass every value explicitly" was false | removed from the diagnostic | — |
| `--dry-run` opened the serial port | `--dry-run` and `--state` never touch hardware | asserted |
| `O_TRUNC` without `O_EXCL`/`O_NOFOLLOW` | unlink stale temp, open `O_EXCL\|O_NOFOLLOW` | 1 failure when reverted |
| 1-char PSK accepted | exactly 64 hex required | asserted |
| legacy perms only repaired on write | repaired on both read paths | 1 failure when reverted |
| PSK printed verbatim by `--state` | redacted (still stored — flashing rewrites the whole NVS partition) | — |
| tests below `unittest.main()` | runner moved to end; direct run now 24/24 | verified |

One test of mine did not bind and was replaced. It asserted the *final* file mode, which the trailing `chmod` fixes whether or not `O_EXCL` is used. The replacement plants a symlink at the predictable temp path — and with the fix reverted it fails, confirming the cleartext password really was written through the symlink.

Two things I did **not** fix, deliberately:

- The MAC gate is fail-open (unknown identity on either side still merges) and has a swap-after-`read-mac` TOCTOU. It is a guard against the common accident — swapping boards on a reused port — not a security boundary. Making it binding would lock out boards that will not answer over serial, which seems worse. Happy to change if you disagree.
- Windows cannot retrofit owner-only ACLs via `chmod`, so the legacy-permission migration is POSIX-only and the tests skip there. Fixing it properly needs an ACL API, which felt out of scope.

Also worth noting: `nvs_partition_gen` accepts the two-namespace CSV, and the firmware reads `security`/`ota_psk` via `nvs_open()` + `nvs_get_str()` — that part was verified end to end.

PR #1761, opened 2026-08-31
fix(ci): publish every firmware variant to the release, not just 8mb

### LINKED ISSUE #1757: fix(release): C6 firmware is built and tested by CI every release but never published
## Summary

`firmware-ci.yml` builds and tests an ESP32-C6 variant on every firmware tag, and the release
notes state it passed, but no C6 artifact has been published since v0.8.0. A C6 operator's only
downloadable option is four minor versions behind.

`.github/workflows/firmware-ci.yml:67-74` defines the `c6-4mb` matrix entry (target `esp32c6`,
`artifact_app: esp32-csi-node-c6.bin`, `artifact_pt: partition-table-c6.bin`), and C6 is the
only variant that additionally builds and runs the ADR-110 host unit tests. The v0.8.4 release
notes say: *"GitHub firmware matrix: ESP32-S3 8 MB, ESP32-S3 4 MB, ESP32-C6 all passed."*

Published assets:

| release | C6 asset |
|---|---|
| v0.8.4-esp32 | none — only `-8mb.bin` and `-8mb-flash-bundle.zip` |
| v0.8.3-esp32 | none |
| v0.8.2-esp32 | none |
| v0.8.1-esp32 | none |
| v0.8.0-esp32 | `esp32-csi-node-c6.bin`, `bootloader-c6.bin`, `partition-table-c6.bin` |

## Why now

Preparing an ESP32-C6 test round. v0.8.0 predates the v0.8.4 empty-room calibration fixes
("calibration now advances from `Uncalibrated` and accepts wide HT40 frames through canonical
resampling"), so a C6 node on the newest published firmware cannot be compared like-for-like
against S3 nodes on v0.8.4.

## Confirmation that the build works

Built v0.8.4 for C6 locally using the exact CI toolchain and recipe — `espressif/idf:v5.4`,
`idf.py set-target esp32c6`, `idf.py build`, no config changes (the `sdkconfig.defaults.esp32c6`
overlay is auto-applied and pins `partitions_4mb.csv` itself):

- binary 1,051,008 bytes = 1026 KB — inside the CI budget (warn 1100 KB / fail 1152 KB)
- ADR-110 host unit tests: 21 passed, 0 failed
- embedded app descriptor: version `0.8.4`, project `esp32-csi-node`, idf `v5.4` — identical to
  the published S3 image

So the artifact the matrix already produces is releasable as-is.

## Done when

Tagged firmware releases publish the C6 artifacts alongside the S3 ones.

## Work items

- [ ] Upload `esp32-csi-node-c6.bin`, `bootloader-c6.bin`, `partition-table-c6.bin` (and
      `ota_data_initial.bin`) from the `c6-4mb` matrix job to the release.
- [ ] Publish a C6 flash bundle equivalent to the S3 `-flash-bundle.zip`.
- [ ] Publish SHA-256 sums for the C6 assets, as the S3 assets already do.
- [ ] Either drop the "ESP32-C6 all passed" line from release notes or make it correspond to a
      downloadable artifact.

proffesor-for-testing:
Fixes #1757.

## The problem

The build matrix has produced ESP32-C6 artifacts on every firmware tag since v0.8.0 — and C6 is the only variant that *additionally* builds and runs the ADR-110 host tests — yet no C6 asset has been published since v0.8.0. The v0.8.4 notes say "ESP32-C6 all passed" while the release carries only the S3 8mb binary. A C6 operator's newest downloadable firmware is four minor versions behind, missing the v0.8.4 calibration fixes.

## Root cause

`firmware-ci.yml` never attached anything to a release. It uploads 90-day Actions artifacts only, so release assets were being attached **by hand** — `v0.8.4` assets show `uploader=ruvnet`. That is why one variant can be silently forgotten. `ci.yml` and `desktop-release.yml` already automate release uploads; firmware did not.

## Changes

**1. Stage a bootloader + initial OTA data for every variant.** Previously only `8mb` got them, so the C6 artifacts could not flash a blank board even if published. The `8mb` asset names are deliberately **unchanged** so existing docs and scripts keep working; other variants get suffixed names (`bootloader-c6.bin`, `ota_data_initial-c6.bin`). Each variant also emits `SHA256SUMS-<variant>.txt`, matching what the S3 release currently publishes by hand.

**2. Add a `release` job** gated on a `v*-esp32` tag that downloads every variant's staged binaries and attaches them, with `fail_on_unmatched_files: true` so a missing variant fails loudly instead of shipping a partial release. Uses the same pinned `softprops` action already used by `desktop-release.yml`, and a `download-artifact` SHA already pinned elsewhere in this repo.

## Verification

I built v0.8.4 for C6 locally using the exact CI toolchain and recipe (`espressif/idf:v5.4`, `idf.py set-target esp32c6`, `idf.py build`, no config changes — the `sdkconfig.defaults.esp32c6` overlay is auto-applied and pins `partitions_4mb.csv` itself):

| check | result |
|---|---|
| binary size (budget warn 1100 KB / fail 1152 KB) | 1,051,008 B = **1026 KB** |
| ADR-110 host unit tests | **21 passed, 0 failed** |
| app descriptor | version `0.8.4`, project `esp32-csi-node`, idf `v5.4` — identical to the published S3 image |

So the artifact this job would attach is known good.

`actionlint` clean on the change. The one remaining `SC2086` at line 88 is pre-existing and untouched.

## Note for review

`fail_on_unmatched_files: true` is deliberate — a release that silently ships fewer variants than it built is the bug being fixed. If you would rather it warn, that is a one-word change.

### DIFF
--- a/.github/workflows/firmware-ci.yml
+++ b/.github/workflows/firmware-ci.yml
@@ -64,6 +64,8 @@ jobs:
             size_limit_kb: 1152
             artifact_app: esp32-csi-node-4mb.bin
             artifact_pt: partition-table-4mb.bin
+            artifact_bootloader: bootloader-4mb.bin
+            artifact_otadata: ota_data_initial-4mb.bin
           # ADR-110: ESP32-C6 research target (Wi-Fi 6 / 802.15.4 / TWT / LP-core)
           - variant: c6-4mb
             target: esp32c6
@@ -73,6 +75,8 @@ jobs:
             size_limit_kb: 1152
             artifact_app: esp32-csi-node-c6.bin
             artifact_pt: partition-table-c6.bin
+            artifact_bootloader: bootloader-c6.bin
+            artifact_otadata: ota_data_initial-c6.bin
 
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
@@ -162,10 +166,18 @@ jobs:
           mkdir -p release-staging
           cp build/esp32-csi-node.bin                      release-staging/${{ matrix.artifact_app }}
           cp build/partition_table/partition-table.bin     release-staging/${{ matrix.artifact_pt }}
+          # Every variant needs a bootloader and initial OTA data to flash a
+          # blank board; staging them only for 8mb left the C6 artifacts
+          # unflashable (RuView#1757). The 8mb names are unchanged so existing
+          # docs and scripts keep working.
           if [ "${{ matrix.variant }}" = "8mb" ]; then
             cp build/bootloader/bootloader.bin             release-staging/bootloader.bin
             cp build/ota_data_initial.bin                  release-staging/ota_data_initial.bin
+          else
+            cp build/bootloader/bootloader.bin             release-staging/${{ matrix.artifact_bootloader }}
+            cp build/ota_data_initial.bin                  release-staging/${{ matrix.artifact_otadata }}
           fi
+          ( cd release-staging && sha256sum -- * > SHA256SUMS-${{ matrix.variant }}.txt )
           ls -la release-staging/
 
       - name: Check QEMU ESP32-S3 support status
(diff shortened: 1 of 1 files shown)


proffesor-for-testing:
Adversarial review of my own patch found a blocker, now fixed in the latest commit.

`version-guard` and `build` run in parallel, and the release job declared only `needs: build`. A tag whose `version.txt` did not match would fail the guard and still publish a release once the matrix went green — defeating the guard's entire purpose. Now `needs: [version-guard, build]`.

The rest of the review came back clean and worth recording: `v0.8.5-esp32` satisfies both the tag filter and the job condition; existing 8mb names stay `bootloader.bin`/`ota_data_initial.bin`; 4mb/C6 names are unique; the app size gate is unchanged; `artifacts/**/*` matches the per-artifact subdirectories so `fail_on_unmatched_files` should not fail a healthy matrix; `contents: write` is sufficient; and all four pinned action SHAs resolve to genuine upstream releases.

PR #1782, opened 2026-09-03
fix: ESP32 flash tooling, occupancy noise floor, multistatic timestamp clamp

scottjoyner:
## Summary

Five commits covering the xwing ESP32 CSI bring-up work and two sensing fixes found while validating the Kipnerter 1.6 ambient-RuView integration against the live deployment.

### 1. scripts: one-shot flash+provision for ESP32 CSI nodes (1f170118)
New scripts/flash_esp32_node.sh — prebuilt release_bins + NVS provisioning. 4MB GD25Q32 layout validated on hardware; classic ESP32 port builds clean under espressif/idf:v5.4 with partitions_4mb.csv + FLASHSIZE_4MB overrides.

### 2. fix(flash): auto-detect classic ESP32 vs S3; correct per-chip offsets (0e8d2cfb)
Classic ESP32 bootloader belongs at 0x1000 — the hardcoded S3 layout (0x0, release_bins) boot-looped D0WD-V3 nodes with 'invalid header'. The script now probes chip type, picks bins (build/ for classic 4MB, release_bins for S3) and resolves idf.py-nested or flat binary layouts. provision.py chip defaults to esptool auto-detect.

### 3. fix(sensing): clamp multistatic frame timestamps to process epoch (#1194, cf7c9a51)
NodeStates created between socket start and first fusion carry Instants older than the LazyLock EPOCH; saturating duration_since produced ts=0 frames whose spread-vs-live-nodes equaled server uptime, failing every mixed trust cycle with TimestampMismatch. Frames now clamp to the epoch floor and are dropped when they carry no positive monotonic offset.

### 4. fix(signal): unbiased Marcenko-Pastur noise floor in occupancy estimator (#1195, becdfa0f)
Bottom-half eigenvalue averaging sat below sigma^2 for white input, placing the significance threshold inside the noise bulk — deterministic phantom occupancy on some BLAS backends. Use tr(Sigma)/r over positive eigenvalues (rank-deficiency safe); test fixture is now true seeded i.i.d. noise.

### 5. test(sensing): pin lazy EPOCH before fixture Instant::now() stamps (12d80e7c)
The #1194 clamp exposed a fixture race: test NodeStates stamped before the first fusion initialized the process-global EPOCH were silently dropped (7 engine_bridge test failures, order-dependent). Adds a #[cfg(test)] pin_epoch() hook and pins it before any fixture captures an instant.

## Verification

- cargo test -p wifi-densepose-signal: 478/478 pass
- cargo test -p wifi-densepose-sensing-server: 485 + 225 + 12 pass, 3 consecutive runs (previously 7 bin-test failures, run-order dependent)
- cargo check --workspace: clean (only pre-existing warnings in wifiscan/matter)
- Live contract check: kipnerter-ios scripts/ruview_smoke.sh (feature/1.6-ambient-ruview-voice) against the running deployment — /health/ready, /api/v1/{status,info,pose/current,pose/zones/summary}, 401 controls, and the authenticated /ws/sensing upgrade all PASS (sensing_update frames with tick/persons/vital_signs)

Part of the Kipnerter 1.6 ambient-RuView integration (scottjoyner/kipnerter-ios feature/1.6-ambient-ruview-voice).

### DIFF
--- a/scripts/flash_esp32_node.sh
+++ b/scripts/flash_esp32_node.sh
@@ -1,8 +1,19 @@
 #!/bin/bash
-# Flash + provision a RuView ESP32-S3 CSI node (prebuilt release binaries).
+# Flash + provision a RuView CSI node (prebuilt binaries, chip auto-detected).
+#
+# Supports BOTH targets:
+#   - Classic ESP32 (D0WD/PICO): uses firmware/esp32-csi-node/build/ (4MB layout,
+#     built locally via dockerized ESP-IDF — see firmware README). Bootloader
+#     offset is 0x1000 on classic parts; flashing it at 0x0 boot-loops with
+#     "invalid header".
+#   - ESP32-S3: uses firmware/esp32-csi-node/release_bins/ (8MB default, or
+#     4MB variant when FLASH_SIZE=4MB).
+#
 # Usage:
-#   bash scripts/flash_esp32_node.sh /dev/ttyACM0 <NODE_ID> <WIFI_SSID> <WIFI_PASSWORD>
-# Defaults: aggregator target = xwing LAN IP 192.168.1.86:5005, 8MB flash.
+#   bash scripts/flash_esp32_node.sh <port> <node_id> <wifi_ssid> <wifi_password>
+# Env overrides:
+#   TARGET_IP   aggregator host   (default 192.168.1.86 = xwing LAN)
+#   FLASH_SIZE  S3 flash size    (default 8MB; ignored for classic = always 4MB)
 set -euo pipefail
 
 PORT="${1:?usage: flash_esp32_node.sh <port> <node_id> <ssid> <password>}"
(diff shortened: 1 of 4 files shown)


ruvnet (commented):
Dream Machine exact-head evidence review — **REJECT**

Frozen hypothesis: this branch fixes flashing portability, phantom occupancy, and timestamp handling without weakening supported environments.

The candidate contradicts the portability claim: the flashing script embeds one maintainer-specific checkout/virtualenv layout and accepts Wi-Fi credentials as a positional argument. The field-model change also labels trace/r over nonzero eigenvalues as an unbiased noise estimate without evidence across rank-deficient aspect ratios. The PR is currently non-mergeable and has no exact-head workflow run.

Acceptance gate: remove machine-specific paths; accept credentials through a non-history mechanism; add a dry-run fixture for a clean checkout; run seeded Monte Carlo cases spanning p<n, p=n, and p>n with a falsifiable bias threshold; preserve an explicit first-frame timestamp contract; and run the complete consumer workflow set on this exact head. This review makes no claim about the private advisories linked from issue #1783.

Issue #1786, opened 2026-09-03
Research: CSI optical flow as a dense RF motion field primitive


ruvnet:
## Finding

WiFlow: Estimating Optical Flow using WiFi Channel State Information, arXiv:2609.02452, announced 2026-09-03, introduces a CSI optical-flow dataset and three architectures intended to estimate dense motion fields from WiFi CSI rather than camera frames. The project reports code and data at https://visinf.github.io/wiflow.

Claim class: originating-team newly released. No RuV performance result is claimed here.

## Why it matters

RuView currently has multiple sensing heads above CSI. A dense motion field could be a more reusable intermediate representation than another activity classifier because motion can feed tracking, pose, occupancy, fall/mobility analysis, world-state updates, sensor fusion, and camera-free spatial inference.

## RuV mapping

RuView ingestion and calibration -> CSI encoder -> RF motion field -> RuField/WorldGraph temporal update -> RuVector memory -> downstream task heads.

Potential consumers include pose/activity heads, WorldGraph motion hypotheses, LiDAR fusion, UWB/FMCW fusion, and edge visualization.

## Reproduction first

Pin the WiFlow repository and dataset version. Record CSI hardware, antenna geometry, carrier/bandwidth, packet rate, synchronization, room layouts, labels, train/test split, model parameters, software versions, accelerator, and license before implementation.

Compare:

1. existing RuView motion/activity representation
2. a simple temporal CSI baseline
3. WiFlow reference architecture
4. a reduced edge candidate only if the reference reproduction succeeds

## Metrics

Optical-flow endpoint error or the paper's released primary metric, cross-room degradation, cross-person degradation, cross-hardware degradation where possible, temporal stability, downstream tracking/pose gain, calibration, p50/p95 latency, memory, model size, power, packet loss sensitivity, missing subcarrier sensitivity, and total sensor cost.

## Falsification

The optical-flow representation should be rejected as a core primitive if it improves its own pixel/field metric but does not improve at least two downstream RuView tasks on held-out rooms, or if a simpler temporal embedding performs within variance at materially lower edge cost.

## Promotion

Require a reproduced reference result plus at least 10% relative gain on two held-out downstream motion/spatial metrics or equivalent quality with at least 30% lower end-to-end inference cost. Camera data may be used for labels during research but must not become an undeclared production dependency for camera-free claims.

Keep this isolated and reversible. No production sensing claim until real RuView hardware reproduces transfer.

ruvnet:
[...]

PR #1787, opened 2026-09-03
research: add WiFlow CSI optical flow reproduction protocol

ruvnet:
Tracks #1786.

Adds a reproduction first protocol for evaluating CSI optical flow as a reusable RuView intermediate representation.

The protocol requires held out rooms, a second RF hardware family, realistic degradation, explicit edge latency and energy accounting, and downstream validation on at least two RuView tasks. A better optical flow proxy score alone does not qualify the representation.

Existing sensing heads remain the rollback path. No production model or deployment behavior changes in this PR.


flwadajran-cyberk (commented):
Open [open```

```]()

PR #1788, opened 2026-09-03
C6: null-calibrated presence and recovery health diagnostics

proffesor-for-testing:
## Review scope

This is an exact-change review PR for the three C6 commits developed and physically exercised on nodes 11-13. It intentionally targets `codex/c6-recovery-review-base` (the original source parent) because current `main` does not contain `calibration_session.rs` and a PR directly to `main` would include thousands of unrelated NLOS/social-history files. **Do not merge this stacked review PR as-is.** Please review the detector, HEALTH wire contract, retry/recovery behavior, and test coverage; integration onto the selected upstream sensing baseline should follow after that decision.

## Changes

- expose per-node and aggregate residual-energy evidence
- normalize presence against learned empty-room residual distributions
- add hysteresis and configurable node quorum
- extend C6 HEALTH with association/channel/disconnect/retry/UDP/heap diagnostics
- raise Wi-Fi retry window to 90 attempts and preserve stale-to-active recovery
- add hardware recovery evidence harness and reset-reason decoding
- support the ESP-IDF 6 GPIO wake API

## Verification

- `python3 -m unittest -v test_wifi_recovery_harness.py`: 4 passed
- `make check LDLIBS=` on macOS: 60 assertions passed across adaptive controller, feature state, and mesh suites
- focused Rust run reached and passed the changed detector/HEALTH tests, including `null_normalized_detector_requires_quorum_and_hysteresis`, calibrated occupancy, and extended HEALTH decoding
- physical nodes 11-13: helper restart, stale-to-active, >90 s AP outage, and AP reboot/channel-change harness scenarios all passed

## Known gate outside this diff

The existing `auth_wiring::with_auth_off_both_listeners_stay_open` fails reproducibly before its assertions: the server refuses unauthenticated startup because calibration guidance requires RuView API authentication. The C6 commits do not modify that policy. This draft does not claim a globally green workspace.

## Review questions

1. Is the learned-null normalization faithful to the contrastive zero-first design?
2. Are quorum and hysteresis defaults appropriate for three-node C6 deployments?
3. Should malformed HEALTH rejection expose decoder reasons and counters?
4. Where should this be rebased now that current `main` lacks the original calibration-session baseline?

### DIFF
--- a/firmware/esp32-csi-node/main/c6_lp_core.c
+++ b/firmware/esp32-csi-node/main/c6_lp_core.c
@@ -13,7 +13,7 @@
  *      `ulp_embed_binary(...)` in main/CMakeLists.txt from lp_core/main.c.
  *
  *   2. DISABLED — falls back to plain deep-sleep + GPIO wake-up
- *      (`esp_deep_sleep_enable_gpio_wakeup`). No debounce, no
+ *      (ESP-IDF deep-sleep GPIO wake API). No debounce, no
  *      sub-10 µA floor, but no LP toolchain dependency either.
  *      This is the path the v0.6.6 firmware shipped with.
  *
@@ -26,6 +26,7 @@
 #if defined(CONFIG_IDF_TARGET_ESP32C6) && defined(CONFIG_ULP_COPROC_TYPE_LP_CORE)
 
 #include "c6_lp_core.h"
+#include "esp_idf_version.h"
 #include "esp_log.h"
 #include "esp_sleep.h"
 #include "driver/rtc_io.h"
@@ -134,10 +135,17 @@ esp_err_t c6_lp_core_arm(int wake_gpio, bool active_high)
 #else
     /* --- Fallback path: plain deep-sleep GPIO wakeup (~10 µA floor) --- */
     uint64_t mask = 1ULL << wake_gpio;
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
+    esp_sleep_gpio_wake_up_mode_t mode = active_high
+        ? ESP_GPIO_WAKEUP_GPIO_HIGH
+        : ESP_GPIO_WAKEUP_GPIO_LOW;
+    esp_err_t err = esp_sleep_enable_gpio_wakeup_on_hp_periph_powerdown(mask, mode);
+#else
     esp_deepsleep_gpio_wake_up_mode_t mode = active_high
         ? ESP_GPIO_WAKEUP_GPIO_HIGH
         : ESP_GPIO_WAKEUP_GPIO_LOW;
     esp_err_t err = esp_deep_sleep_enable_gpio_wakeup(mask, mode);
+#endif
     if (err != ESP_OK) {
         ESP_LOGE(TAG, "enable_gpio_wakeup failed: %s", esp_err_to_name(err));
         return err;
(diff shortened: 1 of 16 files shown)


proffesor-for-testing:
CI triage for the exact-parent review stack:

Representative failing Firmware CI, CSI data policy, fix-marker, and benchmark jobs do not reach their substantive test steps. `actions/checkout` aborts while initializing recursive submodules with:

```text
fatal: No url found for submodule path 'dream-machine' in .gitmodules\n```\n\nThat gitlink/`.gitmodules` inconsistency is in the historical review base. The draft remains useful for an exact three-commit/16-file review, but its current red checks are not C6 test results and it is not merge-ready. The next integration step is to select a current baseline containing the calibration-session contract and transplant these three commits there.

ruvnet:
Reviewed this alongside cognitum-one/ruview-mobile#4 and cognitum-one/cognitum#625 (the session report). Solid, well-tested work — two real, previously-hidden bugs found and fixed:

1. `temporal_motion_score` was comparing `frame_history.back()` — since every caller appends the current frame *before* calling `extract_features_from_frame`, this compared the current frame to itself, so motion detection was structurally always near-zero. Fixed to `frame_history.iter().rev().nth(1)`, with a good regression pair (`identical_consecutive_frames_remain_motionless` vs. an assertion that real motion is actually detected). This looks like it could have been silently degrading motion-dependent features for a while — worth tracking down how long it's been broken.
2. Feature-extraction sample rate was hardcoded to `1000.0/500.0 = 2.0 Hz` regardless of measured cadence — replaced with real per-node measured rate via `observe_accepted_feature_frame`, sane 20 Hz default pre-warmup.

Minor, non-blocking notes:
- The same `frame_history.back()` fix is duplicated verbatim in both `csi.rs` and `main.rs` — pre-existing duplication, not introduced here, but worth unifying eventually so a future fix can't land in only one copy.
- `node_health_json_exposes_diagnostics_without_promoting_health_to_csi_liveness` — the test name implies HEALTH packets shouldn't affect CSI-liveness tracking. I confirmed by reading the UDP receiver that this is true (HEALTH only sets `latest_health`, never touches `last_frame_time`), but nothing explicitly asserts it — a small dedicated test would make the invariant regression-proof rather than implicit.

On your review questions:
- **Null normalization design**: the mechanism itself is sound (residual energy after projecting out learned baseline + environmental modes, compared against learned empty-room p95/p99 quantiles) — I can't confirm fidelity to a specific "contrastive zero-first" design doc without that reference, but the null-space residual approach is methodologically standard for this kind of anomaly/presence detection.
- **Quorum/hysteresis defaults** (quorum=2 of 3, enter=3 frames, exit=10 frames): reasonable for a 3-node deployment — majority quorum, and the asymmetric enter/exit favors stability over responsiveness, which seems like the right tradeoff for a presence detector.
- **Malformed HEALTH rejection reasons**: yes — #625 independently flags this as open defect #3 (the helper log currently only records "Rejected ESP32 mesh HEALTH datagram" with no reason/length/counter). Agree it's worth adding structured rejection telemetry.
- **Rebase target**: that's a real call only ruv can make given the branch topology described in #625 — flagging that it's still open, not attempting to answer it here.

Agree with the PR description that this shouldn't merge as-is (targets a throwaway review branch, and #625 notes RuView's CI is currently broken for unrelated submodule reasons on this branch, so there's no real green-CI signal yet either way).

Issue #301, opened 2026-03-24
Known Issues & Setup Troubleshooting Guide


ruvnet:
This issue collects the most common setup problems and their solutions. Check here before opening a new issue.

---

## Quick Checks

| Symptom | Likely Cause | Fix |
|---------|-------------|-----|
| UI shows same skeleton regardless of people | You're in **simulation mode** | Check `curl http://localhost:3000/health` — if `"source":"simulated"`, see **Simulation vs Live** below |
| Server says "No hardware detected" | ESP32 not sending UDP to your computer | Check `--target-ip`, firewall, WiFi connection (see **No Data Arriving**) |
| Only 1 node detected despite multiple ESP32s | Node IDs not persisted | Flash [v0.5.0-esp32](https://github.com/ruvnet/RuView/releases/tag/v0.5.0-esp32), re-provision each node separately |
| "Always 2 persons" when only 1 present | Multipath ghosting | Use 3+ nodes, update to latest server (PR #295 hysteresis fix) |
| Skeletons shaking / jittery | No trained model + count instability | PR #295 hysteresis fix + record calibration data |
| Watchdog crash `CPU 1: edge_dsp` | Firmware DSP task not yielding | Flash [v0.4.3-esp32](https://github.com/ruvnet/RuView/releases/tag/v0.4.3-esp32) or later |
| Training endpoint returns instantly | No recorded data to train on | Record labeled data first (see **Training** below) |
| `Fall detected!` flooding logs | Threshold too low in old firmware | Flash v0.4.3+ (threshold raised from 2.0 to 15.0) |

---

## Simulation vs Live Mode

**How to tell which mode you're in:**

```bash
curl http://localhost:3000/health
```

- `"source":"simulated"` → You're seeing synthetic data. The UI will not respond to real room conditions.
- `"source":"esp32"` → You're receiving real CSI data from hardware.

**To switch to live mode:**

1. Flash ESP32-S3 with [v0.5.0-esp32 firmware](https://github.com/ruvnet/RuView/releases/tag/v0.5.0-esp32)
2. Provision with your computer's **real LAN IP** (not `0.0.0.0`):
   ```bash
   # Find your IP: ifconfig (macOS) / ipconfig (Windows) / hostname -I (Linux)
   python firmware/esp32-csi-node/provision.py --port YOUR_PORT \
     --ssid "YourWiFi" --password "YourPassword" --target-ip YOUR_COMPUTER_IP
   ```
[...]

Issue #249, opened 2026-03-13
Help Ruvnet! Detection Window showing the same thing regardless of position and people in room.


munozluis1015:
I just created a mesh of ESP32 S3s and when I go to the detection window, it always shows the same thing regardless of position or amount of people in the room. To make a mesh do I have to follow this?, [https://github.com/ruvnet/RuView/issues/34](url)
I didn't follow that and simply flashed all the different ESP32's with a different COM and connected them to my WIFI.

<img width="1241" height="917" alt="Image" src="https://github.com/user-attachments/assets/553de771-ac0f-46c9-a5de-fdbc42d32b9c" />

[https://imgur.com/a/doggYlx](url)

HSIEEN:
Seems we encounter a same problem. I tried traning but not worked.

ruvnet:
What training process are you using?

munozluis1015:
> What training process are you using?
I haven’t done any training yet. This is straight after flashing the ESP32s. It should still work roughly without any training no?

ruvnet:
The detection window currently runs a **simulation mode** when no real-time CSI data is connected via WebSocket. The poses you see are synthetic — they don't change based on actual room conditions.

To get real detection from your ESP32 mesh:

1. **Start the aggregator server** — this collects CSI frames from your ESP32 nodes via UDP and exposes a WebSocket endpoint
2. **Connect the UI** — in the detection/observatory page, enter your WebSocket URL (e.g. `ws://localhost:3030/ws/csi`) and click Connect
3. **Verify CSI data is flowing** — check the serial monitor on each ESP32 to confirm frames are being sent

Without the WebSocket connection, the UI has no way to receive your ESP32 data and falls back to the built-in simulation.

**Training is not required** for basic detection — the pre-trained model should respond to CSI signal changes. But you do need the data pipeline (ESP32 → UDP → aggregator → WebSocket → browser) connected end-to-end.

If you're unsure whether your ESP32s are sending data correctly, check that:
- Each node has a unique `node_id` (set via `provision.py`)
- They're sending UDP packets to the aggregator's IP/port
- The aggregator logs show incoming frames

munozluis1015:
> The detection window currently runs a **simulation mode** when no real-time CSI data is connected via WebSocket. The poses you see are synthetic — they don't change based on actual room conditions.
> 
> To get real detection from your ESP32 mesh:
> 
[...]

PR #1791, opened 2026-09-04
fix(sensing-server): wire --node-positions into live NodeInfo output

clonea1:
## What is wrong

`--node-positions` is parsed and stored correctly, and multistatic fusion uses
it. But the per-node JSON builder never reads it. Both the vitals-only and the
governed/privacy-gated ESP32 live-data paths emit a hardcoded fallback instead:

    position: [2.0, 0.0, 1.5]

That literal appears at four sites in `main.rs`. So **every node reports the
same coordinates regardless of what was configured**, and any consumer doing
something spatial — a map, a localiser, a per-room view — is working from
constants.

The failure is quiet. Nothing errors, nothing warns, and the API returns a
well-formed position that happens to be the same for every node.

## The fix

Store the parsed positions on `AppStateInner`, keyed by 0-based index
(`= node_id`), and read from it at both call sites. 19 insertions, 3 deletions,
one file. Behaviour is unchanged when `--node-positions` is not supplied: the
same fallback is used, just no longer unconditionally.

## On the linked issues

I do not want to overclaim. #249 reports "detection window always shows the
same thing regardless of position or amount of people in the room", which is
the symptom this defect produces, but I cannot confirm from here that it is
that reporter's root cause — several things could present that way. #301's
"Known Issues" page covers the same symptom class. Offered as consistent-with,
not as a diagnosis.

## Testing

Built and tested against `main` with `--no-default-features`; the sensing-server
suite passes. The change is read-path only and adds no new dependency.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -1417,6 +1417,9 @@ struct AppStateInner {
     last_tracker_instant: Option<std::time::Instant>,
     /// Attention-weighted multi-node CSI fusion engine.
     multistatic_fuser: MultistaticFuser,
+    /// Node positions parsed from --node-positions, keyed by 0-based index
+    /// into the parsed list (which corresponds to node_id).
+    node_positions_config: HashMap<u8, [f32; 3]>,
     /// Governed trust-path bridge (ADR-135..146): runs the same live frames
     /// through the privacy/provenance/witness control plane. Does not alter
     /// person-count behavior; its trust state (witness, effective class,
@@ -1678,6 +1681,7 @@ impl AppStateInner {
             pose_tracker: PoseTracker::new(),
             last_tracker_instant: None,
             multistatic_fuser: MultistaticFuser::new(),
+            node_positions_config: HashMap::new(),
             engine_bridge: engine_bridge::EngineBridge::new(
                 wifi_densepose_bfld::PrivacyMode::PrivateHome,
                 1,
@@ -6526,7 +6530,11 @@ async fn udp_receiver_task(
                         .map(|(&id, n)| NodeInfo {
                             node_id: id,
                             rssi_dbm: n.rssi_history.back().copied().unwrap_or(0.0),
-                            position: [2.0, 0.0, 1.5],
+                            position: s
+                                .node_positions_config
+                                .get(&id)
+                                .map(|p| [p[0] as f64, p[1] as f64, p[2] as f64])
+                                .unwrap_or([2.0, 0.0, 1.5]),
                             amplitude: vec![],
                             subcarrier_count: 0,
                             // Vitals-only path; still expose the sync snapshot
@@ -7010,7 +7018,11 @@ async fn udp_receiver_task(
                         .map(|(&id, n)| NodeInfo {
                             node_id: id,
                             rssi_dbm: n.rssi_history.back().copied().unwrap_or(0.0),
-                            position: [2.0, 0.0, 1.5],
+                            position: s
+                                .node_positions_config
+                                .get(&id)
+                                .map(|p| [p[0] as f64, p[1] as f64, p[2] as f64])
+                                .unwrap_or([2.0, 0.0, 1.5]),
                             amplitude: if suppress_raw {
                                 vec![]
                             } else {
@@ -8581,7 +8593,7 @@ async fn main() {
     // threaded into `engine_bridge` so both fusion paths honor the same
     // WDP_TDM_SLOTS/WDP_GUARD_INTERVAL_US-derived guard (#1049/#1057).
     let mut engine_bridge_multistatic_cfg: Option<MultistaticConfig> = None;
-
+    let mut node_positions_config: HashMap<u8, [f32; 3]> = HashMap::new();
     let state: SharedState = Arc::new(RwLock::new(AppStateInner {
         latest_update: None,
         rssi_history: VecDeque::new(),
@@ -8672,6 +8684,9 @@ async fn main() {
                         "Configured {} node positions for multistatic fusion",
                         positions.len()
                     );
+                    for (idx, p) in positions.iter().enumerate() {
+                        node_positions_config.insert(idx as u8, *p);
+                    }
                     fuser.set_node_positions(positions);
                 }
             }
@@ -8681,6 +8696,7 @@ async fn main() {
             });
             fuser
         },
+        node_positions_config,
         engine_bridge: engine_bridge::EngineBridge::new(
             wifi_densepose_bfld::PrivacyMode::PrivateHome,
             1,
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -1417,9 +1417,13 @@ struct AppStateInner {
     last_tracker_instant: Option<std::time::Instant>,
     /// Attention-weighted multi-node CSI fusion engine.
     multistatic_fuser: MultistaticFuser,
-    /// Node positions parsed from --node-positions, keyed by 0-based index
-    /// into the parsed list (which corresponds to node_id).
-    node_positions_config: HashMap<u8, [f32; 3]>,
+    /// Node positions parsed from --node-positions, in the same order given
+    /// to `multistatic_fuser.set_node_positions`. Real fleets use
+    /// non-sequential logical node IDs, so this is NOT keyed by node_id —
+    /// entry `i` applies to the i-th smallest currently-active node_id, the
+    /// same convention `MultistaticFuser::fuse` uses. See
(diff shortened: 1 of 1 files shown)


ruvnet:
The bug is real — `position: [2.0, 0.0, 1.5]` is hardcoded at both live-data JSON-builder call sites, ignoring the parsed `--node-positions` config.

But I think the fix has a real indexing mismatch: it stores `node_positions_config: HashMap<u8, [f32;3]>` keyed by the position list's *index* (0, 1, 2, ...) and looks it up at read time by the node's actual `node_id`. The existing consumer of this same parsed list, `MultistaticFuser::fuse()` in `multistatic.rs`, treats `node_positions[i]` as purely positional — matched to per-frame iteration order, not to node ID. Real fleets use non-sequential logical IDs (11, 12, 13 — see the C6 firmware PRs open right now), so this fix's `HashMap<node_id, position>` lookup will usually miss and silently fall through to the same `[2.0, 0.0, 1.5]` fallback for every node. It likely only actually works when node IDs happen to be 0, 1, 2, ....

Worth either keying by list-index to match `multistatic.rs`'s existing convention, or redefining the CLI arg to accept explicit `node_id:x,y,z` triplets — either way, a test with non-sequential node IDs would catch this class of bug going forward. Happy to be wrong here if I'm missing something about how `--node-positions` is meant to align with node IDs — let me know if there's a convention I'm not seeing.

clonea1:
You're right, and there's no convention you're missing — the mismatch is real and it's mine. Confirmed by reading it back:

```rust
// population — keyed by LIST INDEX
for (idx, p) in positions.iter().enumerate() {
    node_positions_config.insert(idx as u8, *p);
}

// lookup — by NODE_ID
s.node_positions_config.get(&id)
```

`parse_node_positions()` returns a positional `Vec<[f32; 3]>`, so on any fleet whose IDs aren't `0, 1, 2, ...` the lookup misses and falls straight back to the `[2.0, 0.0, 1.5]` this PR exists to delete. It passes here only because our nine boards happen to be `0..8`. Worse than not fixing it, since it now looks fixed.

**Fix: your option (b), with (a) kept for compatibility.** Parse `node_id:x,y,z` when a colon is present, otherwise fall back to positional-by-index. Existing invocations keep working, and the keyed form removes the ambiguity rather than relocating it — under a purely positional scheme the binding of position to board is only ever implied by argument order, which is the fragility that produced this bug.

**One scoping decision worth stating.** `MultistaticFuser::set_node_positions()` consumes the list positionally, matched to per-frame iteration order. I'm deliberately *not* changing that here: this PR is about the `NodeInfo` JSON output, which is genuinely `node_id`-keyed, and silently altering multistatic fusion semantics inside it would be the wrong scope. So the fuser keeps receiving the same positional `Vec` it does today, and the id-keyed map is used only for the `NodeInfo` lookup. That the fuser binds position to iteration order at all is a real and separate concern — worth its own issue, and I'd rather raise it than smuggle it in here.

---

**Secondary finding, and it bears directly on the test you suggested.**

Testing with non-sequential IDs is the right regression, and I went to set it up on real hardware — the fleet is remotely provisionable, so re-homing the nine boards from `0..8` to `0,2,4,6,8,10,12,14,16` gives non-sequential IDs *and* two-digit ones in one pass, with a second run holding a node at `18` to cover a non-zero start.

That test would have quietly measured nothing. `fusion.rs`:

```rust
pub const MAX_NODES: usize = 16;
...
if (node_id as usize) < MAX_NODES && e.nodes & bit == 0 {
```

`16` and `18` are both outside the bound, so those boards are counted in the totals but vanish from every pair statistic — and the doc comment says exactly that. There's no counter and no log for it, unlike the adjacent capacity failure, which has a test literally named `"overflow is counted, not silent"`. By that standard this one is a defect, not a documented limitation.

I'll raise it separately rather than bundling it here — different file, different bug, no dependency on this PR, and it affects anyone running more than sixteen nodes whether or not the position bug exists. That PR should be more than a constant bump: raise `MAX_NODES` to 32 (exactly the ceiling of the `1u32 << ...` mask), add a counter surfaced on `/api/v1/fusion`, and log once per offending ID, so the next person to exceed it finds out immediately.

Flagging it here mainly so nobody spends an afternoon on a 16/18 test that reports a clean pass while measuring nothing.

---

**Testing status, so it's not overstated:** the unit test with non-sequential IDs (11, 12, 13) needs no hardware and lands with the fix. The re-homing validation above is queued behind a fleet experiment currently in flight that I don't want to disturb — it needs its own window and a server restart. I'll post results when I have them rather than claim them now.

ruvnet:
Pushed a fix (9b595b8) for the indexing issue: the position list is now stored raw (in the same order given to \`MultistaticFuser::set_node_positions\`) and resolved at read time by ranking currently-active node_ids ascending — matching exactly how \`fuse()\` itself maps \`node_positions[i]\` to frames. Added \`node_positions_by_active_id()\` (shared by both call sites) plus regression tests for non-sequential node IDs (11/12/13), stale-node exclusion from rank assignment, and the missing-config fallback. Should now work correctly for real fleets, not just coincidental 0/1/2 node IDs. Thanks for the original fix — the bug you caught was real, this just tightens the lookup semantics.

clonea1:
Thanks for landing `9b595b83` — I had a fix in flight for the same diagnosis and stopped rather than force-push over yours. Mine is on `alt/server-node-positions-explicit-id` if it is useful; I am not proposing it as-is, because your consistency argument is one mine does not address.

Two things worth weighing before this settles, then a way the two compose.

**1. Rank is not stable under fleet membership changes.**

`node_positions_by_active_id()` ranks over nodes seen within 10 s. So when any node goes quiet for longer than that, every higher-id node's rank shifts down by one and silently adopts a neighbour's coordinates.

Nine nodes at ids 0-8 with positions P0..P8. Node 0 reboots:

    active = {1..8} -> ranks 0..7
    node 1 -> P0, node 2 -> P1, ... node 8 -> P7

Every remaining node reports the wrong position, and P8 goes unused. Nothing logs it.

This is not hypothetical here. Rolling firmware across our nine boards this morning took each node offline about 30 s in turn, and #1795 exists because a node can be gone for **hours** — that outage was 7.5 h. A reboot is also precisely when someone is likely to be looking at where the nodes are.

**2. I do not think the ranks are guaranteed to agree with fusion either.**

The goal is to match `MultistaticFuser::fuse`, whose input is built by `node_frames_from_states_with_guard()`. That filters `node_states` by frame presence, sorts ascending by node_id, and then applies a **guard-interval filter** (`multistatic_bridge.rs:148`) that can drop further nodes. `node_positions_by_active_id()` instead filters on a 10 s recency window.

Different predicates over the same map, so the two sets can differ — and they differ exactly for a marginal node, which is when a wrong position matters most. Correct me if the guard filter cannot actually drop a node that passes the 10 s test; I traced it rather than ran it.

**3. They compose, and I think the combination is stronger than either.**

Accept an optional explicit id per entry, and keep rank as the fallback:

    x,y,z;...            positional -> rank (your behaviour, unchanged)
    node_id:x,y,z;...    explicit   -> that node, always

Coordinates never contain a colon, so the split is unambiguous, and existing invocations keep working untouched. An operator who writes `13:2.5,4.0,1.5` gets a binding that survives reboots, guard-interval drops and re-numbering; anyone who does not care keeps exactly what you shipped.

The underlying issue is that `set_node_positions()` takes a bare `Vec` and the identity lives only in the ordering, so every consumer has to re-derive it and they can disagree. Explicit ids at the CLI is the cheap half of fixing that; making the fuser id-aware is the other half and clearly its own change.

Happy to rebase mine onto yours as an additive commit if you want the explicit form, or to drop it entirely if you would rather keep the single convention. Your call — you have the better view of what else consumes that list.

The regression tests may be worth taking either way: non-sequential ids (11,12,13), a fleet re-homed to `0,2,4,...,16` with two-digit ids, a fleet not starting at zero, and one asserting the list indices are *not* keys.

clonea1:
One clarification before anyone spends time on that branch, because I under-sold a real weakness in it.

`alt/server-node-positions-explicit-id` falls back to **list index** for entries given without an explicit id. So on a fleet numbered 11,12,13 running the existing `--node-positions "a;b;c"` syntax, that branch changes nothing — same missed lookup, same `[2.0, 0.0, 1.5]`. It only helps once the config is rewritten with explicit ids.

Yours does not have that gap: rank-keying fixes the existing syntax with no operator action at all. That is a straightforward advantage of `9b595b83` over my branch taken on its own, and I should have said so plainly rather than leaving it implicit.

Which is why the composition in my previous comment used **your rank as the fallback**, not my index fallback:

    x,y,z;...            positional -> rank      (your behaviour, no config change needed)
    node_id:x,y,z;...    explicit   -> that node (stable across reboots and re-numbering)

That combination has no weak case. Taken alone, mine does.

So: take the explicit-id parsing as an additive commit on top of yours if you want the stable form available, and please do not take my branch as a replacement — the fallback is wrong in it.

PR #1793, opened 2026-09-04
fix(sensing-server): scale smoothing to the measured CSI rate, and debounce the room

clonea1:
## What is wrong

The smoothing constants are documented against an assumed frame rate:

    /// Debounce frames required before state transition (at ~10 FPS = ~0.4s).
    const DEBOUNCE_FRAMES: u32 = 4;
    /// EMA alpha for motion smoothing (~1s time constant at 10 FPS).
    const MOTION_EMA_ALPHA: f64 = 0.15;

`smooth_and_classify_node()` runs once per incoming frame. ESP32-C6 boards
measured in the field deliver **48-50 FPS**, roughly 5x the assumption. So the
intended ~0.4 s debounce is really ~0.08 s, and the ~1 s motion EMA is ~0.2 s.

At that point per-frame CSI noise passes almost straight through to the reported
`motion_level`, which is visible as rapid label flicker in the UI.

Two commits:

**1. Derive the thresholds from the node's own measured rate.** Rather than
hardcode a different fixed FPS — which just relocates the bug to whatever chip
or capture mode diverges next — each frame's effective alpha and frame-count
threshold come from that node's `csi_fps_ema`, which is already tracked and
burst-filtered.

**2. Debounce the room-level classification too.** `fuse_room` recomputes its
plurality vote every cycle with no memory of its own, unlike each node's
individually debounced classification. With per-node confidences near the
boundary — 44%, 83% and 53% observed simultaneously across three nodes —
whichever two of three agreed on a given cycle could flip the room result while
every node's own reading was stable. Everything gating on presence then flapped,
and each of those looked like a separate bug.

Both thresholds are expressed as durations rather than cycle counts, for the
same reason: a count means a different real interval at every rate.

## Testing

Built and tested against `main` with `--no-default-features`; the sensing-server
suite passes.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -2639,13 +2639,43 @@ fn raw_classify(score: f64) -> String {
 }
 
 /// Debounce frames required before state transition (at ~10 FPS = ~0.4s).
+/// Used only by [`smooth_and_classify`] (simulated / WiFi-scan sources,
+/// which run on a fixed tick cadence close to this design point).
 const DEBOUNCE_FRAMES: u32 = 4;
 /// EMA alpha for motion smoothing (~1s time constant at 10 FPS).
+/// Used only by [`smooth_and_classify`]; see above.
 const MOTION_EMA_ALPHA: f64 = 0.15;
 /// EMA alpha for slow-adapting baseline (~30s time constant at 10 FPS).
+/// Used only by [`smooth_and_classify`]; see above.
 const BASELINE_EMA_ALPHA: f64 = 0.003;
 /// Number of warm-up frames before baseline subtraction kicks in.
+/// Used only by [`smooth_and_classify`]; see above.
 const BASELINE_WARMUP: u64 = 50;
+
+/// Debounce duration required before [`smooth_and_classify_node`] accepts a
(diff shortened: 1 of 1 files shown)


ruvnet:
Verified real: \`DEBOUNCE_FRAMES\`/\`MOTION_EMA_ALPHA\`/etc. are hardcoded at an assumed 10 FPS, but C6 boards actually run ~48-50 FPS per the already-tracked \`csi_fps_ema\` — so these windows silently compress to ~1/5 their intended duration. Converting to time-constants with alpha derived via \`α = 1 - exp(-dt/τ)\` is the correct approach. The independent room-level debounce finding (\`fuse_room\`'s plurality vote had no hysteresis, so near-boundary confidences could flip room state every cycle) is also real and well-tested — six new tests covering flip-suppression, sustained-candidate flip, hold-timer reset, fast-path, and \`unavailable\`→absent.

One thing to flag before merge: this PR and #1791 both add new fields to \`AppStateInner\` and modify the *same two call sites* in \`udp_receiver_task\` within a few lines of each other (#1793 replaces \`classification_from_room(...)\` with \`debounce_room_classification(...)\` at exactly the lines #1791 also touches). Merging one first will shift line numbers under the other and likely break its patch context — worth sequencing these (merge one, rebase the other) rather than merging in parallel.

KOMATOAST86:
finish installing ruview

LalitaSuthimoon9 (approved):

PR #1794, opened 2026-09-04
fix(ui): never present invented data as measured

clonea1:
## What is wrong

On a dropped socket the client falls back to a built-in simulation and renders
synthetic values that are **indistinguishable from live ones**. The dashboard
flip-flops between `absent` and `present_still` after navigating away and back,
which looks like a sensing fault and is not one.

For a sensing product this is the most damaging failure mode available. Every
downstream judgement — where to move a node, whether a change helped, whether
anyone is home — is then made against fiction, with nothing on screen to say so.

## Changes

- **Client-side simulation is default-OFF** and must be opted into explicitly.
- **An unreachable server retries** instead of substituting data.
- **A banner states plainly when what is displayed is not live**, and the labels
  say what they mean: `NO DATA - SERVER UNREACHABLE` for an unreachable server,
  and `INVENTED DATA - NOT MEASURED` for simulation. The previous
  `OFFLINE - CLIENT SIMULATION` read as a status rather than a warning.

Also fixes a reconnect race in the same path: two in-flight connects could each
install handlers, so a stale socket kept overwriting fresher state. That race is
what made the flip-flop intermittent and hard to attribute.

## Note on scope

Simulation is still available, deliberately — it is useful for demos and for
development without hardware. The change is that it must be chosen, and that it
announces itself while running.

## Testing

All changed files parse as ES modules (`node --check`); no dangling imports.


ruvnet (commented):
Dream exact-head review — **INCONCLUSIVE**

Frozen hypothesis: with client simulation disabled by default, every unreachable or stale path emits zero fabricated sensing frames, while reconnect races cannot create parallel sockets or mislabel data as live.

The source changes are directionally sound: simulation is explicit opt-in, stale/unreachable states are visible, and epoch/in-flight guards address overlapping connects. However, this PR changes seven runtime UI files and no test file, and all exact-head workflows currently require authorization.

Acceptance gate:
1. Run exact-head UI unit and browser tests.
2. Prove the default, `?simulate=0`, `?simulate=1`, and local-storage opt-in cases.
3. Exercise two concurrent `start()` calls, stop-during-ticket-mint, a superseded socket closing, reconnect after the retry ceiling, and failed/unauthorized status probes.
4. Assert zero generated frames and a persistent stale/unreachable banner whenever simulation is not explicitly enabled.
5. Preserve build, accessibility, and security gates.

No hardware-accuracy claim is accepted by this review.

ruvnet:
This is a real and, honestly, fairly serious bug for a project whose CLAUDE.md explicitly requires "Never present WiFi sensing as camera-grade" and MEASURED/CLAIMED/SYNTHETIC tagging — a dropped socket silently falling back to fabricated data indistinguishable from live readings is exactly the failure mode that rule exists to prevent.

The fix closes it properly, not just cosmetically: simulation now defaults OFF (opt-in only via \`?simulate=1\`/localStorage), an unreachable server shows \`NO DATA — SERVER UNREACHABLE\` and retries indefinitely instead of fabricating, and the new \`dataSourceBanner\` is global and \`aria-live=polite\` so non-live state is visible everywhere, not just the one panel that had it before. The \`_connectInFlight\`/\`_connectEpoch\` guards look like a real, correctly-reasoned fix for what was likely the actual root cause of the reported flip-flop (a double-connect race), not just a bandaid on the symptom.

Only soft note: the testing claim ("node --check, no dangling imports") doesn't cover the new banner or the race-condition fix with an actual functional test — low risk given this is straightforward guard-clause logic, but a quick test asserting the banner appears/disappears correctly would make this more durable. Not a blocker from me.

PR #1795, opened 2026-09-04
fix(firmware): retry WiFi forever, and add an uplink watchdog (stacked on #1797)

clonea1:
> **Stacked on [#1797](https://github.com/ruvnet/RuView/pull/1797) — please read
> that one first.** This branch sits on top of the thermal work because they
> share `main.c` and `Kconfig.projbuild`, so the diff below currently includes
> #1797's commits as well. Once #1797 merges, this reduces to the WiFi change
> alone. I raised these in the wrong order; apologies for the noise.
>
> **The change this PR is actually about is one commit: `eaca8c17`.**

## What is wrong

`main.c:62` caps reconnection attempts:

    #define MAX_RETRY 10

After ten failures the node stops trying. It stays powered, the LED behaves, and
it never reconnects — so an AP reboot, a brief outage or a channel change
permanently removes a node from the fleet until someone power-cycles it by hand.

For a board mounted on a wall or ceiling, "give up permanently" is the wrong
default. There is nobody at the console to notice.

Measured in this deployment: a **7.5 hour** fleet outage traced to exactly this,
after the APs restarted. The nodes were healthy the entire time and simply never
retried.

## Changes

- **Retry indefinitely with backoff** rather than stopping at ten. The node is
  useless while disconnected, so there is no cost to continuing, and the backoff
  keeps it from hammering an AP that is still booting.
- **An uplink watchdog**: if frames stop being accepted for long enough, the
  node restarts itself rather than sitting in a state that looks alive from the
  outside but delivers nothing.

## Testing

### DIFF
--- a/firmware/esp32-csi-node/main/CMakeLists.txt
+++ b/firmware/esp32-csi-node/main/CMakeLists.txt
@@ -1,5 +1,6 @@
 set(SRCS
-    "main.c" "csi_collector.c" "stream_sender.c" "nvs_config.c"
+    "main.c" "csi_collector.c"
+    "thermal.c" "stream_sender.c" "nvs_config.c"
     "edge_processing.c" "ota_update.c" "power_mgmt.c"
     "wasm_runtime.c" "wasm_upload.c" "rvf_parser.c"
     "mmwave_sensor.c"
--- a/firmware/esp32-csi-node/main/Kconfig.projbuild
+++ b/firmware/esp32-csi-node/main/Kconfig.projbuild
@@ -517,3 +517,56 @@ menu "Onboard LED (ADR-183)"
             colormap, in milli-units (250 = 0.25). Lower = more sensitive
             (reaches yellow with less motion).
 endmenu
+
+menu "Thermal monitoring"
+
+config THERMAL_MONITOR
+    bool "Read the on-chip temperature sensor and throttle the radio"
+    depends on IDF_TARGET_ESP32C6
+    default y
+    help
+      The C6 has a temperature sensor this firmware otherwise never reads.
+      Harmless on a bare bench board; not once a node is sealed in a printed
+      enclosure on a wall with the radio transmitting continuously.
+
+      What is being protected is the ENCLOSURE, not the silicon. The part is
+      rated far past anything it sees here, but PLA softens at its glass
+      transition (~55-60 C) and sags slowly under its own weight, which never
+      announces itself.
+
+config THERMAL_WARN_C
+    int "Warn above this die temperature (C)"
+    depends on THERMAL_MONITOR
+    range 40 100
+    default 65
+
+config THERMAL_THROTTLE_C
+    int "Reduce transmit power above this die temperature (C)"
+    depends on THERMAL_MONITOR
+    range 45 105
+    default 72
+    help
+      Die temperature, not enclosure temperature -- the sensor is on the chip
+      and runs hotter than the plastic around it. ~72 C on the die in a small
+      vented box is roughly a 50-55 C enclosure, which is where PLA begins to
+      matter. Setting this at 55 would throttle a perfectly healthy node.
+
+      PETG (glass transition ~80 C) removes the concern; raise this if the
+      whole fleet is printed in it.
+
+config THERMAL_CRITICAL_C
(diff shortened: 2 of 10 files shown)


ruvnet:
This is the strongest of today's firmware batch. The real, measured 7.5-hour fleet outage (2026-08-30) is solid motivation for unbounded retry with backoff, and decoupling \`WIFI_BOOT_WAIT_ATTEMPTS\` (boot-wait release only) from actual reconnection (now truly unbounded) is the right fix. The watchdog design is careful: it fires only on local \`sendto()\` success/failure, explicitly not remote delivery, so a dead server won't reboot the fleet — and it only arms after the first successful send, avoiding a reboot-loop while an AP is legitimately down. The 1200s timeout is derived from real soak-test percentiles (median 225s, max 510s, 2.35x margin) rather than picked arbitrarily — that's the kind of evidence this repo's CLAUDE.md asks for.

Flagging a real conflict: #1788 (open, reviewed earlier) independently touches the exact same \`MAX_RETRY\` line, bumping it 10→90 — a materially simpler, still-bounded approach. This PR's unbounded+backoff+watchdog design is more thorough and evidence-based; worth treating this as the preferred direction over #1788's \`MAX_RETRY=90\` bump when deciding which lands, to avoid merging two different fixes for the same problem.

clonea1:
Agreed on the #1788 conflict, and thanks for flagging it — two fixes for one bug is the worst outcome here.

For the record on why this is the direction I'd argue for: `MAX_RETRY = 90` is still bounded, so it does not fix the failure, it postpones it. `s_retry_num` is only ever cleared on a successful `IP_EVENT_STA_GOT_IP`, so it accumulates across a node's entire uptime — 90 cumulative disconnects over weeks still ends with `WIFI_FAIL_BIT` set and `esp_wifi_connect()` never called again, recoverable only by a power cycle. The 2026-08-30 outage was 7.5 hours and ended with someone physically unplugging boards.

The two halves matter separately: unbounded retry with backoff fixes the reconnect path, and the uplink watchdog covers the failures we have *not* diagnosed. The watchdog deliberately keys on local `sendto()` success rather than remote delivery, so stopping the aggregator cannot reboot the fleet, and it arms only after the first successful send so a node powered up while the AP is down does not reboot-loop.

Happy for #1788 to be closed in favour of this, or to rebase this on top of it if you would rather land the smaller change first.

One sequencing note: this PR is stacked on #1797, and I am about to push a hysteresis fix there (the `CRITICAL -> THROTTLED` edge is currently undamped). I will rebase this branch once that lands, so it is worth merging #1797 first.

clonea1:
Two updates.

**1. This PR carries hardware evidence that its body does not mention.**

Raised on #1797 as being in that PR; it is actually here. `eaca8c17` reworks `thermal_init()` because the sensor range failed on silicon:

```c
/* ... (-10,110) looks reasonable and is invalid for exactly that reason;
 * it failed on real hardware. */
static const struct { int lo, hi; } spans[] = {
    {20, 100}, {-10, 80}, {-30, 50}, {50, 125},
};
```

`temperature_sensor_install()` walks a table of predefined spans and takes the first that fully contains the requested range, so a span crossing two boundaries returns `ESP_ERR_INVALID_ARG` rather than being clamped. `(-10, 110)` reads as reasonable and is rejected. The fallback list means an unheated room, or a chip revision with a different table, degrades to a usable range instead of leaving the node silently unmonitored.

That is a real finding from real hardware and it belongs in the PR description rather than buried in a comment — I will surface it there. (The confusion is presumably because this PR is stacked on #1797, so its diff shows both sets of commits.)

**2. Rebased onto the updated #1797.**

I have pushed a hysteresis fix to #1797 — the `CRITICAL -> THROTTLED` edge you would expect to be damped was unconditional, so a node on the critical boundary flapped `TXP_MIN <-> TXP_STEP2`. This branch is rebased on top of that, so the stack is consistent again. The rebase touched only the thermal files; the retry/watchdog payload here (`main.c`, `stream_sender.c/h`) is byte-identical, verified by diffing the rebased branch against a pre-rebase backup ref.

Still happy either way on the #1788 question — this can be rebased onto that instead if you would rather land the smaller change first.

Issue #762, opened 2026-05-22
ADR-110: ESP32-C6 firmware extension (Wi-Fi 6 CSI, 802.15.4 mesh, TWT, LP-core)


ruvnet:
Tracking issue for ADR-110 — extending `firmware/esp32-csi-node` with a parallel ESP32-C6 build target that exploits four C6-only capabilities not available on the production S3 node.

## Why

Confirmed on hardware (COM6, ESP32-C6 rev v0.2, MAC `20:6e:f1:17:27:8c`):

- **Wi-Fi 6 / 802.11ax CSI** — HE-LTF gives 242 subcarriers per HE20 frame vs 52 for HT-LTF; no prior work has published WiFi-6-CSI human pose
- **802.15.4 mesh time-sync** — cross-node clock alignment over a separate radio, frees the WiFi channel for sensing (relevant to ADR-029/030 multistatic)
- **TWT (Target Wake Time)** — deterministic CSI cadence negotiated with the AP, replaces today's opportunistic capture
- **LP-core + ~5 µA hibernation** — always-on motion gate for battery-powered Cognitum Seed nodes

S3 production path stays untouched — every C6 module is `#ifdef CONFIG_IDF_TARGET_ESP32C6` gated.

## Design

ADR-110 in `docs/adr/ADR-110-esp32-c6-firmware-extension.md` — covers the dual-target build, the ADR-018 frame extension for HE PPDU metadata, the 802.15.4 beacon-based time-sync protocol, TWT negotiation, and LP-core hibernation.

## Phases

- [ ] **P1** Multi-target build support (`sdkconfig.defaults.esp32c6`, partition selection)
- [ ] **P2** HE-LTF CSI tagging in `csi_collector.c`
- [ ] **P3** TWT setup helper (`c6_twt.c`)
- [ ] **P4** 802.15.4 init + skeleton time-sync (`c6_timesync.c`)
- [ ] **P5** LP-core hibernation stub (`c6_lp_core.c` + `lp_core/main.c`)
- [ ] **P6** Build, flash COM6, capture boot telemetry, S3 regression SHA check
- [ ] **P7** Benchmark C6 vs S3 (CSI fps, RAM, TWT jitter, power)
- [ ] **P8** Witness bundle update, CLAUDE.md / README hardware tables

Each phase will be reported as a comment on this issue and reflected back into ADR-110.

## Related

- ADR-018 (CSI binary frame format) — being extended
[...]

PR #1796, opened 2026-09-04
perf(firmware): stop the C6 sharing its radio with a stack that cannot receive

clonea1:
## What is wrong

The ESP32-C6 has a single 2.4 GHz radio shared between WiFi and 802.15.4 by
coexistence time-slicing. The 802.15.4 stack is started, but nothing in this
firmware receives on it — so every node pays coexistence overhead for a
capability it never uses.

That overhead comes directly out of CSI capture, which is the entire job of the
device.

## The fix

Disable the unused 802.15.4 stack in `sdkconfig.defaults.esp32c6`. Config only,
19 lines, no source change.

## Scope

C6-specific and confined to the C6 defaults file. S3 builds are untouched, since
they have no shared-radio coexistence to begin with.

## Testing

### DIFF
--- a/firmware/esp32-csi-node/sdkconfig.defaults.esp32c6
+++ b/firmware/esp32-csi-node/sdkconfig.defaults.esp32c6
@@ -77,3 +77,22 @@ CONFIG_FREERTOS_TIMER_TASK_STACK_DEPTH=8192
 # ── Power: keep CPU at max 160 MHz (C6 ceiling) for DSP throughput ──
 CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ_160=y
 CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ=160
+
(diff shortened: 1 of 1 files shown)


ruvnet:
Verified: the "safe to disable" claim holds — the ADR-018 sync flag is OR'd across both transports and this change is \`#ifdef\`-gated to C6 only, so it can't regress S3/other targets. Minimal (19 lines, config-only), well-cited (ADR-110 D1 / issue #762, with the real 99.56% ESP-NOW delivery figure backing "the actual sync path doesn't need this radio"). No hardware boot log needed for a change this scoped — looks correct as-is.

clonea1:
Thanks for the review — agreed, nothing to change here.

Confirming the two points that mattered: the change is `#ifdef`-gated to C6 only so it cannot regress S3 or other targets, and the ADR-018 sync flag is OR'd across both transports, so disabling the 802.15.4 radio cannot remove the sync path — ESP-NOW carries it at the measured 99.56% delivery.

Leaving this one as-is.

PR #1797, opened 2026-09-04
feat(firmware): thermal monitoring, radio throttling and node health

clonea1:
## Why

There is no thermal monitoring at all today. A node that overheats behaves in a
way that is easy to misdiagnose: the SoC reduces transmit power, RSSI falls at
**both** ends of every link, and the result is indistinguishable from an
obstruction or a node that has drifted out of position.

Without a temperature to look at, the natural response is to go and move the
hardware — which is the wrong action, and the symptom returns.

## What this adds

- **`thermal.c` / `thermal.h`** — die temperature, a four-level state
  (ok / warn / throttled / critical), and radio throttling when hot.
- **Reset reason and heap health**, reported alongside. A node that reboots
  tells you almost nothing; one that reboots with `Brownout` tells you its
  supply sagged, which on a shared outlet is a wiring fault rather than a
  firmware one.
- **A refactor separating measurement from action**, so reading the temperature
  and deciding what to do about it are not the same function. That is what makes
  the state testable.

334 insertions across 6 files, plus Kconfig entries so it can be compiled out.

## Relationship to the WiFi retry PR

[#1795](https://github.com/ruvnet/RuView/pull/1795) builds directly on this — it
shares `main.c` and `Kconfig.projbuild` — so its diff currently shows these
commits too. **This PR is the base; #1795 should be read after it.** I raised
them in the wrong order and have annotated #1795 accordingly.

## Testing

### DIFF
--- a/firmware/esp32-csi-node/main/CMakeLists.txt
+++ b/firmware/esp32-csi-node/main/CMakeLists.txt
@@ -1,5 +1,6 @@
 set(SRCS
-    "main.c" "csi_collector.c" "stream_sender.c" "nvs_config.c"
+    "main.c" "csi_collector.c"
+    "thermal.c" "stream_sender.c" "nvs_config.c"
     "edge_processing.c" "ota_update.c" "power_mgmt.c"
     "wasm_runtime.c" "wasm_upload.c" "rvf_parser.c"
     "mmwave_sensor.c"
--- a/firmware/esp32-csi-node/main/Kconfig.projbuild
+++ b/firmware/esp32-csi-node/main/Kconfig.projbuild
@@ -517,3 +517,56 @@ menu "Onboard LED (ADR-183)"
             colormap, in milli-units (250 = 0.25). Lower = more sensitive
             (reaches yellow with less motion).
 endmenu
+
+menu "Thermal monitoring"
+
+config THERMAL_MONITOR
+    bool "Read the on-chip temperature sensor and throttle the radio"
+    depends on IDF_TARGET_ESP32C6
+    default y
+    help
+      The C6 has a temperature sensor this firmware otherwise never reads.
+      Harmless on a bare bench board; not once a node is sealed in a printed
+      enclosure on a wall with the radio transmitting continuously.
+
+      What is being protected is the ENCLOSURE, not the silicon. The part is
+      rated far past anything it sees here, but PLA softens at its glass
+      transition (~55-60 C) and sags slowly under its own weight, which never
+      announces itself.
+
+config THERMAL_WARN_C
+    int "Warn above this die temperature (C)"
+    depends on THERMAL_MONITOR
+    range 40 100
+    default 65
+
+config THERMAL_THROTTLE_C
+    int "Reduce transmit power above this die temperature (C)"
+    depends on THERMAL_MONITOR
+    range 45 105
+    default 72
+    help
+      Die temperature, not enclosure temperature -- the sensor is on the chip
+      and runs hotter than the plastic around it. ~72 C on the die in a small
+      vented box is roughly a 50-55 C enclosure, which is where PLA begins to
+      matter. Setting this at 55 would throttle a perfectly healthy node.
+
+      PETG (glass transition ~80 C) removes the concern; raise this if the
+      whole fleet is printed in it.
+
+config THERMAL_CRITICAL_C
+    int "Minimum transmit power above this die temperature (C)"
+    depends on THERMAL_MONITOR
+    range 50 110
+    default 80
+
+config THERMAL_HYSTERESIS_C
+    int "Degrees below the throttle point before recovering"
+    depends on THERMAL_MONITOR
+    range 1 20
+    default 8
+    help
+      Recovering at the same threshold that tripped would oscillate, and every
+      oscillation is a visible RSSI step on every link the node has.
+
+endmenu
--- a/firmware/esp32-csi-node/main/Kconfig.projbuild
+++ b/firmware/esp32-csi-node/main/Kconfig.projbuild
@@ -534,6 +534,19 @@ config THERMAL_MONITOR
       transition (~55-60 C) and sags slowly under its own weight, which never
       announces itself.
 
+config THERMAL_THROTTLE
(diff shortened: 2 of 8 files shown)


ruvnet:
Good design overall — clean measure/act separation (\`THERMAL_THROTTLE\` defaults off, monitoring-only by default), sensible 4-level FSM, and explicit hysteresis (8°C default) specifically to avoid RSSI-visible step oscillation.

One real gap: hysteresis is only applied on the *down* transitions into OK (\`c < recover\`) — the CRITICAL→THROTTLED transition is unconditional with no hysteresis, so a node sitting right at the CRITICAL boundary could flap TXP_MIN↔TXP_STEP2 without damping, which is the same class of RSSI-step problem this PR is otherwise solving for. Not a blocker (THROTTLED→OK still has to clear the WARN threshold's hysteresis fully, so the worst case is bounded), but worth tightening for consistency.

On evidence: the PR body only claims "built in the container," which on its own wouldn't clear this repo's hardware-validation bar. But there's a later commit in this same PR that fixes a real bug in \`temperature_sensor_install\`'s range table because "(-10,110) ... failed on real hardware" — that IS genuine hardware evidence, it's just not stated in the PR description. Worth updating the PR body to surface that finding explicitly so the real validation this already has is visible.

clonea1:
Thanks — the hysteresis gap is real and I'll fix it.

Confirmed: `thermal_apply()` only guards the downward edge into OK (`c < recover`); `CRITICAL -> THROTTLED` is unconditional, so a node sitting on the CRITICAL boundary can flap `TXP_MIN <-> TXP_STEP2` undamped. That is the same RSSI-visible step oscillation this PR exists to prevent, so it should be consistent. I'll apply hysteresis to every downward transition, not just the one into OK, and add a test that walks the temperature across each boundary in both directions.

**On the evidence point I have to correct the record, because it cuts the wrong way.**

There is no later commit in this PR fixing `temperature_sensor_install`'s range table. I checked both commits on the branch (`5419bff2`, `ed006abd`), and `git log --all -S TEMPERATURE_SENSOR_CONFIG_DEFAULT` across the whole repository. The call is `TEMPERATURE_SENSOR_CONFIG_DEFAULT(-10, 110)` in the original commit and it is still exactly that at head — unchanged, never fixed, and there is no commit anywhere describing a real-hardware failure of that range.

So this PR really does have **only a container build**, exactly as the body says. Per this repo's own rule — *"Hardware validation requires evidence from real silicon, normally a captured boot/runtime log. A successful build or simulator is not hardware evidence"* — it should not be treated as hardware-validated, and I'd rather say so than let it merge on evidence that does not exist.

Worth noting the code is written to survive that sensor being unavailable: a failed `temperature_sensor_install()` logs and returns, and the node continues unmonitored rather than failing to boot. So the untested path degrades safely — but that is a design argument, not hardware evidence.

I'll post a boot log from a real C6 with the thermal build before asking for this to merge.

clonea1:
**Correction to my previous comment — you were right about the hardware evidence and I was wrong.**

I claimed there was "no commit anywhere describing a real-hardware failure of that range". That is false. The fix exists, and it says exactly what you said it says:

```c
/* The requested span must fit ENTIRELY inside one of the sensor's predefined
 * measurement ranges -- the driver walks a table and takes the first that
 * contains it, so a span crossing two boundaries is rejected with
 * ESP_ERR_INVALID_ARG rather than clamped. (-10,110) looks reasonable and is
 * invalid for exactly that reason; it failed on real hardware. */
static const struct { int lo, hi; } spans[] = {
    {20, 100}, {-10, 80}, {-30, 50}, {50, 125},
};
```

**Why I missed it.** I searched with `git log --all -S TEMPERATURE_SENSOR_CONFIG_DEFAULT`. `-S` only reports commits where the *number of occurrences* of the string changes. The refactor moved the call inside a loop — one occurrence before, one after — so the count never changed and `-S` returned nothing. `-G` finds it immediately. I then stated the negative conclusion with far more confidence than a pickaxe search can support.

**What was actually right in each direction.** The commit is in **#1795**, not this PR — #1797 is two commits (`5419bff2`, `ed006abd`) and neither contains it. #1795 is stacked on #1797, so its diff carries both, which is presumably how it read as belonging here. So:

- Your substantive point stands: that range fix **is** genuine hardware evidence, and it was not surfaced in any PR body.
- The narrow claim in my last comment — that **#1797 on its own** has only a container build — still holds.
- My sweeping "no such commit anywhere" claim was wrong, and it was the load-bearing part of the correction. Apologies for the confident tone on it.

I have raised the evidence on **#1795**, where it actually belongs.

The range fix itself is better than a one-line threshold change, incidentally: the fallback table means an unheated room or a chip revision with a different span table degrades to a usable range instead of leaving the node unmonitored.

Unchanged from before: I still owe this PR a real C6 boot log, and the hysteresis fix is pushed shortly.

clonea1:
Hardware evidence, as promised earlier in this thread. Real ESP32-C6 (`e8:f6:0a:fc:bf:a0`, node 7, IDF v5.4), flashed over USB with the thermal build and captured on serial:

    I (154910) thermal: die 26.0 C (peak 26.0 C) state=ok tx=20.00 dBm heap 208772 min 184876

and read back independently over the network from the node's sync packet:

    node 7 health: die_c=30, thermal_state=0 (ok), tx_dbm=20, min_heap_kib=164

So on real silicon: the temperature sensor installs, returns a plausible die reading, the state machine sits in `ok`, and transmit power is unthrottled at 20 dBm — which is the correct behaviour well below the 65 C warn threshold.

This clears the "built in the container" gap I flagged earlier: the PR now has evidence from real hardware rather than a successful compile.

Two notes for accuracy:

- The `temperature_sensor_install()` range fix — the spans fallback table replacing the invalid `(-10, 110)` — is in **#1795**, not this PR, as established earlier in this thread. That fix is what makes the sensor install succeed at all, so the reading above depends on both PRs.
- No throttling behaviour was exercised. The board sat at 24-30 C throughout, so `THERMAL_THROTTLE` never engaged and the `CRITICAL -> THROTTLED` hysteresis fix is covered only by the host test, not by hardware. Verifying that on silicon would need a deliberately heated node, which I have not done.

Issue #1759, opened 2026-08-31
fix(calibration): fresh/bound calibration emits no `calibrated_presence_evidence` on any endpoint


proffesor-for-testing:
## Summary

A successfully completed, bound empty-room calibration produced no calibrated output.
`calibrated_presence_evidence` is computed per tick (`main.rs:8211`) and is part of
`SensingUpdate` (`types.rs:130`), but it never appeared on any endpoint on a system where
calibration reported `fresh` and `bound`.

## State at the time

```
/api/v1/calibration/status
  status: fresh          binding_mode: bound       legacy: false
  frame_count: 85731     baseline_eigenvalue_count: 6
  source_node_ids: [1,2,3]   observed_source_node_ids: [1,2,3]
  missing_source_node_ids: []
  model_receipt: ruview.calibration.field-model-receipt.v1 present
```

All three bound nodes `active` and streaming at the time of sampling.

## Observation

`calibrated_presence_evidence` was absent from the response of every endpoint checked:

```
/api/v1/sensing/latest      absent
/api/v1/status              absent
/api/v1/nodes               absent
/api/v1/pose/current        absent
/api/v1/calibration/status  absent
```

`/api/v1/sensing/latest` top-level keys were: `classification, estimated_persons, features,
node_features, nodes, persons, room_inference, signal_field, signal_quality_score, source,
tick, timestamp, type, vital_signs`.

Since the field is `Option<CalibratedPresenceEvidence>`, absence means the chain returned
`None` on every tick.

## Where it could be returning None

Not diagnosed — listing the candidates rather than guessing:

- `main.rs:1636` — `self.field_model.as_ref()?`
- `main.rs:1637` — node `frame_history` lookup
- `field_bridge.rs:146` — `field.check_freshness(observed_at_us) != Fresh`
- `field_bridge.rs:156` — empty frame window after canonical resampling
- `field_bridge.rs:168` — `extract_perturbation(...).ok()?`
- `calibration_session.rs:240-251` — receipt-side conditions (node in `source_node_ids`,
  `observed_at >= completed_at`, `person_count <= 3`, `inference_method` one of
  `field_model_eigenvalue_v1` / `field_model_perturbation_energy_v1`)

**Ruled out:** a timestamp-unit mismatch between `finalize_calibration(now.timestamp_micros())`
(`main.rs:6546`) and the `observed_at_unix_ms * 1_000` passed at `main.rs:1641`. Both are Unix
epoch microseconds; consistent.

[...]

PR #1800, opened 2026-09-04
fix(calibration): gate finalize on wall-clock duration, expose elapsed_s and frames/s

### LINKED ISSUE #1756: fix(calibration): 12,000-frame gate encodes "10 min at 20 Hz" — a faster fleet satisfies it in 90 s with no warning
## Summary

Empty-room baseline calibration is gated on a **frame count** that is documented as standing in
for **wall-clock duration**. A fleet streaming faster than the assumed 20 Hz satisfies the gate
in a fraction of the intended window, with no warning, producing a baseline that passes every
check while covering far less environmental variation than the design intends.

- `ruvsense/field_model.rs:9` — *"Collect CSI during empty-room calibration (>=10 min at 20 Hz)"*
- `ruvsense/field_model.rs:247` — *"Minimum calibration frames before baseline is valid
  (10 min at 20 Hz = 12000)"*, `min_calibration_frames: 12_000`
- every accepted node packet advances the same field model, so N nodes divide the elapsed time
  by N
- finalize checks only `frame_count`, and `/api/v1/calibration/status` exposes neither elapsed
  calibration time nor effective sample rate

The Welford statistics exist to absorb slow environmental variation — HVAC cycles, thermal
drift. A 90-second sample cannot contain that regardless of how many frames it holds.

## Why now

Measured on hardware: three ESP32-S3 nodes at ~127 frames/s aggregate reached 12,000 frames in
**90 seconds** — about 1/7th of the documented window. Nothing in the API or logs indicated the
baseline was short in time. The capture was deliberately extended to ~11 minutes, but only
because the discrepancy was noticed by reading the source.

## Reproduction

Start a calibration with enough nodes to exceed 20 Hz aggregate and poll
`/api/v1/calibration/status`:

```
t+ 0s   frames  1,099   collecting
t+30s   frames  4,862   collecting     (~127/s)
t+90s   frames 12,546   collecting     ← gate already satisfied
```

Stopping at that point yields a valid `fresh` baseline covering 90 seconds.

## Done when

A baseline cannot be finalized from materially less wall-clock time than the design intends,
or the operator is clearly told when it is.

## Work items

- [ ] Gate on elapsed wall-clock **and** frame count, not frames alone.
- [ ] Expose `elapsed_s` and effective frames/s on `/api/v1/calibration/status` so a client can
      see the shape of the capture.
- [ ] Warn on finalize when the frame gate was met in substantially less than the intended
      window.
- [ ] Confirm the intended duration in ADR-135 and make the constant derive from it rather than
      hard-coding a 20 Hz assumption.

HsukqiLee:
Fixes #1756.

## Problem

Empty-room baseline calibration is gated on a **frame count** (12,000) that encodes "10 min at 20 Hz". Every accepted node packet advances the same field model, so a fleet of N nodes satisfies the gate in ~1/N of the intended window — three nodes at ~127 frames/s aggregate reached 12,000 frames in ~90 s, with nothing in the API indicating the baseline was short in wall-clock time. The Welford statistics exist to absorb slow environmental variation (HVAC cycles, thermal drift), which a 90-second sample cannot contain regardless of frame count.

## Change

- **Gate on wall-clock duration as well as frames.** `FieldModelConfig` gains `min_calibration_duration_s` (default `CALIBRATION_DURATION_S` = 600 s). `finalize_calibration` now returns a new `FieldModelError::InsufficientCalibrationDuration` when the elapsed window is short, even if the frame target is met.
- **Frame target derived from the window, not hard-coded.** `min_calibration_frames` default is now `CALIBRATION_DURATION_S * ASSUMED_CALIBRATION_RATE_HZ` (= 12,000), replacing the bare `12_000` literal so the intent is documented in one place.
- **Track and expose the capture shape.** `FieldModel` records the first accepted frame's monotonic timestamp and adds `calibration_elapsed_s()` / `calibration_frames_per_second()` / `min_calibration_duration_s()` accessors. `/api/v1/calibration/status` now returns `elapsed_s`, `frames_per_second`, `min_frames`, and `min_duration_s`.
- **Warn when the frame gate is met early.** `POST /api/v1/calibration/stop` returns a structured `warning` (and logs at WARN) when the frame target is reached in substantially less than the intended window, instead of finalizing silently.
- `reset_calibration` clears the session clock.

Tests set `min_calibration_duration_s: 0.0` to keep the existing instant-frame suites on the legacy frame-only path; a `0.0` value disables the duration gate (tests only). New unit tests cover the derived frame target, the duration gate, and the clock accessors.

Note on work item 4: ADR-135 governs the separate `CalibrationRecorder` baseline path (30 s default). The field model here is ADR-030 territory and its own module docs state ">=10 min", so the duration constant is derived from that documented 600 s window rather than an assumed rate alone.

## Test plan

- [x] `cargo test -p wifi-densepose-signal --no-default-features --lib field_model` — 27 passed (includes 4 new wall-clock gate tests).
- [x] `cargo test -p wifi-densepose-sensing-server --bin sensing-server` — 252 passed.
- [x] `cargo check -p wifi-densepose-sensing-server` — clean.

Evidence classification: the 90 s / 127 fps figures are from the linked issue (reported, not re-measured here). The code behavior and test results above are measured on this branch.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -5822,13 +5822,31 @@ async fn calibration_stop(State(state): State<SharedState>) -> Json<serde_json::
         // caller knows to keep the room empty and poll /calibration/status.
         let have = fm.calibration_frame_count();
         let need = fm.min_calibration_frames() as u64;
-        if have < need {
-            return Json(serde_json::json!({
+        let elapsed_s = fm.calibration_elapsed_s();
+        let need_s = fm.min_calibration_duration_s();
+        if have < need || (need_s > 0.0 && elapsed_s < need_s) {
+            let mut resp = serde_json::json!({
                 "success": false,
-                "error": "Not enough calibration frames yet — keep the room empty and poll /calibration/status until frame_count reaches the target.",
+                "error": "Calibration not complete yet — keep the room empty and poll /calibration/status until both frame_count and elapsed_s reach their targets.",
                 "frame_count": have,
                 "frames_needed": need,
-            }));
+                "elapsed_s": elapsed_s,
+                "duration_needed_s": need_s,
+                "frames_per_second": fm.calibration_frames_per_second(),
+            });
+            // #1756: a fast fleet satisfies the frame gate in ~1/N of the
+            // intended wall-clock window. Make that explicit instead of
+            // silently collecting, so the operator sees why the session
+            // keeps running.
+            if have >= need && elapsed_s < need_s {
+                tracing::warn!(
+                    "Calibration frame target reached in {elapsed_s:.1}s — substantially less than the intended {need_s:.0}s window; continuing to collect for slow environmental variation (#1756)"
+                );
+                resp["warning"] = serde_json::json!(format!(
+                    "frame target reached in {elapsed_s:.1}s — substantially less than the intended {need_s:.0}s window; continuing to collect for slow environmental variation"
+                ));
+            }
+            return Json(resp);
         }
         let ts = chrono::Utc::now().timestamp_micros() as u64;
         match fm.finalize_calibration(ts, 0) {
@@ -5841,6 +5859,7 @@ async fn calibration_stop(State(state): State<SharedState>) -> Json<serde_json::
                     "baseline_eigenvalue_count": baseline,
                     "variance_explained": variance_explained,
                     "frame_count": fm.calibration_frame_count(),
+                    "elapsed_s": elapsed_s,
                 }))
             }
             // ADR-080 #2: finalize error chain stays server-side only.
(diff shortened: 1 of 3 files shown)


ruvnet:
Confirmed this genuinely fixes #1756's root cause, not just a symptom: the new wall-clock gate (\`min_calibration_duration_s\`, default 600s) rejects early finalization even when the frame-count target is hit, and the frame target itself is now derived (\`CALIBRATION_DURATION_S * ASSUMED_CALIBRATION_RATE_HZ\`) rather than a bare unexplained \`12_000\` literal. Verified the slower-than-20Hz case is unaffected — a single slow node already takes ≥600s to hit the frame target, so the new duration gate never becomes the binding constraint there; it only bites the fast-fleet case #1756 actually reported, no regression for the slow path. \`/api/v1/calibration/status\` now exposing \`elapsed_s\`/\`frames_per_second\`/\`min_frames\`/\`min_duration_s\`, plus a structured \`warning\` on early-frame-target stop, is a good honesty improvement on its own. Test coverage is solid (4 new targeted tests + the full 27 field_model + 252 sensing-server suite green).

One note for awareness, not a blocker on this PR: #1759 (bound calibration emits no \`calibrated_presence_evidence\` on any endpoint) is a separate, still-open bug — this PR doesn't fix it and doesn't appear to make it worse, just wanted to flag they're not the same issue in case anyone assumes this closes both.

v2/crates/wifi-densepose-sensing-server/src/main.rs
```diff
@@ -5822,13 +5822,31 @@ async fn calibration_stop(State(state): State<SharedState>) -> Json<serde_json::
         // caller knows to keep the room empty and poll /calibration/status.
         let have = fm.calibration_frame_count();
         let need = fm.min_calibration_frames() as u64;
```
justjojo844-art, line 5824: Big jago

justjojo844-art (approved):

v2/crates/wifi-densepose-sensing-server/src/main.rs
justjojo844-art: Great

PR #1805, opened 2026-09-04
feat(wire): carry transmitter and transmission identity in the CSI frame

clonea1:
Wire v1 identifies neither which transmitter a frame came from nor which
transmission it was. Both are needed before frames from different nodes can be
related to each other.

v2 appends the 802.11 addr2 (transmitter MAC) at bytes 20..25. A node in
promiscuous MGMT+DATA mode with no filter produces CSI for every transmitter
on the channel -- measured 2026-08-28 at ~75% non-AP in a normal home. Each
frame is a valid measurement of a DIFFERENT link, and without the transmitter
on the wire the sink interleaves them into one history with mixed geometry.

v3 appends the 802.11 receive sequence number. v2 makes a frame attributable
to a link; v3 makes it attributable to a transmission. The existing byte 12..15
carries a counter private to each node, so two nodes that captured the same
packet off the air report unrelated numbers and the sink cannot tell it was one
event observed twice. Arrival timestamps do not substitute -- the sink sees
network jitter, not the moment of capture.

This answers the blocker stated in ADR-138 (LinkGroup / ArrayCoordinator),
which is Accepted-partial with integration glue pending because the
mesh_aligned_us plumbing exists today only in the sensing server, not in a
shared FrameMeta. Per-frame identity on the wire is that missing piece.

Sequencing note: a receiver must accept v2 and v3 before a sender emits them,
so the server-side parser should land first.

Stacked on the thermal change: the ESP-NOW sync packet carries die
temperature, thermal state and TX dBm in bytes 29..31, so this file depends on
thermal.h.

---


clonea1:
Cohort map in #1956. This is the firmware half of #1806's wire change — the
node side of the transmitter MAC and `rx_seq` that ADR-345 needs for link
attribution — and it should land **after** #1806, so the server can already
parse v2/v3 before any node starts emitting it and a node never sends
something the server rejects. Touches `firmware/` only, so it merges clean
either way.

---

PR #1806, opened 2026-09-04
feat(server): parse CSI wire v2 and v3 headers

clonea1:
The frame parser accepts one magic (0xC5110001) and a fixed 20-byte header, so
a frame carrying transmitter or transmission identity is rejected outright.
That makes it impossible for a node to tell the server anything about WHICH
link a measurement belongs to, or WHICH transmission produced it.

Adds named constants for the three wire versions and dispatches on the magic:

    v1  0xC5110001  20-byte header, unchanged
    v2  0xC5110008  + transmitter MAC (addr2) at bytes 20..25
    v3  0xC511000A  + 802.11 rx_seq (u16 LE) at bytes 26..27

Esp32Frame gains source_mac and rx_seq, both Option, both None for a v1 frame.
A mixed fleet therefore keeps working: older firmware parses exactly as before
and newer firmware simply carries more.

This is the receiver half of the wire change and should land BEFORE any
firmware starts emitting v2/v3, so there is never a moment where a node sends
something the server rejects.

It is also a prerequisite for two follow-on changes that cannot work without
per-frame identity: per-link CSI state keyed on (receiver, transmitter), and
cross-node pairing keyed on (transmitter, rx_seq).

---

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -307,6 +307,29 @@ struct Esp32Frame {
     /// while the node had a valid IEEE 802.15.4 mesh-time solution.
     adr018_flags: wifi_densepose_hardware::Adr018Flags,
     amplitudes: Vec<f64>,
+    /// 802.11 sequence control of the overheard frame — wire v3
+    /// (`CSI_MAGIC_V3`) only; `None` from v1/v2 firmware.
+    ///
+    /// This is the join key for cross-node fusion. `sequence` above is a
+    /// counter private to each node, so two nodes that captured the same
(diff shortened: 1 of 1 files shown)


clonea1:
Mapped this cohort against ADR-345 in #1956. **This PR is the root of it** —
it is the wire change that lets a measurement say which link it belongs to,
which is ADR-345's whole premise, and ADR-345 measured that on node 2 the AP
was only 34% of frames while the other 66% were peer frames the sink could
not attribute.

One commit, server-side only, no dependencies. Its parser commit is carried
by #1828, #1833 and #1834 because a fork PR cannot be based on another fork
branch; merging this one unblocks all three, and each then drops the
duplicate by patch-id on rebase with nothing to resolve.

---

oga35767-eng (approved):

PR #1808, opened 2026-09-04
feat(server): pair CSI frames across nodes by transmission identity

clonea1:
Two receivers that captured the SAME transmission hold two views of one moment.
Comparing those is the only way to tell a change in the channel from a change
in one receiver, and it is what lets separated receiver groups be reasoned
about independently.

Timestamps cannot do this. Nodes are synchronised to a few hundred
microseconds at best, an unconverged or re-electing node reports a time that is
briefly wrong rather than absent, and network jitter sits on top of both. Any
time-window pairing therefore either misses real pairs or invents false ones.

So pairing is keyed on (transmitter MAC, 802.11 sequence number) instead --
transmission identity, which is a property of the frame rather than of the
clock that observed it. Two nodes reporting the same key saw the same
transmission, whatever their clocks say.

Wire v3 only, deliberately. A v1 or v2 frame carries no sequence number, and
keying it on a placeholder would pair every older node's frames with every
other node's -- exactly the state a fleet passes through during a partial OTA.
It is a `Some(rx_seq)` check, never an `unwrap_or(0)`.

Pairing runs BEFORE the per-node subcarrier-grid gate. That gate locks each
node to the densest grid it has seen and skips anything sparser, which is most
of a building's traffic; running ahead of it keeps those frames available to a
path that can use them.

Amplitude vectors are retained alongside each pairing so a finalised
transmission can be compared across the receivers that captured it rather than
only counted. Retention is best-effort: when the budget is full the pairing
statistics are unaffected and only the vectors are skipped.

Exposes GET /api/v1/fusion with per-pair counts and open-key depth.

---

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/fusion.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/fusion.rs
@@ -1,0 +1,549 @@
+//! Cross-node frame pairing, keyed on `(transmitter, 802.11 rx_seq)`.
+//!
+//! # Why this exists
+//!
+//! Fusing measurements from several receivers needs a way to say "these two
+//! observations are of the *same event*". Everything else the server has fails
+//! at that:
+//!
+//! * The `sequence` field is a counter private to each node. Two nodes that
+//!   captured the same packet report unrelated numbers.
+//! * Arrival timestamps at the sink measure network jitter, not capture time.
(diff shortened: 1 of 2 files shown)


clonea1:
Merge order: #1806 must merge before this PR. This branch already contains #1806's commit (`feat(server): parse CSI wire v2 and v3 headers`) verbatim; merging this first would duplicate it once #1806 lands.

PR #1815, opened 2026-09-04
feat(server): per-node signed Doppler diagnostics

clonea1:
Reports two figures per node on /api/v1/nodes: the moving energy in its Body
Velocity Profile, and the signed, energy-weighted mean radial velocity.

Retains sanitized per-node phase alongside amplitude, in lockstep so the two
can be zipped by index. Without phase there is no complex CSI, and without
complex CSI a BVP cannot carry direction at all.

The zero-velocity deadband is expressed in BINS, not m/s, and this is the part
worth reviewing. The natural way to write it is a velocity threshold -- exclude
anything slower than 0.05 m/s. That is wrong whenever the threshold is narrower
than one bin: with the default BvpConfig the spacing is
max_velocity * 2 / n_velocity_bins = 2.0 * 2 / 64 = 0.0625 m/s, so a 0.05 m/s
threshold excludes only the single exact-zero bin, and the immediate
neighbours, into which a Hann-windowed STFT always leaks real DC and static
energy, count as motion by construction. The result is a saturated metric that
reports near-maximum movement for a still room -- and a saturated metric looks
like a working one on a dashboard.

Excluding a bin COUNT, derived from each call's actual velocity_resolution,
stays correct if BvpConfig ever changes. The zero bin is found by minimum
absolute velocity rather than an equality test, so it is robust to even/odd bin
counts and float rounding.

MEASURED with the accompanying fixtures, static (0.05 Hz) against moving (2 Hz),
which differ only in rate so the deadband is the only thing separating them:

    deadband = 0 bins:  19.13 / 310.89 = 6.2%
    deadband = 2 bins:   0.56 / 122.59 = 0.45%

These are DIAGNOSTICS, not a position estimate. A magnitude alone carries no
directional information, and no position tier is built on them here.

---

### DIFF
--- a/v2/Cargo.lock
+++ b/v2/Cargo.lock
@@ -13445,6 +13445,8 @@ dependencies = [
  "mdns-sd",
  "midstreamer-attractor",
  "midstreamer-temporal-compare",
+ "ndarray 0.17.2",
+ "num-complex",
  "opentelemetry-appender-tracing",
  "opentelemetry-otlp",
  "opentelemetry_sdk",
--- a/v2/crates/wifi-densepose-sensing-server/Cargo.toml
+++ b/v2/crates/wifi-densepose-sensing-server/Cargo.toml
@@ -22,6 +22,16 @@ path = "src/main.rs"
 [dependencies]
 # Web framework
 axum = { workspace = true }
+
+# Pinned through the workspace so the sensing server and the signal crate agree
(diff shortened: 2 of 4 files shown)


clonea1:
Merge order: #1816 must merge before this PR. This branch already contains #1816's commit (`feat(signal): extract a signed BVP so Doppler direction survives`) verbatim; merging this first would duplicate it once #1816 lands.

PR #1822, opened 2026-09-04
test(sensing-server): cover room-builder config validation + persistence

clonea1:
Extracted POST /api/v1/config/room's validation into a pure
validate_room_config(&RoomConfig) -> Result<(), String> function so it's
unit-testable without a full SharedState/axum harness; behavior
unchanged, the handler just delegates to it now.

9 new tests: load-missing/malformed-file defaults, save+load round trip,
and validation for non-positive/non-finite dimensions, non-finite node
coordinates, duplicate node ids, and an empty node list.

Also adds a "Reload from Saved" button to the Room Builder UI (discards
unsaved local edits, re-fetches the live server config).

---

### DIFF
--- a/ui/app.js
+++ b/ui/app.js
@@ -5,6 +5,7 @@ import { DashboardTab } from './components/DashboardTab.js';
 import { HardwareTab } from './components/HardwareTab.js';
 import { LiveDemoTab } from './components/LiveDemoTab.js';
 import { SensingTab } from './components/SensingTab.js';
+import { RoomBuilderTab } from './components/RoomBuilderTab.js';
 import { apiService } from './services/api.service.js';
 import { wsService } from './services/websocket.service.js';
 import { healthService } from './services/health.service.js';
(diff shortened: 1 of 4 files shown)


clonea1:
Closing in favour of #1823, which contains every line of this PR.

The three Room Builder PRs (#1822, #1823, #1824) are one body of work opened
as three nested variants. Consolidating on the largest removes two PRs from
the review queue without losing anything. This was verified at tree level
before closing, not assumed:

**Patch-id.** Both commits of this PR — `feat(ui): add Room Builder tab`
(`b04ca8aa`) and `test(sensing-server): config validation + persistence`
(`59e460be`) — appear verbatim on #1823.

**The tip commit is the exception, and it is why patch-id alone is not
enough here.** This PR's `fix(sensing-server): follow node_positions_config
becoming a positional Vec` has patch-id `c20a5e0c`; #1823's has `596a9375`.
They are the same fix applied to different surrounding content (the file is
899 lines longer there), so the diffs differ and the patch-ids differ. A
patch-id sweep reports one commit unique to this PR and, read naively, would
say it holds work #1823 does not.

**So the test used was the tree.** `git diff --name-only` between the two
heads returns exactly three paths — `ui/components/RoomBuilderTab.js`,
`ui/mesh3d.html`, `v2/crates/.../main.rs`. Every other file is bit-identical.
Direction is strictly forward: +2234 / −43, and each of the 43 removed lines
has a successor — `RoomNode` grew a `floor` field so the test literals had to
change, the `label` lookup was refactored to serve `floor` too, and
`_toPixel`/`_toRoom` were rewritten against a footprint-aware `_view()`
instead of assuming a rectangle, which is the point of that branch.
Mechanically: every identifier in this PR's copy of `RoomBuilderTab.js` also
occurs in #1823's.

**One thing #1823 was genuinely missing, now fixed there.** The same check on
`main.rs` was not empty — it returned six tokens, all of them words from two
rationale comments this PR wrote and #1823 dropped while leaving the code
they explain unchanged: why `config_set_room` indexes `positions` by node id
rather than pushing, and the failure mode clause of the #1791 note. Both are
restored on #1823 as a comment-only commit, so this close costs nothing at
all.

Thanks — the validation and persistence tests here are the ones #1823 carries.

PR #1823, opened 2026-09-04
feat(room): draw the building's footprint instead of assuming a rectangle

clonea1:
The Room Builder modelled a storey as width x depth. Most buildings are not
rectangles, and the ones that are not tend to be exactly where the interesting
propagation happens: a wing off the main block, an L around a stairwell, a
garage bay that swallows a link.

Adds a footprint editor. A storey is a polygon of points placed on the canvas
rather than two numbers, so a node in a wing has coordinates that mean
something relative to the building it is actually in, instead of being placed
in empty space outside an assumed rectangle.

Width and depth are kept as the polygon's bounding box, so existing configs and
anything reading those two fields keep working unchanged.

---

---

**Update:** the UI (`RoomBuilderTab.js`) has read and written `config.footprint`
since this branch's first commit, but `RoomConfig` never had a matching
server-side field, and `config_get_room`'s hand-maintained response never
served it back — every drawn outline saved fine and then vanished on the
next page load. This PR now adds the `Footprint` struct, the field, the
GET-side fix, and validation (minimum 3 points per ring, non-finite-vertex
and oversized-ring rejection, floor-membership check). `Emitter`/`Room` are
deliberately untouched here — those belong to other in-flight PRs
(#1833/#1834) and a later storey-canvas PR respectively, and don't
interact with this field.

Also adds `v2/data/examples/generic-house/`: a synthetic ~1,500 sq ft
single-story house with a basement (invented rooms, nodes, and
locally-administered example MACs — nothing from a real residence), so
this PR has something a reviewer can actually run and see the footprint
outline on, rather than landing with no demo until a later PR:

```bash
RUVIEW_DATA_DIR=v2/data/examples/generic-house cargo run -p wifi-densepose-sensing-server
```

See that fixture's own `README.md` for exactly what renders on this
branch today vs. what's carried in the same JSON for follow-up PRs.

### DIFF
--- a/ui/app.js
+++ b/ui/app.js
@@ -5,6 +5,7 @@ import { DashboardTab } from './components/DashboardTab.js';
 import { HardwareTab } from './components/HardwareTab.js';
 import { LiveDemoTab } from './components/LiveDemoTab.js';
 import { SensingTab } from './components/SensingTab.js';
+import { RoomBuilderTab } from './components/RoomBuilderTab.js';
 import { apiService } from './services/api.service.js';
 import { wsService } from './services/websocket.service.js';
 import { healthService } from './services/health.service.js';
@@ -159,6 +160,15 @@ class WiFiDensePoseApp {
     const sensingContainer = document.getElementById('sensing');
     if (sensingContainer) {
       this.components.sensing = new SensingTab(sensingContainer);
+    }
+
+    // Room Builder tab
+    const roomBuilderContainer = document.getElementById('roombuilder');
+    if (roomBuilderContainer) {
+      this.components.roomBuilder = new RoomBuilderTab(roomBuilderContainer);
+      this.components.roomBuilder.init().catch(error => {
+        console.error('Failed to initialize room builder:', error);
+      });
     }
 
     // Training tab - lazy load to avoid breaking other tabs if import fails
@@ -344,6 +354,14 @@ class WiFiDensePoseApp {
         }
         if (this.components.modelPanel && typeof this.components.modelPanel.refresh === 'function') {
           this.components.modelPanel.refresh();
+        }
+        break;
+
+      case 'roombuilder':
+        // Re-fetch the live config each time the tab is opened, in case it
+        // changed elsewhere (another browser tab's save, a server restart).
+        if (this.components.roomBuilder) {
+          this.components.roomBuilder._load();
         }
         break;
     }
--- a/ui/components/RoomBuilderTab.js
+++ b/ui/components/RoomBuilderTab.js
@@ -45,10 +45,18 @@ export class RoomBuilderTab {
     // node with the wall below it -- which is the whole point of stacking
     // storeys on a shared origin.
     this._activeFloor = 1;
-    // 'select' drags nodes and the AP; 'wall' draws segments.
+    // 'select' drags nodes and the AP; 'wall' draws segments; 'footprint'
+    // clicks out the building's outline vertex by vertex.
     this._mode = 'select';
     this._wallStart = null;
     this._wallPreview = null;
+    // The outline being traced right now, as {x, y} in room coordinates. It
+    // is not part of `config` until closed: a half-traced ring is not a shape,
+    // and the server rejects one with fewer than three vertices.
+    this._ringDraft = null;
+    // Where the pointer is, so the segment about to be committed is visible
+    // before the click that commits it.
+    this._ringHover = null;
     this._loaded = false;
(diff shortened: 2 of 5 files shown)


clonea1:
Consolidating the Room Builder work here. #1822 and #1824 are now closed in
favour of this PR, which contains every line of both.

The three were one body of work opened as three nested variants —
`ui-room-builder` ⊂ `ui-room-builder-geometry` ⊂ `ui-room-builder-footprint`.
Rather than take the nesting on trust, it was checked at tree level:
`git diff --name-only` between this head and #1822's returns three paths and
between this head and #1824's returns one, every other file in all three
trees being bit-identical; direction is strictly forward (+2234 / −43 from
#1822, +368 / −20 from #1824); and every identifier in either PR's copy of
`ui/components/RoomBuilderTab.js` also occurs in this one's.

Worth recording because it nearly produced the wrong answer: the tip commit
does **not** share a patch-id across the three. #1822 carries `c20a5e0c` and
this PR carries `596a9375` — the same `node_positions_config` fix applied to
a file 899 lines longer, so the diffs and therefore the patch-ids differ. A
patch-id sweep reports a commit unique to #1822. Patch-id is a commit
identity test; supersetness is a property of the tree.

The one place this branch was genuinely not a superset was comments, not
code: `ed878ca9` restores two rationale blocks #1822 wrote that were dropped
here while the code they explain stayed the same — why `config_set_room`
indexes `positions` by node id rather than pushing (a sparse or
non-contiguous id set still lands each node at the right slot), and the
failure-mode clause of the #1791 note. Comment-only;
`cargo check -p wifi-densepose-sensing-server --no-default-features` passes.

This PR is now the Room Builder base. Reviewing it unblocks the per-storey
canvas, magnetic room and server-side persistence work that stacks on it.

clonea1:
Correction to my note above, from the CI run that comment triggered: **this
branch merges cleanly but does not compile against current `main`.** The
comment-only commit is not the cause — what it did was re-run a check that
had been stale for ten days.

```
main.rs:12813  error[E0277]  `(&u8, &[f32; 3])` cannot be indexed by `{integer}`   (×3: x, y, z)
main.rs:12984  error[E0308]  expected `HashMap<u8, [f32; 3]>`, found `Vec<[f32; 3]>`
```

The tip commit here, *"follow node_positions_config becoming a positional
Vec"*, followed #1791 (`9b595b83`, 2026-09-04). That was reverted the same
afternoon by `d6347fe1` — "key node positions by node id, not by list
position" — and hardened on 2026-09-08 by #1860 (`0d1a247b`, "a silent node no
longer moves its neighbours' coordinates"). These PRs were opened at 19:07 on
2026-09-04, about seven hours after the revert, so the branch has been chasing
a convention that no longer exists. `node_positions_config` is
`HashMap<u8, [f32; 3]>` on `main` and has been since.

The last CI run here was 2026-09-04T20:28Z — minutes after the breaking
commit landed and before it was in the merge base. The checks have read green
ever since while the branch had stopped compiling, which is worth flagging
generally: `mergeStateStatus: CLEAN` and `git merge-tree` are textual, and a
type that changes underneath a branch conflicts semantically and merges
silently.

This does not change the consolidation — #1822 and #1824 carry the identical
tip commit (`node_positions_config: Vec<[f32; 3]>` in all three), so closing
them removed two broken duplicates rather than a working alternative, and this
PR is still their strict superset.

I have not pushed the fix, because the right shape is a decision about the
node-position model rather than a compile fix, and getting it wrong
reintroduces the rank-is-not-identity defect #1860 fixed. For whoever picks it
up: `config_get_room` needs to iterate the map rather than `.enumerate()` a
Vec, and `config_set_room` needs to keep both representations — a `HashMap`
insert for `node_positions_config` and a positional `Vec` for
`multistatic_fuser.set_node_positions`, which are genuinely different things.

Happy to push that rework here if you would like it in this PR rather than a
follow-up.

clonea1:
Cohort map in #1956. ADR-345 computes its bearings from `room_config.json`
and its whole argument rests on *"both endpoints known, fixed, and under our
control"* — this PR is the **known** half. It turns `room_config.json` from a
CLI flag over an assumed rectangle into a real building: storeys, walls and
the actual footprint.

Independent of the wire/link PRs — it overlaps nothing in #1806 or #1828 —
but its four Room Builder commits are also carried by #1833 and #1834, which
need them to compile. Verified: `main` + #1806 + #1805 + #1828 + this merges
clean and compiles (`cargo check -p wifi-densepose-sensing-server
--no-default-features`, exit 0).

---

PR #1824, opened 2026-09-04
feat(room): storeys, walls and AP placement in the Room Builder

clonea1:
Extends the Room Builder from a single flat rectangle to a building: multiple
storeys, walls drawn per storey, and a placed access point.

Storeys share one origin. The origin (0, 0, 0) is the north-west corner of the
first floor at floor level, and every storey is measured from it rather than
being re-zeroed, so a node's X/Y means the same thing on every floor and
distances between nodes on different storeys are plain Euclidean. Nodes and
walls on other storeys are drawn faintly rather than hidden, which is what lets
an upper-floor node be lined up against the wall below it.

Storey elevation is derived from ceiling height plus subfloor thickness rather
than typed in directly, because those are the two numbers someone can actually
measure. Heights are entered in inches.

A node carries `floor` alongside `z`. That is redundant for geometry and
deliberately so: it is what the builder groups by and what walls are associated
with, and deriving it from `z` would guess wrong for a node mounted high in a
stairwell. `floor` is optional, so a config written before storeys existed keeps
working unchanged.

Every persisted field is served back by GET /api/v1/config/room. That response
body is hand-maintained and will not complain when it falls behind the struct,
so it carries a comment saying so at the point where the next field will be
added -- a field that is saved but not served looks exactly like "the setting
will not stick".

Wall entry takes a start and an end and refuses a zero-length wall, which is
otherwise easy to stage by double-tapping a corner.

---

### DIFF
--- a/ui/app.js
+++ b/ui/app.js
@@ -5,6 +5,7 @@ import { DashboardTab } from './components/DashboardTab.js';
 import { HardwareTab } from './components/HardwareTab.js';
 import { LiveDemoTab } from './components/LiveDemoTab.js';
 import { SensingTab } from './components/SensingTab.js';
+import { RoomBuilderTab } from './components/RoomBuilderTab.js';
 import { apiService } from './services/api.service.js';
 import { wsService } from './services/websocket.service.js';
 import { healthService } from './services/health.service.js';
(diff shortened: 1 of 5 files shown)


clonea1:
Closing in favour of #1823, which contains every line of this PR.

The three Room Builder PRs (#1822, #1823, #1824) are one body of work opened
as three nested variants. Consolidating on the largest removes two PRs from
the review queue without losing anything. Verified at tree level before
closing:

**Patch-id.** All three of this PR's substantive commits appear verbatim on
#1823 — `feat(ui): add Room Builder tab` (`b04ca8aa`),
`test(sensing-server): config validation + persistence` (`59e460be`), and
`feat(room): storeys, walls and AP placement` (`dd02ff74`), the commit unique
to this PR against #1822. The tip commit matches too (`596a9375`).

**Tree.** `git diff --name-only` between this head and #1823's returns a
single path: `ui/components/RoomBuilderTab.js`, +368 / −20. Every other file
in the two trees — `ui/mesh3d.html` and `v2/crates/.../main.rs` included — is
bit-identical. The 20 removed lines are the rectangle-assuming `_toPixel` /
`_toRoom` bodies, replaced by the footprint-aware `_view()` that is the
subject of #1823. Every identifier in this PR's copy of that file also occurs
in #1823's.

#1823 is this PR plus one commit.

Thanks — the storeys/walls/AP work here is what #1823 builds the footprint on.

PR #1827, opened 2026-09-04
feat(firmware): change node configuration over the network

clonea1:
Changing any node setting meant a USB cable and physical access. On a fleet
mounted on walls and ceilings that is the difference between a five-second
change and taking a board down.

Adds an authenticated HTTP config endpoint on each node. Settings are typed and
range-checked at the boundary, persisted to NVS, and applied without a reflash.
Includes LED mode and brightness: the status LED is bright enough to be
unwelcome in a bedroom, and that is a configuration question, not a reflash.

TRIAL AND REVERT. A setting that breaks connectivity would otherwise brick a
node remotely -- change the WiFi password and the node is simply gone. So a
change that could sever the uplink is applied on trial: the old values are
banked, the node reboots, and the change is confirmed only once the node
reassociates and gets an address. If it does not within the deadline, the node
restores the banked values and comes back on the old settings. Verified on
hardware with a deliberately wrong password: recovered on its own in 60.5 s.

The reply to a trial push is sent before the reboot rather than after it, so
the caller learns the trial was accepted instead of seeing a dropped
connection and having to guess.

Mutating requests require a pre-shared key checked in constant time, and the
endpoint FAILS CLOSED when no key is provisioned. The key is read from a file
path given at build time with no default, so no personal path is baked into a
published tree.

---

Rebased onto current `main`. One conflict, in `firmware/esp32-csi-node/main/CMakeLists.txt`: `main` has since gained `thermal.c` in the source list while this branch adds `config_api.c`. Resolved by keeping both. Firmware builds clean for esp32c6 on ESP-IDF v5.4.


clonea1:
Both CI failures addressed.

**`json` on ESP-IDF 6.0.2 — ours, fixed in this push.** cJSON ships as the bundled `json` component through 5.x and moved out to the component manager in v6, where requiring it fails outright. It is now required only when `IDF_VERSION_MAJOR < 6`, with `espressif/cjson` declared under an `idf_version >=6.0` rule so the two never coexist in one build. Verified building on esp32c6/16MB with ESP-IDF 5.4.

**The size-gate failure is pre-existing and not specific to this branch.** This PR was DIRTY, so CI had never run on it; making it mergeable is what surfaced it. `main` itself builds the 8MB variant at 1,171,248 bytes — 99.3% of the 1152 KB gate, but only 56% of the 2 MB partition it actually flashes into. This branch adds ~26 KB and crosses a ceiling that is roughly half the real space.

Rather than trim working code to fit a budget that does not match the hardware, that is fixed properly in #1891, which sizes each variant's limit to its own `ota_0` slot. This PR should go green once that lands.

https://claude.ai/code/session_01PVWMiHQifoYXL7uL3bphrZ

PR #1828, opened 2026-09-04
feat(server): per-link CSI metrics keyed by (receiver, transmitter)

clonea1:
Adds a LinkTable that accumulates CSI per (rx_node, tx_mac) pair, and runs it
BEFORE the existing per-node subcarrier-grid gate.

The gate locks each node to the densest grid it has seen. A node associates
with one AP and receives HE-SU data from it at 256 bins, while everything else
it merely overhears sends beacons and management frames at 64. The lock
therefore always settles on the associated AP's format, and every
non-associated transmitter is sparser BY CONSTRUCTION -- dropped by a bare
continue before it can become a link.

Grid consistency is a property of the TRANSMITTER, so it belongs on the link
rather than the node. LinkTable applies the same densest-wins policy per
(rx_node, tx_mac).

The existing gate is NOT modified. The per-node feature path behaves exactly
as before; the link path simply runs ahead of the continue, so a frame the
node path cannot use stays available to consumers that can.

MEASURED 2026-09-01, before -> after:
  links         31 -> 137      transmitters  10 -> 32
  renderable    29 ->  88      grids in use  256 only -> 64:97, 256:33, 128:7
  illuminators with two or more receivers: 1 -> 24

97 of the 137 recovered links are 64-bin: precisely the population being
discarded. No firmware, hardware or network change -- the frames were always
arriving.

Adds GET /api/v1/links and GET /api/v1/links/inventory. The inventory endpoint
reports the table's full contents including links the metrics view omits, with
the reason for each, because that filter_map dropped links silently.

---

Rebased onto current `main`. Two conflicts in `main.rs`, both purely additive and resolved by keeping both sides: `main` has since gained the room-debounce fields (#1793) where this branch adds `link_table`. Compiles clean and the server test suite passes.

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/links.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/links.rs
@@ -1,0 +1,1357 @@
+//! Per-link CSI state, keyed by (receiver, transmitter).
+//!
+//! The rest of the server models one link per node: `NodeState` holds a single
+//! `frame_history` per `node_id`. That was accurate while the only labelled
+//! traffic was AP -> node. It is not accurate for what the radios actually
+//! hear: a node in promiscuous MGMT+DATA mode receives CSI from the AP, from
+//! every peer node's ESP-NOW beacon, and from unrelated household traffic —
+//! measured 2026-08-28 at roughly 75% non-AP on a normal home channel.
+//!
+//! Interleaving those into one history mixes links with completely different
+//! geometry. Until wire v2 carried the transmitter MAC there was no way to
+//! separate them, so the only remedy was `filter_mac`, which fixes the mixing
+//! by throwing ~75% of the measurements away.
+//!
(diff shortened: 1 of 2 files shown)


clonea1:
Cohort map in #1956. This is ADR-345's per-link model itself: CSI keyed on
(receiver, transmitter) instead of per node. `MEASURED 2026-09-01`, before →
after: links 31 → 137, transmitters 10 → 32, illuminators heard by two or
more receivers 1 → 24 — no firmware, hardware or network change, those
frames were always arriving and were being dropped by a gate that assumed one
transmitter per node.

It carries #1806's parser commit (`b85eb629df8e`) because it needs it to
compile and a fork PR cannot be based on another fork branch. If #1806 lands
first, that commit drops out of this one automatically on rebase.
`main` + #1806 + this merges clean.

---

PR #1832, opened 2026-09-04
feat(provision): keep credentials out of the state file entirely

clonea1:
provision.py cached every provisionable attribute in a per-port JSON state file
so a re-run could merge prior values. That list included 'password' and
'seed_token', so every successful run wrote the WiFi passphrase in cleartext to
a file under the user's config dir -- defeating the point of keeping the
credential in a file outside the repo, and leaving stale copies of retired
passwords lying around after an SSID change.

Secrets are now excluded from the merge list AND filtered again on write, so a
credential cannot reach the disk by a route someone adds later. The cost is
that a secret must be supplied on every run rather than merged from prior
state; provision_node.py already does exactly that, reading them from files,
and a direct provision.py call now fails loudly instead of silently reusing an
old credential.

Related upstream work: PR #1760 makes the same state file owner-only, which is
worth having, but still lists password and seed_token as mergeable -- so it
protects a file that should not contain the secret at all. These compose: their
permissions plus this exclusion.

board_index.py additionally prefers esptool's 'BASE MAC' line over any MAC-like
string in the output; the other lines are derived forms and picking one of
those keys a board by an address it does not provision under.

---

Rebased onto current `main`; one additive `.gitignore` conflict resolved by keeping both sides. Scanned the diff for credential literals before pushing (none), and the three scripts byte-compile.

### DIFF
--- a/.github/workflows/firmware-qemu.yml
+++ b/.github/workflows/firmware-qemu.yml
@@ -262,6 +262,32 @@ jobs:
             firmware/esp32-csi-node/test/timeout-*
           retention-days: 30
 
+  # ---------------------------------------------------------------------------
+  # Provisioning tooling tests (firmware/esp32-csi-node/tests).
+  #
+  # These existed but nothing ran them: the repo's pytest invocations cover
+  # archive/v1, python/ and tests/performance, and the `Tests (3.x)` job marks
+  # every step continue-on-error, so it cannot fail a PR. A change that put the
+  # WiFi passphrase back into the per-port state file -- or took it out and
+  # broke the merge contract -- would have gone in green. Stdlib unittest, no
+  # dependencies, under a second.
+  # ---------------------------------------------------------------------------
+  provision-tests:
+    name: Provisioning Tooling Tests
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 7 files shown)


ruvnet (commented):
Dream portfolio exact-head review (2026-09-05)

Frozen hypothesis: after provisioning or re-provisioning, `password` and `seed_token` must never enter the per-port JSON state, missing fresh secrets must fail loudly, and the supported provisioning paths must retain deterministic board identity and OTA behavior under targeted regression tests.

Reachable code now removes the two secret attributes from the merge list and filters them again on write. All six exact-head workflows pass, including CI, firmware, QEMU, security, and policy gates. However, this five-file change adds no targeted regression test for legacy state containing either secret, save/load round-trips, missing-secret failure, or file cleanup. The new wrapper also deliberately passes the WiFi password and OTA key through child-process argv; that residual is explicitly disclosed but is CWE-214 and remains observable to local process inspection.

The state-file direction is sound, and the argv limitation should not block a narrowly scoped state-migration fix, but the frozen state contract has not been executed. Add failure-injection tests that seed legacy secret-bearing state, verify scrubbed output, and exercise both direct and wrapper paths; retain the current hardware/field limitations.

INCONCLUSIVE

PR #1833, opened 2026-09-04
feat(emitters): four-state triage for the transmitters a fleet can hear

clonea1:
#1791 landed node_positions_config as a positional Vec<[f32;3]> keyed by
active-node rank, not the HashMap<u8,[f32;3]> the Room Builder handlers carried
in this stack, so they stopped compiling once rebased onto current main.

Read path enumerates and derives the id from the index, matching the convention
the fusion path uses. Write path builds the vector indexed BY node id before
assigning, so a sparse id set still lands each node in the right slot rather
than shifting everything after a gap.

---

**Stacked**, eight commits — it carries the Room Builder work and the per-link metrics as well as its own. Merge the earlier PRs in the stack first (#1822/#1824 room builder, #1828 links) and this shrinks accordingly.

Rebased onto current `main`. Conflicts were additive and resolved by keeping both sides; one real fix was needed because #1791 changed `node_positions_config` from a `HashMap<u8,[f32;3]>` to a positional `Vec`, which the Room Builder handlers in this stack were written against. 329 server tests pass.

### DIFF
--- a/ui/app.js
+++ b/ui/app.js
@@ -5,6 +5,7 @@ import { DashboardTab } from './components/DashboardTab.js';
 import { HardwareTab } from './components/HardwareTab.js';
 import { LiveDemoTab } from './components/LiveDemoTab.js';
 import { SensingTab } from './components/SensingTab.js';
+import { RoomBuilderTab } from './components/RoomBuilderTab.js';
 import { apiService } from './services/api.service.js';
 import { wsService } from './services/websocket.service.js';
 import { healthService } from './services/health.service.js';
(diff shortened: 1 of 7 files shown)


clonea1:
Cohort map in #1956. This is the **fixed** half of ADR-345's *"both endpoints
known, fixed, and under our control"*: once the fleet can hear every
transmitter on the channel, something has to decide which of them can be
trusted to stand still. A neighbour's router is fixed; a phone in a pocket is
not, and an emitter that moves teaches an association that is only true while
it stands still — which is worse than having no illuminator at all.

It carries #1806, #1828 and #1823's commits because it needs all three to
compile. Clean against `main` today and after #1806, but it will conflict
once #1828 or #1823 land. **That one is mine** — when they land I will rebase
and force-push, at which point it reduces to the emitter-triage commit alone.
No action needed here.

---

PR #1834, opened 2026-09-04
feat(ui): integrated 3D view of the building, fleet and illuminators

clonea1:
#1791 landed node_positions_config as a positional Vec<[f32;3]> keyed by
active-node rank, not the HashMap<u8,[f32;3]> the Room Builder handlers in this
stack were written against.

Read path enumerates and derives the id from the index, matching the convention
the fusion path uses. Write path builds the vector indexed BY node id before
assigning, so a sparse id set still lands each node in the right slot.

---

**Stacked**, nine commits — it carries the Room Builder work as well as its own. Merge the earlier PRs in the stack first and this shrinks.

Rebased onto current `main`. Conflicts additive, resolved by keeping both sides; one real fix needed because #1791 changed `node_positions_config` from a `HashMap<u8,[f32;3]>` to a positional `Vec`. Server test suite passes.

### DIFF
--- a/ui/app.js
+++ b/ui/app.js
@@ -5,6 +5,7 @@ import { DashboardTab } from './components/DashboardTab.js';
 import { HardwareTab } from './components/HardwareTab.js';
 import { LiveDemoTab } from './components/LiveDemoTab.js';
 import { SensingTab } from './components/SensingTab.js';
+import { RoomBuilderTab } from './components/RoomBuilderTab.js';
 import { apiService } from './services/api.service.js';
 import { wsService } from './services/websocket.service.js';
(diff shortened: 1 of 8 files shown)


clonea1:
Cohort map in #1956. This is the one that makes the rest checkable: a
geometry error in `room_config.json` is invisible in JSON and obvious in a
picture of the building with the fleet and the illuminators in it.

Mechanically it is #1833 plus exactly one new file, `ui/world3d.html` —
`git diff` between the two PR heads is that file and nothing else, and
`main.rs` is the same blob on both. Review #1833 first; this reduces to the
single file once #1833 lands, and rebasing it is mine to do.

---

PR #1838, opened 2026-09-04
feat(firmware): count the edge pipeline's frames, so a dead one is visible

clonea1:
The subcarrier-grid guard in process_frame() rejects any frame wider than
EDGE_MAX_SUBCARRIERS and records nothing. That silence hid a real bug for
weeks: the grid was sized 128 for pre-HE parts while an HE-capable C6 delivers
256-bin HE20 frames, so every frame took that return and vitals, presence and
fall detection all produced nothing -- on nodes reporting full frame rate, good
RSSI, healthy heap and a normal DSP banner. Nothing observable distinguished it
from working.

Two counters, monotonic since boot, in the existing NodeHealth carrier:

  frames_processed  passed the guard and were processed
  frames_rejected   discarded by the guard (empty, or wider than the grid)

Separately each is ambiguous; together they separate the two silences:

  processed == 0 && rejected == 0   nothing reaches the DSP task at all
  processed == 0 && rejected  > 0   frames arrive and every one is discarded,
                                    i.e. the grid does not match the radio

The second row is the bug, and it now names itself from the server with no
cable -- which matters because these boards are on walls with no console, and
the counter that would have caught this was reachable only over serial.

Wire: sync packet 50 -> 58 bytes, proto v3 -> v4, edge counters at [50..57].
The parser was already versioned for this, so both directions stay compatible:
an old server requires only len >= its own size and ignores the tail, and a new
server admits the field only when proto_ver >= 4 AND len >= 58. Tests cover the
padded-v3 and truncated-v4 cases specifically, because either guard alone would
report a confidently wrong zero for a node whose pipeline was never measured --
worse than absent, exactly as with the MAC and TX counters before it.

/api/v1/mesh needs no change: NodeSyncSnapshot already derives Serialize over
NodeHealth, so `edge` appears beside `tx` automatically.

Verified: firmware builds for esp32c6 in espressif/idf:v5.4; 28 sync_packet
tests pass including five new ones; sensing-server checks clean. Not yet
observed on hardware -- the counters are the instrument for that, and the fleet
is mid-experiment.

---

## Stacked on #1818

Four commits; only the last is new. The sync packet has to reach 50 bytes
before these counters can be appended at 50..57, so this depends on the wire
and TX-counter work in #1818 (which itself carries #1805 and #1807). Merge
those first and this reduces to one commit.

Firmware builds clean for esp32c6 on ESP-IDF v5.4; the 28 sync-packet parser
tests pass, including the padded-older-version and truncated-newer-version
guards -- either alone would report a confidently wrong zero for a node whose
pipeline was never measured, which is worse than reporting nothing.


clonea1:
Merge order: #1805 must merge before this PR. Both touch csi_collector.c/.h, esp_stubs.h and test_csi_sanitize.c with overlapping CSI wire v3 (transmitter identity) changes; merging this first will conflict when #1805 is rebased afterward.

PR #1840, opened 2026-09-05
Improve empty room bootstrap confidence

ruvnet:
## Summary

Learn a hardware adaptive empty room residual boundary from bounded runtime sized windows.

Bind the single link model to exactly one ESP32 source and prevent other radios from contaminating calibration or runtime comparison.

Persist only aggregate scalar references with exact byte integrity verification. Restore them as a negative only startup prior.

Apply the same numeric vital gate to both sensing_update and the legacy edge_vitals WebSocket message.

Expose background conformance, source binding, and authority state through calibration status.

## Validation

1. cargo test --workspace --no-default-features
2. cargo test -p wifi-densepose-signal --features eigenvalue
3. cargo test -p wifi-densepose-sensing-server --no-default-features
4. git diff --check
5. Secret pattern scan across every changed file

All commands passed locally. Existing unrelated compiler warnings remain unchanged.

## Measured physical evidence

One empty home capture bound to Node 5 collected 1762 accepted frames over 655 seconds and produced 35 privacy reduced reference windows. Promotion observed 12 of 12 fresh empty samples with zero stale samples and zero numeric vital samples. After restart, 12 of 12 fresh samples matched the stored background and exposed no numeric rates. Empirical background conformance ranged from 50.0 to 64.3 percent with a median of 54.3 percent.

This proves negative background behavior for one installation only. No occupied holdout was captured, so this pull request makes no claim about person recall, person counting accuracy, or cross home accuracy.

## Authority boundary

The bootstrap image can suppress a matching empty background. It cannot authorize positive presence, pose, identity, calibrated evidence, or numeric vital signs. Explicit room calibration remains required.

### DIFF
--- a/docs/adr/ADR-355-local-empty-room-bootstrap-prior.md
+++ b/docs/adr/ADR-355-local-empty-room-bootstrap-prior.md
@@ -23,17 +23,26 @@ directory, which is unreliable for embedded application hosts.
 RuView may persist a completed empty room field model as a local bootstrap
 prior only after two server controlled stages:
 
-1. Normal calibration reaches both 12,000 accepted frames and 600 seconds.
-2. Promotion observes 12 new one second samples. At least 10 must resolve
-   empty, all ticks must be fresh, and none may publish numeric vital signs.
+1. Normal calibration reaches both 1,000 accepted frames and 600 seconds.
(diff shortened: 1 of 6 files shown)


heamrann-cloud:
Coppy

PR #1860, opened 2026-09-07
fix(sensing-server): key node positions by node id, not by list position

clonea1:
`node_positions_config` is keyed by **list position**, but read back by **node id**. Those are only the same thing when node ids happen to be `0,1,2,…` in order.

## The failure

A position ends up depending on which *other* nodes happen to be alive. Give the Nth-lowest active node id the Nth entry of an ordered list, and one node going quiet re-ranks every node above it — so its neighbours silently inherit each other's coordinates. Nothing errors; the fleet just quietly believes its sensors are somewhere they are not.

On a nine-node fleet a single missed beacon is enough to trigger it.

## The change

Key the map by node id, so the mapping is **stated rather than inferred**:

- the struct field, its initialiser, and the local in `main()`
- `node_positions_by_active_id`, whose active-node filter is unchanged
- identity comes from an explicit `node_id:` prefix when given, and from list index otherwise, built by a tested function rather than a loop inside `main()` that no test can reach

The fuser's positional list is deliberately left alone — this change is about the `NodeInfo` output, and quietly altering fusion semantics inside it would be the wrong scope.

## Tests

The three existing tests for the helper are ported to the keyed form, and one is added that pins the difference:

```
a_stale_node_does_not_move_the_others
```

It asserts node 13 does not inherit node 12's coordinates when node 12 goes quiet — the exact silent failure above.

Full suite passes.

https://claude.ai/code/session_01BHNNWbmKvxXrVaEDwXEBNo


ruvnet (commented):
Dream exact-head review — **REJECT** for the frozen end-to-end identity hypothesis; the NodeInfo-only improvement is acknowledged.

All five observed workflows pass, and explicit IDs are correctly preserved in NodeInfo. The effectful fusion path still converts the configured map into a positional vector before `set_node_positions`, while frames are ordered by observed node ID. Sparse IDs, reversed configuration order, or node churn can therefore associate a frame with the wrong position downstream.

Issue #1866 records the minimal acceptance gate: keep the mapping keyed by ID through fusion (or derive positions from the exact same ordered active-ID cohort), add sparse/reordered/missing/join-leave regressions, preserve documented legacy behavior, and rerun the full exact-head suite. No hardware-accuracy claim is made here.

oga35767-eng:
Ty

clonea1:
Pushed `0d1a247b`, which addresses the #1866 acceptance gate.

**The defect.** `MultistaticFuser::fuse` addressed positions by the frame's index in the cohort it was handed. That index is a rank, not an identity: `node_frames_from_states_with_guard` compacts the cohort before `fuse` sees it, dropping stale, out-of-guard and empty-history nodes. One board missing a beacon therefore re-ranked every node above it and handed each its neighbour's coordinates. `MultiBandCsiFrame` already carries `node_id`, so the identity was present at the boundary and simply unused.

**The change.** Positions are stored keyed by node id and resolved per frame by that id. `main` now passes the same `node_positions_by_id` map that already feeds `NodeInfo`, so one parsed identity serves both the metadata and the effectful path. `FusedSensingFrame::node_positions` stays parallel to `node_frames`, each entry resolved from its own frame's id, so the pairing survives compaction.

**Legacy compatibility.** `set_node_positions(Vec)` is kept as an explicit, documented mapping — entry `i` is node `i` — which is exactly what a bare `--node-positions` list has always meant. `node_positions()` returns that dense view.

**One behaviour change, called out deliberately.** With a bare (unprefixed) list on a fleet whose ids are not dense from zero, a node now reads the entry declared for its id rather than the entry at its rank. That is the defect being fixed. Explicit `node_id:x,y,z` entries are unaffected, as is any dense-from-zero fleet with every node reporting.

**Tests.** Sparse ids, reversed configuration order, a configured-but-absent node, an unconfigured node, join/leave churn, the preserved legacy mapping, and a negative control pinning the compacted-cohort case. Negative-controlled against the pre-fix lookup: six of the eight fail without it. The legacy-mapping test is expected to pass both ways, and order-independence became structural once storage was keyed.

`cargo test --workspace --no-default-features` (excluding `wifi-densepose-desktop`, which needs GTK headers unavailable on this machine): **4558 passed, 0 failed**, 241 binaries. No new warnings.

Software-only scope; no hardware-accuracy claim.

Two related notes:

- The Room Builder branches (#1823, #1824) add a second call site in `config_set_room` that builds a Vec indexed by node id and hands it to the previously rank-addressed fuser. Those become correct automatically once this lands, since the preserved legacy semantics are exactly what that code already builds. No change needed there.
- `StreamingEngine` has its own `MultistaticFuser` that is never given positions at all, on the governed `fuse_scored_calibrated` path. Different instance, out of scope here — filed separately as #1870.

clonea1:
Pushed `0d1a247b`, which takes this from the NodeInfo-only change you accepted to the full boundary fix. Mapping it to the acceptance criteria in #1866:

**1. Positions keyed by node ID through the fusion boundary.** Taken as stated rather than the equivalent-cohort alternative. `MultistaticFuser` stores `HashMap<u8, [f32; 3]>`, and `fuse` resolves each entry via `node_frames[i].node_id`. `MultiBandCsiFrame` already carried its own id, so the identity was present at the boundary and simply unused. `main` now passes the same `node_positions_by_id` map that feeds `NodeInfo`, so one parsed identity serves both paths.

**2. Adversarial regressions.** Sparse ids (11/12/13), reversed configuration order, a configured-but-silent node, a configured-but-absent node, an unconfigured node, and a join/leave churn walk (`4 → 3 → 2 → 1 → 2 → 4` nodes).

**3. Every fused contribution uses its own node's position.** `assert_each_frame_owns_its_position` zips `node_frames` against `node_positions` and checks each pair by id; every case above routes through it.

**4. Legacy positional configuration preserved.** `set_node_positions(Vec)` is kept and documented as the compatibility mapping — entry `i` is node `i`, which is what a bare `--node-positions` list has always meant. `node_positions_by_id` still applies that rule to unprefixed entries. Covered by `legacy_positional_setter_still_means_index_equals_node_id`, and the pre-existing `fused_positions_filled` / `node_positions_set_and_retrieved` pass unchanged.

**5. Workflows.** Green on the exact head (`0d1a247b`): 24 success, 3 skipped, 0 failures.

**6. Scope.** Software-only; no hardware-accuracy claim.

---

**One judgment call worth your eye before merging.** There is a real behaviour change: with a bare, unprefixed `--node-positions` list on a fleet whose ids are not dense from zero, a node now reads the entry declared for *its id* rather than the entry at its cohort rank. That is the defect being fixed, and entries given explicitly as `node_id:x,y,z` are unaffected, as is any fleet with dense zero-based ids and every node reporting — but it is a semantics change on the bare-list path, so I did not want it to arrive silently. It is called out in the commit body too.

**On why the cohort shape matters.** `node_frames_from_states_with_guard` compacts before `fuse` sees anything — stale, out-of-guard and empty-history nodes are dropped — so a frame's index is its rank this cycle, not its identity. Under the previous lookup, one node missing a single beacon re-ranked every node above it onto its neighbour's coordinates, with no error and no log.

**On test strength.** I reverted the lookup and re-ran: six of the eight new tests fail against the pre-fix code. The other two pass both ways by design — the legacy-mapping test is meant to, and order-independence became structural once storage was keyed rather than indexed. `rank_addressing_would_have_failed_this_cohort` is a negative control pinning the exact compacted case, since rank and id only diverge once a node drops.

**Scope note.** `FusedSensingFrame::node_positions` has no consumer in-tree today, so this changes no current output — it is a latent-correctness fix, not a measured improvement. Separately, `StreamingEngine` (`wifi-densepose-engine/src/lib.rs:116`) holds a second `MultistaticFuser` that is never given positions at all, and `set_multistatic_config` rebuilds it wholesale, so anything set would be discarded. That predates this issue and is on the governed `fuse_scored_calibrated` path. Left alone as out of scope here — happy to file it separately or fold it in, whichever you prefer.

PR #1880, opened 2026-09-09
fix(engine): wire node positions into StreamingEngine's governed fuser

### LINKED ISSUE #1870: StreamingEngine's multistatic fuser is never given node positions
## Gap

`StreamingEngine` owns its own `MultistaticFuser` (`v2/crates/wifi-densepose-engine/src/lib.rs:116`) and nothing ever calls `set_node_positions` / `set_node_positions_by_id` on it. A repo-wide grep for either setter under `v2/crates/wifi-densepose-engine/` returns nothing.

Every `FusedSensingFrame` produced on this path therefore carries `node_positions` filled entirely with the origin, regardless of what the deployment configured. This is the **governed** path — `lib.rs:487` calls `fuse_scored_calibrated`, whose score feeds provenance and the witness — not a diagnostic side path.

A second, narrower hazard sits alongside it: `set_multistatic_config` (`lib.rs:239`) rebuilds the fuser wholesale —

```rust
pub fn set_multistatic_config(&mut self, cfg: MultistaticConfig) {
    self.fuser = MultistaticFuser::with_config(cfg);
}
```

so any positions that *were* set would be silently discarded by a later config override. The existing doc comment already warns callers to "call before any frames are processed"; that warning would need to cover positions too.

## Relationship to #1866

Separate. #1866 is about the `AppState` fuser in `wifi-densepose-sensing-server`, where the configured map existed but was flattened to a positional vector and addressed by cohort rank. That one is fixed in #1860. This one is a different instance of the fuser that is never configured at all, so it was not in scope and is not fixed by that change.

Both share a root cause worth noting: node positions are configured per fuser instance rather than owned once and handed to every consumer.

## Scope note

Software-only. No hardware-accuracy claim is made here, and no measurement of downstream impact is offered — `FusedSensingFrame::node_positions` currently has no reader in the tree (`geometric_diversity()` is defined but called only from its own unit tests), so this is a latent correctness gap rather than an observed output error.

## Suggested acceptance criteria

- `StreamingEngine` accepts node positions keyed by node id and passes them to its fuser.
- `set_multistatic_config` preserves configured positions across the rebuild, or the API makes the ordering requirement impossible to get wrong.
- A regression asserting a governed `fuse_scored_calibrated` frame carries each node's configured position, and one asserting positions survive a later `set_multistatic_config`.

ruvnet:
Fixes #1870.

## What is wrong

`StreamingEngine` owns its own `MultistaticFuser` instance — a different one from `wifi-densepose-sensing-server`'s `AppState`, which was fixed for this same class of bug in #1860. A repo-wide grep for `set_node_positions`/`set_node_positions_by_id` under `v2/crates/wifi-densepose-engine/` returned nothing: nothing ever configured this instance.

This is the **governed** path — `fuse_scored_calibrated` at `lib.rs:487` feeds provenance and the witness, not a diagnostic side path — so every `FusedSensingFrame` produced through it carried every node's position as the origin, regardless of what the deployment configured.

A second, narrower hazard sat alongside it: `set_multistatic_config` rebuilds the fuser wholesale (`self.fuser = MultistaticFuser::with_config(cfg)`), which would silently discard any positions configured before it, even after this fix.

## The fix

- `StreamingEngine::set_node_positions(HashMap<u8, [f32; 3]>)` — delegates to the fuser's existing `set_node_positions_by_id` (added in #1860).
- `set_multistatic_config` now snapshots the current positions before rebuilding the fuser and re-applies them after, so callers don't need to remember an ordering between the two setters.

## Verification

Two regression tests, matching the issue's own acceptance criteria:
- `governed_fuse_carries_each_nodes_configured_position` — calls `fuse_scored_calibrated` directly and asserts each fused contribution carries its own node's configured position (not the origin, not another node's position).
- `positions_survive_a_later_set_multistatic_config` — asserts a position set before a later `set_multistatic_config` call is still present after.

`cargo test -p wifi-densepose-engine`: 35/35 pass (33 pre-existing + 2 new).
`cargo check --workspace --no-default-features`: clean (only pre-existing, unrelated warnings).

### DIFF
--- a/v2/crates/wifi-densepose-engine/src/lib.rs
+++ b/v2/crates/wifi-densepose-engine/src/lib.rs
@@ -234,9 +234,30 @@ impl StreamingEngine {
     /// WiFi/ESP-NOW-synced multi-node deployments spuriously fail governed
     /// trust cycles even after widening the guard elsewhere.
     ///
-    /// Rebuilds the fuser, so call before any frames are processed.
+    /// Rebuilds the fuser, so call before any frames are processed. Preserves
+    /// any node positions configured via [`Self::set_node_positions`] across
+    /// the rebuild (issue #1870) — a caller that configures positions once at
+    /// startup and calls this later (e.g. after deriving a TDM schedule) must
+    /// not have that configuration silently discarded.
     pub fn set_multistatic_config(&mut self, cfg: MultistaticConfig) {
+        let positions = self.fuser.node_positions_by_id().clone();
         self.fuser = MultistaticFuser::with_config(cfg);
+        self.fuser.set_node_positions_by_id(positions);
+    }
+
+    /// Configure node positions for the governed multistatic fuser, keyed by
+    /// node id (issue #1870). Without this, `StreamingEngine`'s fuser — the
+    /// one behind `fuse_scored_calibrated`, which feeds provenance and the
+    /// witness — always fuses with every node at the origin, regardless of
+    /// deployment configuration, because nothing ever called this on it. This
+    /// is a separate `MultistaticFuser` instance from the one
+    /// `wifi-densepose-sensing-server`'s `AppState` owns (fixed for that
+    /// instance in #1860); each fuser instance needs positions set on it
+    /// independently. Safe to call before or after [`Self::set_multistatic_config`] —
+    /// order does not matter, since that method now preserves whatever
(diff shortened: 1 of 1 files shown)


oga35767-eng (approved):

oga35767-eng:
Yy

PR #1882, opened 2026-09-09
fix(sensing-server): don't let edge-vitals sentinel RSSI poison mean_rssi

ruvnet:
## Summary
- The ESP32 edge-vitals packet (magic `0xC511_0002`, ADR-039) carries a raw `i8` RSSI byte with no "valid" flag. When the edge pipeline hasn't sampled a real reading yet, it has been observed sending physically implausible near-zero sentinel values (`-1`, `-2` dBm).
- These sentinels were being pushed unguarded into the node's `rssi_history` and the room's fused `mean_rssi`, producing a one-tick reading of `-2.0` dBm on node 3 while that same node's CSI path simultaneously reported a real `~-50` dBm signal.
- Adds `is_plausible_rssi()` (mirroring the existing `update_csi_fps_ema()` plausibility-filter pattern already in this file) so an implausible RSSI falls back to the node's last known-good reading instead of overwriting it.
- Scope is intentionally narrow: this does not touch the vitals-only packet's `amplitude`/`subcarrier_count` overwrite behavior for the `nodes[]` array, nor the separate one-occupant vitals-authority gate — those are distinct, higher-risk areas that weren't part of the diagnosed root cause.

## Test plan
- [x] `cargo test -p wifi-densepose-sensing-server --no-default-features` — 908 tests pass, 0 failed (includes 3 new `rssi_plausibility_tests`)
- [x] Verified live against the running `ruv-mac-mini` sensing-server instance: node 3 currently reports a real `-50` dBm CSI RSSI while its edge-vitals path was observed emitting the `-2.0` dBm sentinel that this fix filters out

### DIFF
--- a/v2/crates/wifi-densepose-sensing-server/src/main.rs
+++ b/v2/crates/wifi-densepose-sensing-server/src/main.rs
@@ -819,6 +819,51 @@ mod debounce_room_classification_tests {
     }
 }
 
+/// Upper plausibility ceiling (dBm) for a real 2.4/5 GHz WiFi RSSI reading.
+///
+/// A received WiFi signal is always attenuated by free-space path loss and
+/// receiver noise floor; no deployed node has ever measured better than
+/// roughly -20 dBm. The ESP32 edge-vitals packet (magic 0xC511_0002,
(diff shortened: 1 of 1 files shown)


oga35767-eng:
Ty

PR #1889, opened 2026-09-10
fix(provision): stop persisting WiFi password world-readable (0644)

### LINKED ISSUE #1754: fix(provision): WiFi password persisted world-readable (0644) in per-port state file

ruvnet:
## Summary

Closes #1754. `provision.py` writes the WiFi password in cleartext to a per-port state file by design (see #391/#574), but `save_state()` created the directory and file with `os.makedirs()`/`open()` using no explicit mode, so both landed at whatever the umask allowed — `0755`/`0644` on a default macOS or Linux install. Any other local user or process on a shared machine could read every network the operator has provisioned.

## Fix

- `save_state()` now creates the state directory at `0700` and the state file at `0600`, using `os.open()` with an explicit mode so the temp file is never briefly world-readable between creation and a later `chmod`.
- Also tightens an existing looser directory, and any sibling `*.json` state files left behind by an earlier, unpatched version of this script — not just the file the current invocation happens to write.
- Left password persistence itself untouched (issue's own "Consider whether..." item is a separate design question, not required for the fix).

## Test plan
- [x] `python -m unittest tests.test_provision_state` — 14 tests pass (3 permission-assertion tests `skipIf(win32)`, since Windows has no POSIX permission bits to assert on)
- [x] Ran the same suite on a real POSIX host (Linux) to get genuine verification of the permission assertions rather than trusting the Windows skip — all 14 pass there, 0 skipped

### DIFF
--- a/firmware/esp32-csi-node/provision.py
+++ b/firmware/esp32-csi-node/provision.py
@@ -144,13 +144,42 @@ def load_state(port: str, state_dir: str) -> dict:
     return {}
 
 
+def _harden_state_dir(state_dir: str) -> None:
+    """Ensure state_dir is 0700 and tighten any state files an earlier,
+    unpatched version of this script may have left behind at 0644.
+
+    os.chmod is best-effort on platforms without POSIX permission bits
+    (Windows): it never raises for an unsupported mode there, it just has
+    a narrower effect (roughly the read-only attribute), so this is safe
+    to call unconditionally on every platform.
+    """
+    os.makedirs(state_dir, mode=0o700, exist_ok=True)
+    try:
+        os.chmod(state_dir, 0o700)
+    except OSError:
+        pass
+    try:
+        for name in os.listdir(state_dir):
+            if name.endswith(".json"):
+                try:
+                    os.chmod(os.path.join(state_dir, name), 0o600)
+                except OSError:
+                    pass
+    except OSError:
+        pass
+
+
 def save_state(port: str, state_dir: str, state: dict) -> str:
     """Write `state` to the per-port file, creating dirs as needed. Returns path."""
-    os.makedirs(state_dir, exist_ok=True)
+    _harden_state_dir(state_dir)
     path = _state_path_for(port, state_dir)
     # Sort keys for deterministic on-disk content (easier to diff).
     tmp = path + ".tmp"
-    with open(tmp, "w", encoding="utf-8") as f:
+    # Create the temp file pre-restricted (mode applies at creation, before
+    # any data is written) so the secret is never briefly world-readable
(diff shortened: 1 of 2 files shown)


ruvnet (commented):
REJECT at `8898a3f`. Fresh state-directory/file permissions improve, but the credential-lifecycle hypothesis remains false. `save_state()` reuses a fixed staging pathname without guaranteeing its mode when that path already exists; the creation mode does not repair an existing file. Legacy repair occurs too late to cover read-only and failure exits, while generator-failure and dry-run outputs can remain credential-bearing 0644 artifacts under the default umask. The CLI also retains argv/stdout credential exposure. Please use securely created unique/exclusive staging, enforce descriptor permissions before writing, move legacy repair before every read/early exit, restrict every intermediate artifact, add non-argv/non-stdout input and inspection paths, and run staging-reuse, failure, dry-run, ownership/type, Linux, and macOS regressions in exact-head CI.

PR #1845, opened 2026-09-06
fix: enable macOS RSSI sensing via CoreWLAN

faizanshah1404:
## Summary

Fixes RSSI sensing on macOS by using the existing CoreWLAN adapter instead of the Windows `netsh` pipeline.

- Routes Wi-Fi capture and auto-detection through `MacosCoreWlanScanner` on macOS
- Adds `--scan-once` support to the Swift `mac_wifi` helper
- Preserves the helper's existing streaming mode
- Adds a timeout so outdated/non-terminating helpers are cleaned up safely

## Testing

Tested on Apple Silicon macOS:

- `cargo test --locked -p wifi-densepose-wifiscan` — 155 passed
- sensing-server binary tests — 269 passed
- sensing-server library tests — 535 passed, 1 ignored
- optimized release build succeeded
- the rebuilt CoreWLAN helper returns finite connected-link RSSI/noise/channel data on Apple Silicon macOS
- with the identity-safety fix, the scanner abstains when macOS redacts both SSID and BSSID rather than inventing an ambiguous identity
- a separate Location-authorized AppKit diagnostic exposed both SSID and a non-zero BSSID on the same machine, identifying helper packaging/authorization as the remaining live-server integration blocker

This addresses connected-link RSSI acquisition and safe macOS Wi-Fi identity handling. Full live server integration still requires a Location-authorized helper packaging path; this does not add CSI support or establish person-count, pose, vital-sign, or other sensing accuracy.

## Related issues

Related to #56 and #99.

PR #1898, opened 2026-09-11
fix: correct ESP32 source detection and honest labelling of commodity-WiFi data

Liammcl:
## Summary

Two related fixes to how the sensing stack detects and labels its data source. Both were
found while bringing up a real ESP32-S3 node against `ruvnet/wifi-densepose:latest`, and both
cause the server or UI to misreport what the operator is actually looking at.

1. **`probe_esp32_udp()` only inspected the first datagram**, so a live, streaming ESP32 was
   detected as absent and the server silently fell back to the commodity RSSI collector.
2. **The UI labelled real commodity-WiFi measurements as `SIMULATED — NO HARDWARE`**, which is
   the inverse of the honesty rule the repo enforces everywhere else.

## 1. ESP32 detection (`archive/v1/src/sensing/ws_server.py`)

An `esp32-csi-node` interleaves several frame types on port 5005. Measured arrival order from
a live ESP32-S3 node:

```
#1  magic=0xC5110006   60B
#2  magic=0xC5110006   60B
#3  magic=0xC5110001  148B   <- CSI frame, third in line
#4  magic=0xC5110002   32B
```

Distribution over 25 s: `0xC5110006` ×125, `0xC5110001` ×33, `0xC5110002` ×24,
`0xC5110003` ×24. Since the probe judged only the first datagram, it returned `False` most of
the time. The node was healthy and streaming, but every broadcast frame carried
`"source": "macos_wifi"` with `subcarrier_count: 0` instead of real CSI.

The fix reads until the deadline and returns on the first CSI frame. The receive buffer also
grows from 256 to 2048 bytes — v0.8.8 firmware emits frames up to 404 B, which the old buffer
truncated.

Against the same node, the probe now returns `True` in 0.54 s (the time spent skipping non-CSI
datagrams) and the server correctly selects `Esp32UdpCollector`.

## 2. Data-source labelling (`ui/`)

`ws_server.py` emits `macos_wifi` / `linux_wifi` / `windows_wifi` for platform adapters. The UI
whitelist only recognised `esp32` / `wifi` / `live`, so all three fell through to
`server-simulated`.

Rather than folding these into an existing state, this adds a third one. Promoting them to
`live` would be equally dishonest — they carry no CSI — while `server-simulated` denies they
were measured at all:

| source | state | banner |
|---|---|---|
| `esp32` | `live` | `LIVE — ESP32 HARDWARE` |
| `macos_wifi` / `linux_wifi` / `windows_wifi` | `rssi-only` | `LIVE RSSI — REAL DATA, NO CSI` |
| `simulated` | `server-simulated` | `SIMULATED — NO HARDWARE` |

The "About This Data" copy is now source-aware. The default text claims metrics come from CSI
with *N* ESP32 nodes and offers breathing estimation — none of which holds for an RSSI-only
source, where the signal is one aggregate power scalar at 1 dBm quantisation.

This is a follow-up to #1845, which added the macOS CoreWLAN collector without updating the UI
whitelist. Linux and Windows collectors were already affected before that.

## Verification

Source mapping, including the conservative defaults and ADR-295 precedence:

```
esp32                        -> live
macos_wifi                   -> rssi-only
linux_wifi                   -> rssi-only
windows_wifi                 -> rssi-only
simulated                    -> server-simulated
<unknown>                    -> server-simulated   (conservative default kept)
macos_wifi + state=synthetic -> server-simulated   (ADR-295 precedence kept)
```

Repository validation matrix for the touched areas:

```
$ python archive/v1/data/proof/verify.py
    Computed: f8e76f21a0f9852b70b6d9dd5318239f6b20cbcb4cdd995863263cecdc446f7a
    Expected: f8e76f21a0f9852b70b6d9dd5318239f6b20cbcb4cdd995863263cecdc446f7a
    Status:   MATCH (bit-exact)
  VERDICT: PASS

$ cd archive && python -m pytest v1/tests/unit/test_sensing.py -q
45 passed
```

Hardware used: Waveshare ESP32-S3 AMOLED 1.8" (SH8601), 16 MB flash / 8 MB PSRAM, chip rev
v0.2, on a 2.4 GHz AP (channel 1), streaming to `ruvnet/wifi-densepose:latest` with
`CSI_SOURCE=esp32`.

`ui/utils/data-source-banner.js` intentionally needs no change: it already hides its global
warning bar for unrecognised states, and that bar warns "nothing on this screen is current",
which does not apply to real RSSI.

## Note

While bringing this node up I also hit a separate, unrelated problem: the pre-built binaries in
`firmware/esp32-csi-node/release_bins/` are several releases behind the published GitHub
releases, and on an ESP32-S3 with a display panel they yield ~1 pps — below
`CALIBRATION_GRID_MIN_RATE_HZ = 2.0`, so room calibration can never start. Flashing the
published v0.8.8 bundle to the same board raised it to 34 pps. That is a firmware/packaging
matter rather than a code change, so I have kept it out of this PR and will file it separately.

### DIFF
--- a/archive/v1/src/sensing/ws_server.py
+++ b/archive/v1/src/sensing/ws_server.py
@@ -215,14 +215,23 @@ def probe_esp32_udp(port: int = ESP32_UDP_PORT, timeout: float = 2.0) -> bool:
     """Return True if an ESP32 is actively streaming on the UDP port."""
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-    sock.settimeout(timeout)
     try:
         sock.bind(("0.0.0.0", port))
-        data, _ = sock.recvfrom(256)
-        if len(data) >= 20:
-            magic = struct.unpack_from('<I', data, 0)[0]
-            return magic == 0xC5110001
-        return False
+        # Keep reading until the deadline instead of judging the first datagram.
+        # An esp32-csi-node interleaves several frame types on this port (CSI
+        # 0xC5110001, sync 0xC5110002, feature 0xC5110006), so the first packet
+        # to arrive is frequently *not* the CSI frame. Returning False on it
+        # made a live, streaming node look absent and silently demoted the
+        # server to the commodity-RSSI collector.
+        deadline = time.monotonic() + timeout
+        while True:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                return False
+            sock.settimeout(remaining)
+            data, _ = sock.recvfrom(2048)
+            if len(data) >= 20 and struct.unpack_from('<I', data, 0)[0] == 0xC5110001:
+                return True
     except (socket.timeout, OSError):
         return False
     finally:
--- a/ui/components/SensingTab.js
+++ b/ui/components/SensingTab.js
@@ -108,7 +108,7 @@ export class SensingTab {
           <!-- Setup info -->
           <div class="sensing-card">
             <div class="sensing-card-title">About This Data</div>
-            <p class="sensing-about-text">
+            <p class="sensing-about-text" id="sensingAboutText">
               Metrics are computed from WiFi Channel State Information (CSI).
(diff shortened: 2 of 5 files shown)


navghanthakor524-hub (commented):
banner.textContent = cfg.text;

Issue #27, opened 2026-02-16
Has a single person tried running it? Even the author?


CelestialCreator:
Needs a clarity on how to setup what all wifi devices are supported, how will one way communication solve this
This kind of stuffy is destroying the community trust on git trendings

axgoodhart:
Why bother? it's entirely fictional.

hoangnv170752:
this is the reason why github need downvote button

manhtai:
Try to run it then see this issue, what a joke

PR #1904, opened 2026-09-11
fix(firmware): restore native C6 onboarding delivery

proffesor-for-testing:
## Outcome

Restores the Mac USB/JTAG onboarding contract and makes firmware releases consumable as exact artifacts.

## Changes

- make USB Serial/JTAG the primary ESP32-C6 console so `RUVIEW_HELLO_V1` reaches stdin;
- remove the ineffective `/dev/secondary` reader (ESP-IDF secondary consoles are output-only);
- replace the tracked C6 onboarding bundle so it identifies itself as `0.8.12`;
- preserve PR #1891's partition-relative size budgets, allowing the S3 8 MB build to use its real 2 MB OTA slot while retaining 20% headroom;
- publish exact-run S3 8 MB, S3 4 MB, and C6 4 MB archives on dedicated `vX.Y.Z-esp32` releases;
- document that automated `vNNNN` server releases such as v2655 are not firmware releases;
- add a static delivery contract to the existing host-test gate.

## Verification

- red-first contract reproduced both faults: C6 was not configured for primary native USB and the tracked app embedded `0.8.11`;
- `make host_tests` passes;
- local ESP-IDF 6.0.2 C6 build passes and generated sdkconfig contains `CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG=y` plus `CONFIG_ESP_CONSOLE_SECONDARY_NONE=y`;
- tracked-bundle SHA-256 manifest verifies and the app embeds `0.8.12` and `RUVIEW_HELLO_OK_V1`;
- workflow YAML parses and `git diff --check` passes.

The tracked binary will be reconciled once more against this PR's exact ESP-IDF 5.4 CI artifact before merge. Physical HELLO and configure verification remains an explicit post-build gate.


proffesor-for-testing:
Exact-head verification for 2277b88a44873d0e152a604ee6d7c4231b08282e:

- all GitHub checks are terminal with no failures; both ESP32-C6 builds and both ESP32-S3 size variants pass;
- local host delivery-contract tests and manifest verification pass;
- the CI-built C6 bundle identifies as 0.8.12; app size 1,078,224 bytes;
- the CI-built S3 4 MB app is 975,728 bytes and its manifest verifies.

Merge remains gated on the requested physical native USB/JTAG RUVIEW_HELLO_V1 check. Release assets intentionally publish only from a dedicated vX.Y.Z-esp32 tag; the unrelated server tag v2655 is not being retroactively populated.

proffesor-for-testing:
Physical native USB/JTAG gate passed at exact head 2277b88a44873d0e152a604ee6d7c4231b08282e on 2026-09-11.

- identified /dev/cu.usbmodem14401 from ROM as ESP32-C6 revision 0.2 in USB-Serial/JTAG mode;
- flashed the exact CI artifact: bootloader, C6 partition table, initial OTA data, and 1,078,224-byte application;
- esptool verified the written hash for every region;
- the running firmware answered the nonce-bound command with RUVIEW_HELLO_OK_V1, chip=esp32c6, version=0.8.12, configured=1, and the matching nonce.

This closes the physical HELLO gate. The later RuView Mac qualification will separately exercise the complete in-app Scan -> Flash -> HELLO -> Configure -> server-verification journey.

proffesor-for-testing:
2026-09-11 session handoff at exact head `2277b88a44873d0e152a604ee6d7c4231b08282e`: all checks remain terminal green. The direct physical native USB/JTAG C6 gate passed, and mobile PR #27 has since merged the official-variant filename correction used by the Mac picker. A new Release Catalyst build from mobile main `0c188f5b` is running as one UI PID plus one same-bundle helper.

Not yet claimed: the complete new-build in-app flow on all three C6 devices, any physical S3 result, or final three-C6 calibration/output evidence. One C6 carries the exact 0.8.12 CI image; the other two remain for in-app flashing. No S3 is attached. PR #1904 remains open for the next session decision.

proffesor-for-testing:
Rebased onto `main` (`45fc41ac`). Head is now `0c1efda4`; was DIRTY, now mergeable.

**One commit dropped as redundant.** `9907f062 fix(ci): size-gate each firmware variant against its own OTA slot` loosened all three gates:

```
             was     this PR proposed
8mb         1152  ->  1638
4mb         1152  ->  1484
c6-4mb      1152  ->  1484
```

Since then `af2d6981` landed on main and made a narrower change — raising only the 8 MB gate to **1200 KiB**, with the reasoning recorded inline ("Current 8 MB image is 1163 KiB after USB onboarding. The OTA app partition is 2048 KiB"), and leaving 4 MB and C6 at 1152.

Main's values already cover this PR's own artifacts, so the commit had nothing left to do:

```
C6 onboarding app   1,078,224 bytes = 1052 KiB   vs c6-4mb gate 1152  OK
S3 8 MB CI artifact 1,191,440 bytes = 1163 KiB   vs 8mb gate    1200  OK
```

Taking main's tighter bound rather than re-applying the looser one — it is the newer deliberate decision and it is sufficient here. The firmware delivery commits (`b1be9147`, `0c1efda4`) rebased cleanly.

**Artifacts verified unchanged after the rebase**, against the tracked `SHA256SUMS.txt`:

```
bootloader.bin         19,456 B   12e9fe5bd0533c3e…  OK
partition-table.bin     3,072 B   4c2cc4ffd52641e2…  OK
esp32-csi-node.bin  1,078,224 B   baff3940ede0ce16…  OK
ota_data_initial.bin    8,192 B                      OK
```

That application binary is the one flashed and verified on three physical ESP32-C6 boards from the Mac app — every region MD5-checked on readback, and the boards kept their node IDs and configuration through the flash.

PR #1909, opened 2026-09-12
feat(calibration): per-node ESP32 enrollment/calibration and a presence gate that works on real hardware

jingilibili:
# Per-node ESP32 enrollment/calibration + a presence gate that works on real hardware

Real-hardware work on three ESP32-S3 CSI nodes (ESP-IDF v5.4, ADR-151 Stage 2).
Enrollment was rejecting every person-present anchor; the investigation below
found three independent defects in the enrollment path, plus a gate that was
testing a quantity that cannot separate an empty room from an occupied one.

### Calibration / CLI (`wifi-densepose-cli`)

* **`calibrate --node-id` / `--all-nodes`** — a single-node baseline, or one
  independent baseline per discovered node written as
  `baseline-node<N>.bin`. Each node gets its own `CalibrationRecorder`.
* **`enroll --node-id` / `--all-nodes`** — one shared capture per anchor: the
  operator performs each pose once and every configured node records it.
  Adds `--baseline-dir`, `--output-dir`, `--lead-in-s` (countdown before each
  capture, so the operator can reach the pose) and gate overrides
  (`--min-presence-z`, `--empty-max-z`, `--max-still-motion`, `--min-move-motion`,
  `--min-mean-shift`, `--presence-votes`).
* **Fix: the per-anchor acceptance map is reset for every anchor.** It was
  declared outside the anchor loop, so once a node accepted anchor 1 it stayed
  "accepted" and every later anchor skipped it. A live run finished with
  `1/8 anchors accepted` and *no reject reasons at all* for anchors 2..8.
* **Fix: frames whose subcarrier count differs from the baseline are skipped.**
  A 192-bin baseline must never consume 64/128-bin frames — that compares
  different physical bins and inflates the deviation score. This was the
  failure mode behind inflated `presence_z` on some nodes.
* **`--lead-in-s`** replaces the hard-coded 3 s countdown, which is not enough
  time to move into position between anchors.

### Signal crate (`wifi-densepose-signal`)

* `CalibrationConfig::ht40_192()`, `ht40_128()`, `ht40_306()` — the raw
  subcarrier counts an ESP32-S3 actually delivers on HT40 legacy-PPDU capture
  (measured 306 bins dominant when the link is busy, 192 when it is not).
  Without a matching tier the recorder rejects every frame
  (`subcarrier count mismatch`).

### Presence gate (`wifi-densepose-calibration`)

* `AnchorQualityGate` gains `min_mean_shift`, and `evaluate()` takes an
  optional relative mean-amplitude shift, compared by **magnitude**. Passing
  `None`, or leaving `min_mean_shift` at its `0.0` default, keeps the existing
  `presence_z` behaviour exactly, so no current caller changes.
* `enroll` measures the shift against **the session's own `empty` capture**, so
  slow baseline drift no longer dominates the decision.
* `enroll --all-nodes` adds **room-level presence consensus**
  (`--presence-votes`, default 1): a person is a property of the room, not of
  one link, so a node with an insensitive geometry cannot veto an anchor that
  other nodes clearly see. Motion is still enforced per node.

## Why: what the data showed

Three nodes, ~1300 frames per 30 s phase, alternating empty room and operator
standing still, `192`-subcarrier frames only. All figures below are **MEASURED**
on the captured data.

**1. `presence_z` cannot separate the two states.** It is the mean over frames
of the *median across subcarriers* of `|amp - baseline| / sigma`, i.e. a
frame-to-frame jitter statistic. It read ≈ `0.8` for the empty room *and* for
the occupied room, so no threshold on it works:

| node | empty | occupied |
|---|---|---|
| 1 | 0.82 | 1.08 – 1.48 |
| 2 | 0.97 | 1.23 – 1.45 |
| 3 | 0.82 | 1.05 – 2.47 |

**2. The capture mean amplitude carries the presence signal.** It shifted by
`0.4 % – 13.7 %` when a person was present, with a standard error of
~`0.08` amplitude units over a capture — 6 to 12 sigma:

| node | empty (192-bin) | standing still | shift |
|---|---|---|---|
| 1 | 31.476 ± 1.844 | 31.944 ± 1.522 | +0.47 |
| 2 | 29.664 ± 2.371 | 30.432 ± 2.472 | +0.77 |
| 3 | 28.325 ± 3.133 | 30.096 ± 3.578 | +1.77 |

**3. The sign is not stable** (multipath phase change), which is why the test
uses the magnitude: across a full enrollment the same node produced shifts from
`-13.7 %` to `+3.7 %` for the same room state.

**4. Sensitivity is per-link geometry.** Node 1 (RSSI `-44 dBm`, dominant direct
AP path) responded by only `0.0 – 2.4 %`, while nodes 2/3 reached `-13.7 %` to
`+3.7 %`. A per-node veto therefore fails; hence consensus.

**5. The old empty-room gate is not stable either.** Within an hour of
calibration the empty-room `presence_z` drifted above `empty_max_z = 1.0`
(observed `1.05 – 2.17`), so an absolute z threshold rejects a genuinely empty
room.

A second, independent observation on the same captures: `192`-bin frames arrive
from several transmitters (per-frame RSSI spread 9-32 dB, per-subcarrier CV
~23 %), whereas `306`-bin frames come from one transmitter (RSSI spread 0-1 dB,
CV ~11-15 %). Mixing them in one baseline is another source of noise; the
geometry guard above prevents the worst case.

## Validation

```
cargo test -p wifi-densepose-cli --lib    -> 37 passed
cargo test -p wifi-densepose-calibration  -> 78 passed
cargo build -p wifi-densepose-cli         -> ok (dev profile)
```

No simulator was used: the numbers above come from live ESP32-S3 captures on
the target hardware, and the gate changes are additive (legacy behaviour is the
default) so no existing test needed changing.

## Not included in this PR (and why)

* Firmware `acquire_csi_ht40 = 0`: measured to be a **no-op on the ESP32-S3**.
  That bitfield only exists on `CONFIG_SOC_WIFI_HE_SUPPORT` targets; the S3
  compiles the legacy `wifi_csi_config_t` bool layout
  (`ltf_merge_en = true`, `channel_filter_en = false`). Dropped rather than
  shipped as dead code.
* A Windows/WSL UDP relay and local test-data JSON: environment-specific.

## Known limitations

* The consensus rule makes the gate depend on at least one sensitive link.
  A deployment where every node has a strong direct path would still need the
  geometry fixed rather than the gate relaxed.
* `min_mean_shift` is a relative threshold (0.75 % worked for these three
  nodes). It is exposed as a flag because it is site-dependent and should be
  calibrated per installation.
* The `Delta`-z motion measure is noise-dominated on this hardware (empty room
  52-66 %, still person 63-78 %), so `--max-still-motion` needs raising in
  practice. A motion statistic derived from the same mean-shift channel would
  be the natural follow-up.

### DIFF
--- a/v2/crates/wifi-densepose-calibration/src/enrollment.rs
+++ b/v2/crates/wifi-densepose-calibration/src/enrollment.rs
@@ -30,6 +30,16 @@ use crate::anchor::{Anchor, AnchorLabel, AnchorQuality};
 pub struct AnchorQualityGate {
     /// Minimum mean amplitude z-score to consider a person present.
     pub min_presence_z: f32,
+    /// Minimum **relative** mean-amplitude shift versus the session's own
+    /// `empty` capture that counts as "a person is present" (0.0075 = 0.75%).
+    ///
+    /// The per-subcarrier `presence_z` test measures frame-to-frame jitter, not
+    /// the presence shift: live ESP32-S3 data showed a person standing still
+    /// shifting the capture mean by +0.5..+1.8 amplitude units (6-12 sigma over
+    /// ~1300 frames) while `presence_z` stayed flat at ~0.8 for both the empty
+    /// and the occupied room. When `mean_shift_rel` is supplied to `evaluate`,
+    /// it replaces the z test. `0.0` keeps the legacy z behaviour.
+    pub min_mean_shift: f32,
     /// For `empty`: maximum mean z-score to consider the room truly empty.
     pub empty_max_z: f32,
     /// For "still" anchors: maximum motion-flag rate tolerated.
@@ -45,6 +55,7 @@ impl Default for AnchorQualityGate {
         Self {
             min_presence_z: 1.5,
             empty_max_z: 1.0,
+            min_mean_shift: 0.0,
             max_still_motion: 0.6,
             min_move_motion: 0.3,
             min_frames: 60,
@@ -61,6 +72,7 @@ impl AnchorQualityGate {
         presence_z: f32,
         motion_rate: f32,
         frames: u32,
+        mean_shift_rel: Option<f32>,
     ) -> (AnchorQuality, Option<String>) {
         let mut reason: Option<String> = None;
 
@@ -70,11 +82,24 @@ impl AnchorQualityGate {
                 self.min_frames
             ));
         } else if label.expects_presence() {
-            if presence_z < self.min_presence_z {
-                reason = Some(format!(
-                    "no person detected (presence_z {presence_z:.2} < {:.2}) — move closer / face the sensor",
-                    self.min_presence_z
-                ));
+            // Magnitude, not sign: a person changes the multipath phase, so the
+            // capture mean can move either way (live data: -13.7%..+3.7%).
+            let present = match mean_shift_rel {
+                Some(shift) => shift.abs() >= self.min_mean_shift,
+                None => presence_z >= self.min_presence_z,
+            };
+            if !present {
+                reason = Some(match mean_shift_rel {
+                    Some(shift) => format!(
+                        "no presence shift (|capture mean {:+.2}%| vs this session's empty capture, need >= {:.2}%) — move into the sensed area",
+                        100.0 * shift,
+                        100.0 * self.min_mean_shift
+                    ),
+                    None => format!(
+                        "no person detected (presence_z {presence_z:.2} < {:.2}) — move closer / face the sensor",
+                        self.min_presence_z
+                    ),
+                });
             } else if label.expects_still() && motion_rate > self.max_still_motion {
                 reason = Some(format!(
                     "too much motion ({:.0}% > {:.0}%) for a still anchor — hold still",
@@ -203,12 +228,18 @@ impl AnchorRecorder {
 
     /// Evaluate the capture against the gate and produce an `Anchor` (accepted
     /// or not) plus a rejection reason.
-    pub fn finalize(&self, gate: &AnchorQualityGate, at_unix_s: i64) -> (Anchor, Option<String>) {
+    pub fn finalize(
+        &self,
+        gate: &AnchorQualityGate,
+        at_unix_s: i64,
+        mean_shift_rel: Option<f32>,
+    ) -> (Anchor, Option<String>) {
         let (quality, reason) = gate.evaluate(
             self.label,
             self.presence_z(),
             self.motion_rate(),
             self.frames,
+            mean_shift_rel,
         );
         (
             Anchor {
@@ -244,7 +275,7 @@ mod tests {
         for &z in zs {
             r.record_score(&score(z));
         }
-        r.finalize(&AnchorQualityGate::default(), 100)
+        r.finalize(&AnchorQualityGate::default(), 100, None)
     }
 
     /// Constant z (a perfectly still capture at the given presence strength).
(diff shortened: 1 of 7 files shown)


jingilibili:
## Summary of the changes in this PR

| Area | File | What |
|---|---|---|
| CLI | `wifi-densepose-cli/src/calibrate.rs` | `--node-id` + `--all-nodes` (one independent baseline per node, `baseline-node<N>.bin`), `--output-dir`, and the `esp32_ht40_192/128/306` tiers |
| CLI | `wifi-densepose-cli/src/room.rs` | `enroll --node-id/--all-nodes` (one shared capture per anchor), `--baseline-dir`, `--output-dir`, `--lead-in-s`, gate overrides, per-anchor acceptance-map reset, subcarrier geometry guard, session-relative presence reference, room-level presence consensus |
| CLI | `wifi-densepose-cli/src/calibrate_api.rs` | accept the new tiers, thread the optional shift through |
| Signal | `wifi-densepose-signal/src/ruvsense/calibration.rs` | `CalibrationConfig::ht40_192()`, `ht40_128()`, `ht40_306()` |
| Calibration | `wifi-densepose-calibration/src/enrollment.rs` | `AnchorQualityGate::min_mean_shift` + magnitude-based mean-shift presence test (opt-in) |

### Defects fixed

1. **Anchors 2..N were never evaluated.** The per-anchor acceptance map was declared outside the anchor loop, so once a node accepted `empty` it stayed accepted and every later anchor skipped it. A live multi-node run finished `1/8 anchors accepted` with no reject reasons at all for anchors 2..8.
2. **A 192-bin baseline consumed 64/128-bin frames.** `deviation()` accepts a frame whose count matches either the stored count or the tier's canonical count, so 128-bin frames were compared against 192-bin statistics. Frames that do not match the baseline are now skipped, with a count in the log.
3. **The presence gate tested the wrong quantity.** `presence_z` is a mean-of-median jitter statistic: it read ≈ `0.8` for an empty room *and* for an occupied room, so no threshold on it can work.
4. **No tier matched what the hardware delivers.** An ESP32-S3 on legacy-PPDU HT40 capture emits `192` bins when the channel is quiet and `306` when it is busy, so the recorder rejected most frames as `subcarrier count mismatch`.

### Measured evidence (3 ESP32-S3 nodes, ESP-IDF v5.4, ~1300 frames per phase)

* `presence_z` for empty vs occupied: `0.82` vs `1.08-1.48` (node 1), `0.97` vs `1.23-1.45` (node 2), `0.82` vs `1.05-2.47` (node 3) — overlapping.
* Capture mean amplitude shift with a person present: `+0.47` (node 1), `+0.77` (node 2), `+1.77` (node 3), standard error ≈ `0.08` → 6-12 sigma.
* The shift flips sign across captures (`-13.7 %` … `+3.7 %`) because a person changes the multipath phase, hence the magnitude test.
* Per-link sensitivity differs: node 1 (RSSI `-44 dBm`, dominant direct AP path) moved only `0.0-2.4 %` while nodes 2/3 reached `-13.7 %` … `+3.7 %`, hence consensus instead of a per-node veto.
* Empty-room `presence_z` drifted to `1.05-2.17` within an hour of calibration against `empty_max_z = 1.0`.

### Tests

```
cargo test -p wifi-densepose-cli --lib     -> 37 passed
cargo test -p wifi-densepose-calibration   -> 78 passed
cargo build -p wifi-densepose-cli          -> ok
```

### Compatibility

The gate change is additive: `min_mean_shift` defaults to `0.0`, and passing `None` keeps the previous `presence_z` path unchanged, so no existing caller or test needed to change. Every new CLI flag also defaults to the previous behaviour.

PR #1911, opened 2026-09-12
fix(ui): make the live CSI view reach LIVE against the bundled sensing server

jingilibili:
## What this fixes

The pose-fusion UI (`/ui/pose-fusion.html`) served by the bundled sensing server
stayed on watermarked **SYNTHETIC** data even with three ESP32-S3 nodes
streaming CSI at ~30 fps per node. Three independent defects sat on that path;
this PR fixes the two that live in the repository and adds the server-side
diagnostics needed to see them.

### 1. `ui/pose-fusion/js/main.js` — the WS port guess never matched the bundled server

The page derived `ws://<host>:<httpPort + 1>/ws/sensing`, which is the Docker
`3000 -> 3001` convention. The bundled sensing server serves HTTP + UI on
**8080** and the sensing WebSocket on **8765** (`main.rs` binds `args.ws_port`,
default `8765`), so the guess resolved to `8081`. `connectLive()` failed
silently and the watermarked demo view stayed up.

`ui/services/sensing.service.js` already encodes the correct mapping
(`{ '3000': '3001', '8080': '8765' }`); this PR uses the same table in
pose-fusion and keeps `httpPort + 1` only as the fallback for host-port
remappings the table does not know.

### 2. `ui/pose-fusion/js/csi-simulator.js` — an amplitude-stripped cycle was treated as a verified frame

```js
const ampArr = node.amplitude || msg.amplitude;
if (ampArr && Array.isArray(ampArr)) { ...; this._markVerifiedFrame(); }
```

`Array.isArray([])` is `true`, and the live buffers were reset to zeros at the
top of `_handleJsonFrame` on **every** `sensing_update`. So a cycle that
carried no CSI still promoted the view to LIVE *and* blanked the amplitude
buffers, making the live readout show `0.00`.

Those cycles are legitimate: a governed cycle emitted at privacy class
`Restricted` strips the per-node raw amplitude/phase proxies by design
(ADR-141 output mapping — `trust.raw_outputs_suppressed: true` on
`/api/v1/status`).

**MEASURED** on this array (channel 11, 192/306/128/64-bin CSI over UDP,
~30 fps per node, fusion `soft_guard_us` at its 20 ms default):
**316 of 689** `sensing_update` messages (**45.9 %**) arrived with
`"amplitude": []` and `"subcarrier_count": 0`. `/api/v1/status` flipped
between `effective_class: "Anonymous", demoted: false` and
`effective_class: "Restricted", demoted: true` roughly once per second
(`engine_error_count: 0`, so this is a tolerated-contradiction demotion, not an
engine failure). The trigger is `ContradictionFlag::TimestampMismatch` in
`wifi-densepose-signal`'s `fuse_scored_calibrated`: it compares the **host
arrival** timestamps of the newest frame per node against `soft_guard_us`.
With three nodes each emitting one frame per ~33 ms, that spread is
approximately uniform over one frame interval, so a 20 ms soft guard sits right
at the median — ~46 % of cycles exceed it.

The fix requires a **non-empty** amplitude/phase/IQ payload before decoding,
keeps the last verified frame when a cycle carries no decodable CSI, and moves
the RSSI/presence refresh into `_updateSensingMetadata` so a suppressed cycle
still updates metadata without touching the verified frame.

### 3. `wifi-densepose-sensing-server/src/main.rs` — make the disconnect reason visible

`handle_ws_client` broke out of its `select!` on six distinct conditions and
logged none of them, so a dropped `/ws/sensing` stream was indistinguishable
from one that never opened. Each break now logs its concrete reason, and a
client `Ping` is answered with a `Pong`. No other behaviour change.

## Reproducer

```bash
# 1. run the sensing server with the fusion soft guard matched to the array's
#    inter-frame interval (see "Operator note" below; 20 ms default still works,
#    it just strips raw amplitudes on ~46% of cycles)
cd v2
OPENBLAS_SYSTEM=1 WDP_SOFT_GUARD_US=55000 cargo run -p wifi-densepose-sensing-server -- \
  --source esp32 --udp-port 5005 --udp-bind 0.0.0.0 --udp-allow 172.22.96.0/20 \
  --bind-addr 0.0.0.0 --ui-path ../ui

# 2. open the UI in a browser (no manual WebSocket URL needed after this PR)
#    http://localhost:8080/ui/pose-fusion.html
#    -> status label "LIVE CSI", button "Live ESP32", RSSI from the node
```

Raw-message check used for the 45.9 % figure (a minimal WS client that counts
per-node `amplitude` lengths) reported, over a 14 s window at the 20 ms
default: `total=689 full_amplitude=373 suppressed=316`. With
`WDP_SOFT_GUARD_US=55000`: `total=882 full_amplitude=860 suppressed=22`
(**2.5 %**), `/api/v1/status` steady at `raw_outputs_suppressed: false`.

## Verification of the UI change

Headless Chrome, fresh profile, against the bundled server, no manual WS URL:

```
<span id="status-label">LIVE CSI</span>
<button class="btn active" id="connect-ws-btn">&#10003; Live ESP32</button>
<span class="rssi-dbm" id="rssi-value">-54 dBm</span>
<span id="fps-display" class="fps-badge">31 FPS</span>
[CONSOLE] "[PoseFusion] frame 1 OK — mode=dual, csi.bufLen=1, embPts=1"
```

Reproduced twice in two independent browser profiles.

## Operator note (not part of this PR)

`WDP_SOFT_GUARD_US` is an existing override, and the upstream relation asserted
by `wifi-densepose-signal` is `stagger < soft_guard_us < guard_interval_us`. A
20 ms soft guard assumes nodes whose frames arrive within 20 ms of each other
(>50 fps per node); an ESP32-S3 CSI array at ~30 fps cannot satisfy that, so
half of the governed cycles are demoted to `Restricted` and lose their raw
amplitudes. A follow-up worth considering upstream: log a rate-limited warning
(or expose the demotion reason on `/api/v1/status`) when a cycle is demoted by
`TimestampMismatch`, instead of silently stripping the raw proxies.

## Scope

- No secrets, CSI captures, or personal data. No calibration/baseline artifacts.
- Rust change is log/diagnostic only; `cargo check -p wifi-densepose-sensing-server` clean.
- Untested in CI on this fork: the UI files are static JS covered by the
  headless-Chrome check above, not by a JS test suite.

### DIFF
--- a/ui/observatory/js/main.js
+++ b/ui/observatory/js/main.js
@@ -437,40 +437,101 @@ class Observatory {
   // ---- WebSocket live data ----
 
   _autoDetectLive() {
-    // Probe sensing server health on same origin, then common ports
+    // Probe by opening a WebSocket, not by fetching `${base}/health`.
+    //
+    // MEASURED failure of the previous HTTP-probe version (bundled sensing
+    // server, three live ESP32 nodes):
+    //
+    //   6.59s  fetch OK 200 4237ms http://localhost:8080/health
+    //   7.89s  Access to fetch at 'http://localhost:8765/health' blocked by CORS policy
+    //   9.61s  [Observatory] No sensing server detected, using demo mode
+    //
+    // The same-origin `/health` did answer, but only after 4.2 s: this page
+    // blocks its main thread building the three.js/WebGL scene, and the probe's
+    // own deadline was `AbortSignal.timeout(1500)`, so the one reachable
+    // candidate was aborted before it completed. The second candidate
+    // (`:8765/health`) is CORS-blocked — that router serves /health with no
+    // CORS headers — and the third (`:3000`) is a Docker-era HTTP port that
+    // never serves a WebSocket. Every load ended in DEMO.
+    //
+    // A WebSocket handshake is not subject to CORS and completes off the main
+    // thread, so open one and require an actual frame before claiming live.
     const host = window.location.hostname || 'localhost';
-    const candidates = [
-      window.location.origin,                   // same origin (e.g. :3000)
-      `http://${host}:8765`,                     // default WS port
-      `http://${host}:3000`,                     // default HTTP port
-    ];
-    // Deduplicate
-    const unique = [...new Set(candidates)];
+    const proto = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+    // Same HTTP -> WS mapping the other UI services use (ui/services/sensing.service.js).
+    // The bundled server serves HTTP/UI on 8080 and the sensing WebSocket on 8765;
+    // its HTTP router also answers /ws/sensing, so the page's own port is a valid
+    // second candidate. 3000 -> 3001 is the Docker image.
+    const WS_PORT_BY_HTTP_PORT = { '3000': '3001', '8080': '8765' };
+    const ports = [
+      WS_PORT_BY_HTTP_PORT[String(window.location.port)],
+      String(window.location.port),
+      '8765',
+      '3001',
+    ].filter(Boolean);
+    const candidates = [...new Set(ports.map(p => `${proto}//${host}:${p}/ws/sensing`))];
+    void this._tryLiveCandidate(candidates, 0);
+  }
 
-    const tryNext = (i) => {
-      if (i >= unique.length) {
-        console.log('[Observatory] No sensing server detected, using demo mode');
+  /**
+   * Open `candidates[i]`, adopt it on the first real frame, else move on.
+   * A socket that opens but never delivers a frame is not live data, so the
+   * first frame — not `onopen` — is what promotes the view (ADR-295).
+   */
+  _tryLiveCandidate(candidates, i) {
+    if (i >= candidates.length) {
+      console.log('[Observatory] No sensing server detected, using demo mode');
+      return;
+    }
+    const candidate = candidates[i];
+    // ADR-272: mint a single-use ticket when the deployment has auth on; the
+    // helper returns the URL unchanged when no token is configured.
+    Promise.resolve(withWsTicket(candidate)).catch(() => candidate).then((url) => {
+      let ws;
+      try {
+        ws = new WebSocket(url);
+      } catch {
+        this._tryLiveCandidate(candidates, i + 1);
         return;
       }
-      const base = unique[i];
-      fetch(`${base}/health`, { signal: AbortSignal.timeout(1500) })
-        .then(r => r.ok ? r.json() : Promise.reject())
-        .then(data => {
-          if (data && data.status === 'ok') {
-            const wsProto = base.startsWith('https') ? 'wss:' : 'ws:';
-            const urlObj = new URL(base);
-            const wsUrl = `${wsProto}//${urlObj.host}/ws/sensing`;
-            console.log('[Observatory] Sensing server detected at', base, '→', wsUrl);
-            this.settings.dataSource = 'ws';
-            this.settings.wsUrl = wsUrl;
-            void this._connectWS(wsUrl);
-          } else {
-            tryNext(i + 1);
-          }
-        })
-        .catch(() => tryNext(i + 1));
+      let settled = false;
+      const next = () => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        try { ws.close(); } catch { /* already gone */ }
+        this._tryLiveCandidate(candidates, i + 1);
+      };
+      const timer = setTimeout(next, 4000);
+      ws.onopen = () => console.log('[Observatory] Sensing WebSocket open at', url);
+      ws.onmessage = (evt) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        try { this._liveData = JSON.parse(evt.data); } catch { /* non-JSON frame */ }
+        console.log('[Observatory] Sensing server verified at', url);
+        this.settings.dataSource = 'ws';
+        this.settings.wsUrl = url;
+        this._adoptLiveSocket(ws);
+      };
+      ws.onerror = next;
+      ws.onclose = next;
+    });
+  }
+
+  /** Take ownership of a socket that has already delivered a live frame. */
+  _adoptLiveSocket(ws) {
+    this._ws = ws;
+    ws.onerror = () => {};
+    ws.onmessage = (evt) => { try { this._liveData = JSON.parse(evt.data); } catch { /* non-JSON frame */ } };
+    ws.onclose = () => {
+      console.log('[Observatory] WebSocket closed, falling back to demo');
+      this._ws = null;
+      this._liveData = null;
+      this.settings.dataSource = 'demo';
+      this._hud.updateSourceBadge('demo', null);
     };
-    tryNext(0);
+    this._hud.updateSourceBadge('ws', ws);
   }
 
   // async: `/ws/sensing` is gated (ADR-272); mint a single-use ticket first.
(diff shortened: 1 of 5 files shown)


jingilibili:
/cc @ruvnet — raw evidence behind this PR, kept separate from the description so it can be audited line by line. Everything below is **MEASURED** on the reporter's hardware; no synthetic numbers are used.

## Test array

| item | value |
|---|---|
| nodes | 3x ESP32-S3 (node 4 not powered), firmware v0.8.8 / ESP-IDF v5.4 |
| channel | 2462 MHz (ch 11), 20 MHz |
| CSI wire format | `0xC5110001`, `n_antennas = 1`, `n_subcarriers` mixed 192 / 306 / 128 / 64, datagram length exactly `20 + n_ant * n_sub * 2` (verified by decoding the UDP datagrams directly) |
| rate | ~30-35 fps per node, ~130 datagrams/s aggregate |
| path | ESP32 -> Windows host -> UDP forwarder -> WSL2 guest -> sensing server `:5005` |
| server | `sensing-server --source esp32 --http-port 8080 --udp-port 5005 --ui-path ../ui` (debug build) |

A first check ruled out the obvious suspects: the raw datagrams carry full, untruncated IQ data (e.g. `n_sub=192, len=404, expected=404`), so the amplitudes exist in the server long before they reach the wire.

## Finding 1 — the governed trust cycle strips raw amplitudes on ~46% of live cycles

A minimal WebSocket client counted, per `sensing_update`, whether every node's `amplitude` array was non-empty:

```
WINDOW 14s: total=689 full_amplitude=373 suppressed=316 other={}
```

`316 / 689 = 45.9%`. In the same window `GET /api/v1/status` alternated roughly once per second:

```
20:13:35 class= Anonymous  suppress= False demoted= False errs= 0
20:13:37 class= Restricted suppress= True  demoted= True  errs= 0
20:13:38 class= Anonymous  suppress= False demoted= False errs= 0
20:13:39 class= Restricted suppress= True  demoted= True  errs= 0
20:13:45 class= Anonymous  suppress= False demoted= False errs= 0
```

`engine_error_count` stayed `0` throughout, so this is a *tolerated-contradiction* demotion, not an engine failure. `EngineBridge::suppress_raw_outputs()` is true at `Restricted`, and `main.rs` then publishes `"amplitude": []` with `"subcarrier_count": 0` (the ADR-141 output mapping, review finding 1c). The governed engine is doing exactly what it documents — the defect is that its input signal fires on almost half of all cycles.

The trigger is `ContradictionFlag::TimestampMismatch` in `wifi-densepose-signal`'s `fuse_scored_calibrated`, which compares the **host arrival** timestamps of the newest frame per node against `config.soft_guard_us` (default `20_000`). `node_frames_from_states_with_guard` falls back to host-arrival time whenever mesh time is unavailable for any node, which is the case for plain CSI-over-UDP nodes. Each node contributes one frame per ~30 ms, so the newest-arrival spread is approximately uniform over one frame interval and a 20 ms soft guard lands at the median.

A/B on the same array, changing only that one parameter:

| `WDP_SOFT_GUARD_US` | window | total | full amplitude | suppressed | rate |
|---|---|---|---|---|---|
| `20000` (default) | 14 s | 689 | 373 | 316 | **45.9%** |
| `55000` | 14 s | 882 | 860 | 22 | **2.5%** |

With `55000` the status endpoint stayed at `raw_outputs_suppressed: false`. The hard guard (60 ms) is untouched, and the upstream invariant `stagger < soft_guard_us < guard_interval_us` still holds.

Reproducer for the counting client (dependency-free, stdlib only):

```python
# connect to ws://127.0.0.1:8765/ws/sensing, then per sensing_update:
lens = [len(nd.get('amplitude') or []) for nd in m.get('nodes') or []]
# full = all(l > 0 for l in lens); suppressed = all(l == 0 for l in lens)
```

## Finding 2 — the UI promoted an amplitude-stripped cycle to LIVE

`csi-simulator.js` used `Array.isArray(ampArr)` as its "a real frame was decoded" test, and `Array.isArray([])` is `true`. Combined with Finding 1, 45.9% of messages both promoted the watermarked view to LIVE and blanked the amplitude buffers to zeros, so the live readout showed `0.00` while `isLive` was already true. The fix in this PR makes the test `ampArr.length > 0` and keeps the last verified frame when a cycle carries no decodable CSI.

## Finding 3 — the port guess never matched the bundled server

```js
const defaultWsUrl = `ws://${window.location.hostname}:${httpPort + 1}/ws/sensing`;
```

`main.rs` binds `args.ws_port` (default `8765`) for the sensing WebSocket while the HTTP API + UI are on `8080`, so the guess resolved to `8081`. `ui/services/sensing.service.js` already carries the correct `8080 -> 8765` mapping; pose-fusion did not use it. This alone kept the page on SYNTHETIC data even after Findings 1 and 2 were addressed.

## End-to-end verification after the fixes

Headless Chrome, fresh profile, no manual WebSocket URL, twice:

```
<span id="status-label">LIVE CSI</span>
<button class="btn active" id="connect-ws-btn">&#10003; Live ESP32</button>
<span class="rssi-dbm" id="rssi-value">-54 dBm</span>
<span id="fps-display" class="fps-badge">31 FPS</span>
[CONSOLE] "[PoseFusion] frame 1 OK - mode=dual, csi.bufLen=1, embPts=1"
```

RSSI values match the node-side readings (`rssi_dbm: -54` for the node at the far end of the room vs `-31` for the node beside the AP), so the displayed number is the live capture, not the demo generator's `-78 dBm`.

## Suggested follow-up (not in this PR)

`WDP_SOFT_GUARD_US` is an operator override; the underlying surprise is that a cycle can silently lose its raw proxies. A rate-limited warning (or a `demotion_reason` field on `/api/v1/status`) naming `TimestampMismatch` would have turned this from a two-day hunt into one log line. Happy to open that separately if it is wanted.

ruvnet:
I'd suggest using a better sensor like the esp32 c5 or c6. The s3 are garbage

jingilibili:
Added commit `e2331154` — the follow-up suggested at the end of the description, now that the cause is proven rather than inferred.

## What changed

`/api/v1/status` reported `demoted: true` + `raw_outputs_suppressed: true`, and `main.rs` then published `"amplitude": []`. Nothing on any surface said *why*. That is the exact gap that made this bug expensive: an operator sees empty amplitude arrays and reasonably concludes the CSI pipeline is broken, when the pipeline is fine and a privacy mapping removed the field on purpose.

`EngineBridge` now records a stable slug for the trigger, exposed as `trust.demotion_reason` together with a cumulative `trust.demotion_count`, and the first demotion is logged immediately (later ones rate-limited to one per 30 s, on a limiter separate from the engine-error limiter so a demotion storm cannot hide an engine error):

`timestamp_mismatch`, `calibration_id_mismatch`, `phase_alignment_failed`, `drift_profile_conflict`, `coherence_drop`, `geometry_insufficient`, `mesh_at_risk` — joined with `+` when several fired, and `unnamed` if a demotion ever arrives with none of those inputs, rather than attributing a cause the engine did not produce.

## Live demonstration on the reporter's array

Same server, same array, only `WDP_SOFT_GUARD_US` differs, each sampled after ~12 s of live traffic:

**A — default soft guard (20 ms):**

```json
{ "demoted": true,
  "demotion_reason": "timestamp_mismatch",
  "demotion_count": 409,
  "effective_class": "Restricted",
  "raw_outputs_suppressed": true,
  "engine_error_count": 0 }
```

```
WARN sensing_server::engine_bridge: governed cycle demoted: ADR-141 strips the
per-node raw amplitude/phase proxies from the live publish for this cycle (see
/api/v1/status trust.demotion_reason). `timestamp_mismatch` means the newest
per-node frame timestamps are further apart than the fusion soft guard
(WDP_SOFT_GUARD_US); on a node array slower than 1/soft_guard frames per second
that fires on a large fraction of cycles
  demotion_reason="timestamp_mismatch" total_demotions=1 guard_interval_us=60000
```

**B — `WDP_SOFT_GUARD_US=55000`:**

```json
{ "demoted": false,
  "demotion_reason": null,
  "demotion_count": 27,
  "effective_class": "Anonymous",
  "raw_outputs_suppressed": false,
  "engine_error_count": 0 }
```

The counter stays cumulative while the reason describes the **most recent** cycle, so a consumer polling the endpoint never reads a stale cause; `a_clean_cycle_clears_the_demotion_reason` pins that behaviour. The 27 residual demotions in B are the genuine near-miss cycles inside the 55-60 ms band — the soft guard is still doing its job, it is just no longer below the array's own frame cadence.

## Tests

```bash
cargo test -p wifi-densepose-sensing-server --bin sensing-server engine_bridge
# running 13 tests ... test result: ok. 13 passed; 0 failed
```

- `timestamp_mismatch_demotion_is_named` — drives a 25 ms arrival spread (above the 20 ms soft guard, inside the 60 ms hard guard) and asserts the slug, the counter and the ADR-141 suppression.
- `a_clean_cycle_clears_the_demotion_reason` — a realigned cycle reports `null` and does not reset the counter.

The PR is now 4 commits / 4 files. The UI was re-verified against the rebuilt server afterwards: `LIVE CSI`, `Live ESP32`, `-54 dBm`, `31 FPS`.

jingilibili:
Hi im in iran and esp c5 or c6 cost me a lot by the way in your RuView
documents you said that it's ok on s3 and it has been test.

Viva people of Iran

jingilibili:
I am from Iran, and in Iran, obtaining these kits is very expensive, and working in Iran is difficult, and one has to work with the bare minimum. May the noble people of Iran prevail.

